PR #1382, opened 2013-03-07
Support GZIP request content

henryzhao81:
If request size is too big, enable gzip could help to make request smaller during transaction and improve the performance, especially string content.

Enable gzip just add header( Header header = new Header("Accept-Encoding", "gzip");), and encode json string content to "UTF-8".


deacix:
I am trying to get auth token and got with zuul spring cloud proxy between:
c.o.o.s.n.p.http.ONetworkProtocolHttpDb  : ->127.0.0.1: Error on invalid content:
31

Request content: username=admin&password=admin&grant_type=password

henryzhao81 wrote: "If request content format is json, yes. ". So i think this is a bug.

PR #1428, opened 2013-04-08
Require root password on server startup and store hash rather than plain text for verification

h00gs:
... root password on startup, store the hash of the root password (not the plain text) in the server config xml file, and require that it be changed from the default value.  A passphrase is permitted and a startup option "-p" is provided which allows the user to run the server for the sole purpose of generating a password hash that can be pasted into the server config xml file for subsequent verification.

Minimal code changes, uses existing encryption calls.  Does create a problem if you want to automatically spawn server instances.


lvca:
This has been implemented in 2.2.

PR #2281, opened 2014-04-25
- fixed javascript exception cannot flush to client properly

paynechu:
Because ScriptException and NoSuchMethodException is unserializable,
So it will throw exception again when it serialising the exception to client.
It caused never call to `channel.flush()`
Can test by call "SELECT test(:aid)" command through binary protocol.
There may have better way to fix. This issue just make a bit difficult to implement binary client.

```
function test(aid) {
  return db.query("SELECT * FROM V WHERE @rid=" + aid)
}
```


tglman:
Hi,

This is old, not valid any more  closing

PR #2326, opened 2014-05-11
Added gexf command to create gexf XML from query. This command uses

gianpaol0:
I added a command to create a gexf XML from query. This command is similar to gephi command.

Is my first contribution ;)


tglman:
Hi,

Really old, closing

PR #2549, opened 2014-07-08
Add System V init script data.

lewellyn:
N.B. I'm not trying to be elitist or exclusionist. I do not currently use any systems which require use of Upstart or Systemd or the ilk, so I can only contribute the header information for Linux distributions which use the System V chkconfig utility (RHEL [and its clones] at least through 6.x, ALT Linux, and others).

The block added to the orientdb.sh script in this PR allows for easier and faster installation of the database. It also helps remove user error from the equation when it comes to system startup.

As I find that the added block saves me a bit of time during installation, I'm hopeful it will find utility for others. I also hope that it will spur others to update it to make their own lives easier on other distributions and platforms, as well. :+1:

### DIFF
--- a/server/script/orientdb.sh
+++ b/server/script/orientdb.sh
@@ -1,5 +1,28 @@
 #!/bin/sh
 
+### BEGIN INIT INFO
+# Provides:          orientdb
+# Required-Start:    $remote_fs $network
+# Required-Stop:     $remote_fs $network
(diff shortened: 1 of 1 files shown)


lvca:
@tglman could you merge it manually? @lewellyn doesn't answered to my question months ago.

PR #2996, opened 2014-10-29
New test to validate the creation of indexes on sub properties

dcardin:
Feel free to move the test in a better place, or to modify it altogether


lvca:
@laa Could you take a look at this test and in case merge it?

lvca:
Not in master, in 2.0.x or develop.

laa:
@dcardin we do not have support for indexing of sub-properties.  I do not see a point to keep failing tests.
Thank you for contribution but I am afraid I have to reject it.

Issue #3268, opened 2014-12-24
JS API library doesn't support empty passwords


DanielRuthardt:
The server configuration allows empty passwords, the JS API library however doesn't support empty passwords.

lvca:
Not relevant anymore, closing it.

PR #3269, opened 2014-12-24
Resolves orientechnologies/orientdb #3268

DanielRuthardt:
JS API library doesn't support empty passwords


lvca:
I'm not sure if OrientDB supports empty password in the backend. Have you tried it?

DanielRuthardt:
OrientDB supports empty passwords, at least via the REST-API used by the JS API library.

lvca:
It's ok for me, but please send it to "develop" branch, not master: we use master only for releasing.

PR #3322, opened 2015-01-02
-fxied: issue with 'in' operator filter when array contains escaped sequence string

danrub:
issue: when using 'IN' operator inside predicate when array contains escaped sequence string, the query failed, and query returns empty result set.

user case:
CREATE DATABASE remote:localhost/temp root 1234 memory graph
INSERT INTO cluster:default CONTENT { value:"\"" }
SELECT FROM cluster:default where value in ["\""]


lvca:
Thanks!

danrub:
there are also issue with the escaped \uxxxx and comma inside a string example:

WHERE value IN [ "testing, command failure", "testing\u002c command failure"  ] 

both of them will cause a parsing error. the parser, looking for a first ',' encounter for item split decision, without testing if it inside valid string phrase.

I saw from other posts, that you are working on new parser, do you have an ETA ?
if not I can try to fix this issue my self, and send pull request when I'm done

PR #3356, opened 2015-01-09
-fxied: issue with 'in' operator filter when array contains string sequences with comma character or hex character

danrub:
Related to issue #3322.

Issue with the escaped \uxxxx and comma characters inside a string.

USE CASE:

INSERT INTO default CONTENT { value: "testing, command failure" }

SELECT FROM cluster:default
WHERE value IN [ "testing, command failure"]

SELECT FROM cluster:default
WHERE value IN ["testing\u002c command failure"]

both of them will return empty result. the parser fails, cause of looking for a first ',' encounter, 
causing to split prematurely moving to the next item, without testing if it inside valid string phrase.


lvca:
Thanks!

PR #3364, opened 2015-01-09
Remove reference to pipes in graphdb

mcculls:
Not sure if the use of `com.tinkerpop.pipes.util.structures.ArrayQueue` was intentional here, as AFAIK you shouldn't need to have the pipes library on the classpath in order to use the graphdb (at least wrt. minimal functionality). I replaced its use with `ArrayDeque` since orientdb now builds against a minimum of JDK6.


lvca:
Thanks again!

PR #3367, opened 2015-01-11
Use $ORIENTDB_HOME to calculate where to change directories

jdillon:
Change where directory is changed so that you can run server.sh and console.sh outside of the bin directory.

After this change, these commands should work:
```
./target/orientdb-community-2.0-SNAPSHOT/bin/server.sh
./target/orientdb-community-2.0-SNAPSHOT/bin/console.sh
```

... where previously you'd need to be in the 'bin' directory.

```
cd ./target/orientdb-community-2.0-SNAPSHOT/bin
./server.sh
./console.sh
```

NOTE: lots of other quoting issues in these scripts which could fail if the installation directory had spaces... may want to tidy these up generally.

### DIFF
--- a/graphdb/script/console.sh
+++ b/graphdb/script/console.sh
@@ -3,9 +3,6 @@
 # Copyright (c) Orient Technologies LTD (http://www.orientechnologies.com)
 #
 
-#set current working directory
-cd `dirname $0`
-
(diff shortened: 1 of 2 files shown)


lvca:
Thanks, I'll see other scripts adding " for directories.

PR #3374, opened 2015-01-12
shutdown needs to be before resetListener

rtm81:


laa:
Could you provide pull request without code stile reformatting. It is hard to understand what was really changed.

laa:
Did you move reset listeners at the end ?

laa:
If that is only change I will apply it manually then.

rtm81:
All i did: In method shutdown() i moved

      for (final WeakHashSetValueHolder<OOrientShutdownListener> wl : weakShutdownListeners)
          try {
              if (wl != null) {
                  final OOrientShutdownListener l = wl.get();
                  if (l != null)
                      l.onShutdown();
              }
              
          } catch (Exception e) {
              OLogManager.instance().error(this, "Error during orient shutdown.", e);
          }

before 
resetListeners();

laa:
But "strong" listeners are still cleared before calling them.
Do you mind if we will do following move reset listeres bellow and revert rwspinlock cleanup I think it is not needed and main issue was absence of clean up on more high level object. Do not you mind to check it ?

laa:
I merged changes manually.

laa:
Thank you for your help !

Issue #3377, opened 2015-01-12
doc.removeField() works not as expected


PhantomYdn:
Guys,

Following testcase is failing (please read comments):

```java
       @Test
	public void testRemovingReadonlyField()
	{
		ODatabaseDocument db = ODatabaseRecordThreadLocal.INSTANCE.get();
		OSchema schema = db.getMetadata().getSchema();
		OClass classA = schema.createClass("TestRemovingField");
		classA.createProperty("name", OType.STRING);
		OProperty property = classA.createProperty("property", OType.STRING);
		property.setReadonly(true);
		
		ODocument doc = new ODocument(classA);
		doc.field("name", "My Name");
		doc.field("property", "value1");
		doc.save();
		
		doc.field("name", "My Name 2");
		doc.field("property", "value2"); //setting of value to readonly document
		doc.removeField("property");  //we changed our mind: decided don't modify readonly field
		doc.save();  // failing right here
	}
```

PhantomYdn:
Btw, this failing as well:

```java
@Test
	public void testRemovingReadonlyField2()
	{
		ODatabaseDocument db = ODatabaseRecordThreadLocal.INSTANCE.get();
		OSchema schema = db.getMetadata().getSchema();
		OClass classA = schema.createClass("TestRemovingField2");
		classA.createProperty("name", OType.STRING);
		OProperty property = classA.createProperty("property", OType.STRING);
		property.setReadonly(true);
		
		ODocument doc = new ODocument(classA);
		doc.field("name", "My Name");
		doc.field("property", "value1");
		doc.save();
		
		doc.field("name", "My Name 2");
		doc.field("property", "value2");
		doc.undo(); //we decided undo everything
		doc.field("name", "My Name 3"); //change something
		doc.save();
	}
```

In a meanwhile, let me propose small enhancement: please allow to undo just 1 field.

tglman:
Hi @PhantomYdn,

The scope of removeField is not revert to the previous value, as today there's no api for do that, the correct way to do it is with undo, as is shown in your second test case.
I fixed it, now the second test case work as expected.

Anyway thank you very much, you issue are everytime filled of perfect details and test cases.

closing

PhantomYdn:
Thank you, @tglman,

Can you please add "undo()" just for 1 field?

Let me clarify a use case.

[...]

PR #3423, opened 2015-01-19
Add ability to undo one field

PhantomYdn:
Related to issue #3377 

Guys, please consider merging this into 2.0 Final.


lvca:
@tglman please could you check this PR and in case merge it?

tglman:
Hi @PhantomYdn,

Nice pull request, anyway there are a couple of things to correct,  the first is if no value was changed fieldOriginalValues can be null, the second if you do an undo of a never changed filed it will set it to null, instead of do nothing, you can use containsKey to check that the value was changed.

a couple of tests also in this two cases will be cool ;)

PhantomYdn:
@tglman 
You are right! Btw, your implementation of undo() was valnurable for NPE too.

Guys,

Please-please, include it to 2.0 Final.
It's really important for raleasing Orienteer 1.0... Without per field basis undo: it's not possible to have calculable attributes.

lvca:
Thanks!

PR #3487, opened 2015-01-28
Resolves #3268 - JS API library doesn't support empty passwords

### LINKED ISSUE #3268: JS API library doesn't support empty passwords

DanielRuthardt:


lvca:
Thanks.

Issue #3561, opened 2015-02-11
Not unique identifier in detachAll [orientdb-object]


thewall13:
This issue is based on: #2525 

To avoid cycles in the detachAll method, all detached objects will be added to a map. The key is based on ODocument.hashCode().

But the method hashCode() has to much collisions to use it as key. 

Example:   
The Object with the ORID   #9:31 has the same hashcode as an Object with #10:0

So I suggest to use the ORID itself as key. The only difference would be a real unique key in the alreadyDetached map.

thewall13:
https://github.com/orientechnologies/orientdb/pull/3562

tglman:
the related pull requests are merged, closing

PR #3562, opened 2015-02-11
replaced hashcode() with getIdentity() as indentifier in the map of

thewall13:
already detached objects


tglman:
hi @thewall13  this look good, does all the tests pass with this ?

thewall13:
the detached-tests worked fine. I have three test-failures with "ant test" but it doesn't seems to be something with these changes.

I only switched the key in the map, so everything should work fine.

thewall13:
Here is the missing related issue: https://github.com/orientechnologies/orientdb/issues/3561

wouterv:
I can confirm I had problems with this issue (which I introduced) and that this pull request fixes it. I have no clue why I used hashCode in the first place...

So I hope it can be merged into 2.0.3!

lvca:
Merged. Thanks guys. I'm backporting it to 2.0.3...

PR #3563, opened 2015-02-11
small javadoc change

balamaci:
small change javadoc removed wrong misleading reference to Edge class when creating a Vertex


lvca:
Thanks @balamaci for the PR, may you do it against "develop" branch? We use master only for releasing. Thanks.

PR #3565, opened 2015-02-11
small javadoc change

balamaci:
small change javadoc removed wrong misleading reference to Edge when actually creating Vertex.


lvca:
Thanks!

PR #3567, opened 2015-02-11
Update dserver.sh to allow ./some/path/to/dserver.sh

vitorenesduarte:


lvca:
Vitor, what's the goal of this change?

vitorenesduarte:
It's only possible to run dserver.sh from __ORIENTDB_HOME/bin__ folder.
As it's possible to run server.sh from anywhere, I think dserver.sh should also be.
It's just a suggestion. You probably have a better way to do this.

lvca:
Sounds good. May you do it against "develop" branch? We use master only for releasing. Thanks.

lvca:
Waiting for new PR ;-)

vitorenesduarte:
#3569 Don't know If I did it right.

vitorenesduarte:
#3570 Sorry for the three PR. I think now it's ok.

PR #3589, opened 2015-02-15
Work around for NullPointerException in OpenJDK if no args

PhantomYdn:
Guys,

There is a bug in OpenJDK: if arguments for JS function invokation were supplied as null (instead of new Object[0]) you will have NullPointerException. Everything is find on Oracle JDK in this case.

Please approve PR and release in your nearest hot-fix version.


lvca:
Thanks! Ported also into 2.0.3-hotfix.

PR #3603, opened 2015-02-17
Corrected typo: "No enough" to "Not enough"

teachrdan:


lvca:
Thanks for the PR, may you do it against "develop" branch? We use master only for releasing. Thanks.

PR #3610, opened 2015-02-17
Update dserver.sh to allow ./some/path/to/dserver.sh

vitorenesduarte:


distributed/script/dserver.sh
```diff
@@ -1,5 +1,5 @@
-#!/bin/sh
+#!/bin/bash
```
vitorenesduarte: prevent BAD SUBSTITUTION errors: sh sometimes points to dash instead of bash (from what I understood)

distributed/script/dserver.sh
```diff
@@ -1,5 +1,5 @@
...
 #
 # Copyright (c) Orient Technologies LTD (http://www.orientechnologies.com)
 #
-./server.sh -Ddistributed=true $*
+"${0/%dserver.sh/server.sh}" -Ddistributed=true $*
```
vitorenesduarte: - paths with spaces were breaking this
- there might be other situations where this breaks

PR #3617, opened 2015-02-17
Changed "No enough" to not enough across 3 files

teachrdan:
Found every instance of "no enough" and changed them to "not enough," and changed minor language in one case for clarity.


lvca:
Hi @teachrdan, thanks again. I see you have many commits out of scope from your changes. I.E. you have 2.0.2 in your pom.xml, but develop has 2.1-SNAPSHOT. Could you align your repo with our "develop" and update this PR?

Issue #3628, opened 2015-02-19
orientdb.sh for RedHat chkconfig


gfinger:
For enabling the orientdb.sh script with the chkconfig tool on RedHat Linux, certain headers must be included at the beginning of the script. These headers are quite simple, and I think they are of no harm for other Linuxes. Thus I would like to propose to include these headers in the delivery.
These headers are:

```
# chkconfig: 2345 20 80
# description: OrientDb init script
# processname: orientdb.sh
```

they should be included right on the next line after the shebang.

lvca:
What does mean 

    chkconfig: 2345 20 80

? Could you send us a Pull Request to "develop" branch, please? We'll more than happy to merge it.

gfinger:
This denotes the script will start the OrientDB server for the runlevels 2, 3, 4 and 5. In addition, the start priority will be set to 20 while the stop priority will be 80.
Yes, I will send a pull request to the develop branch.

lvca:
Thanks!

robfrank:
Headers are present on orientdb.sh:

https://github.com/orientechnologies/orientdb/blob/develop/distribution/script/orientdb.sh

PR #3642, opened 2015-02-23
Add headers for RedHat chkconfig

gfinger:
See issue #3628
https://github.com/orientechnologies/orientdb/issues/3628


lvca:
Thanks

Issue #3644, opened 2015-02-24
Detaching in non-proxied with multiple refs to the same object, creates copies


wouterv:
If you have a parent object which contains a reference to a child object and this child object is also contained in a collection of the parent, detaching non-proxied will give two different instances of this detached object, while they actually should be the same (working on a pull request right now for this).

wouterv:
Pull request: #3645

tglman:
pull request merget closing

PR #3645, opened 2015-02-24
Fix for #3644

wouterv:
Have multiple refs to the same object always return the ame instance when detaching non-proxied

Extended unit test to try this case. All other tests still run smoothly. Tried it in our environment en looks good.

Candidate for 2.0.4 hotfix?


lvca:
Thanks @wouterv, you're becoming the highest contributor this month!

wouterv:
Lol, really?

@lvca  PS: do you think this will be backported to 2.0.4 hotfix?

lvca:
Yes, I'm merging it now in 2.0.4-hotfix branch...

lvca:
Done.

wouterv:
Thanks!

PR #3664, opened 2015-02-26
small service methods for ODocument

svenruppert:


svenruppert:
sorry, wrong branch..  will redo on develop

lvca:
Hey @svenruppert I don't like to mix Object and Document concepts. Why should you pass an object or class to ODocument? Are you using the Object API?

svenruppert:
Hi, I am using the Document API, but mapping parts from Objects.. but you are right, mixing concepts is not nice.

Issue #2741, opened 2014-08-29
Cannot insert composite key into manual index using SQL/console


ghost:
OrientDB  Community 1.7.8

According to the documentation, this is possible:

   select from index:books where key = ["Donald Knuth", "The Art of Computer Programming", 1968]

However, I cannot figure out how to insert such a value

   insert into index:books (key, rid) values (["Donald Knuth", "The Art of Computer Programming", 1968], #12:0)

   Error: com.orientechnologies.orient.core.exception.OCommandExecutionException: Error on execution of command: sql.insert into index:Test1 (key,rid) values (["a","b","c"], #12:0)
   Error: java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.lang.Comparable

The same occurs with this alternate syntax:

   insert into index:books set key = ["Donald Knuth", "The Art of Computer Programming", 1968], rid = #12:0

danrub:
The issue still remains, when using a manual composite indexes.

USE CASE:

CREATE INDEX test unique string,string
INSERT INTO INDEX:test (key,rid) values (['test','success'],#1:0)

failing at OSBTree line 188.
java.lang.ClassCastException: java.util.ArrayList cannot be cast to com.orientechnologies.orient.core.index.OCompositeKey

lvca:
@danrub Have you tried with 2.0-SNAPSHOT or 2.0-rc2?

danrub:
Yes, the case is from the last pull of the develop branch, one day ago 12 January 2015

tglman:
fixed, closing

tglman:
i was forgetting to mention to work correctly the if the index already exist should be dropped and recreate.

Solido:
I need to use composite index on 2.2.15 and I get
 _ClassCastException : ArrayList cannot be cast to Comparable_ 
in Studio

tglman:
hi @Solido,

What are the steps are you doing to get this error? I understood the error but would be easier to fix with the list of steps.

Regards

PR #3677, opened 2015-03-02
fix partial delete multi value index unexpected behavior

danrub:
manual index command, DELETE FROM index:<indexName> WHERE key = <key>
was failing on composite key, when rid wasn't specified.

expected behavior, to remove all values under specified (partial/full) key

Use case:
CREATE INDEX index:test notunique string,string
INSERT INTO index:test (key,rid) values (['test','me'],#1:1)
INSERT INTO index:test (key,rid) values (['test','me'],#1:2)

DELETE FROM index:test where key = ['test']
SELECT FROM index:test

expected return 1 row 
actual return 2 row

related to #2741


lvca:
Thanks!

PR #3681, opened 2015-03-03
- use precompiled patterns

machak:


machak:
this is a micro optimization (as compiling patterns is quite "expensive" operation)

lvca:
Thanks!

PR #3696, opened 2015-03-04
- make inner classes static (use little bit less memory & less prone to ...

machak:
...memory leaks)

- do not allocate zero-length arrays
(cherry picked from commit 71e29b3)


lvca:
@tglman Fee free to merge it ;-)

tglman:
i don't love the one class costants but it can make sense, merging i already tested that don't break base tests

PR #3701, opened 2015-03-05
No restriction on reserved keywords ('increment') ?

matanshukry:
It seems nothing stops you from using a reserved keywords on a field name.
At start this may seems fine (although design issues may come up), but it does seem to make issues;
e.g. When trying to create a property with name 'increment', you'll get an exception.

I tracked it down to the parser for a like:
SELECT FROM <class> WHERE <property>.type() not in [..] and <property> not null

which when <property> is 'increment', throws an exception.

This pull request contains a test for creating reserved keywords, and fix to throw an exception if one tries to create a property with a reserved keyword.


lvca:
@matanshukry thanks for this PR. Today we introduced the escaping of field to allow using some keyword like:

```sql
select from from V
```

Where from is the field name. Now this works:
```sql
select `from` from V
```

What problem do you have with having "increment" as field name? Only on using SQL? If yes, this is already resolved with escaping.

matanshukry:
This fixes my issue, thanks.

PR #3705, opened 2015-03-05
Definition of line feed and basic java style

krichter722:
The management of line feeds/EOL should be managed by git by setting `git config core.autocrlf true` and changed once in the source. The request than contains basic java code style fixes for brace usage in if-else and for-statements.


tglman:
we don't have so strict role on if-else or for statement brackets, i appreciate your effort but i'll suggest you to start from something more logic meaning full and that change a bit less files,

krichter722:
I won't argue about the refusal of the pull request because of the right points you make. It'd recommend to make a rule, though, because missing braces cause compiler warnings and therefore their enforcement is not exclusively up to project definitions.

PR #3710, opened 2015-03-05
added .gitattributes to enforce LF as line feed

krichter722:


tglman:
I'm ok with this, any against?

tglman:
checking the file the setting is commented: '#* text=auto' so it won't change anything, if you make one with the setting enabled i'll merge straight ;)

krichter722:
Well, that's slightly embarrassing, thanks for your patience, see #3720.

PR #3711, opened 2015-03-05
Warnings public constructors

krichter722:
Hiding constructors in helper classes is more elegant and avoids compiler warnings. Providing explicit constructors with minimal visibility makes detection in architecture and abstraction more easy.


tglman:
on final point, can you do the pull request to develop instead of master ? we use that as main development branch, this is a good canditate for merge.

PR #3712, opened 2015-03-05
String cleanups

machak:
some more micro-optimizations & some code cleanup here and there, won't hurt ;-)


tglman:
Hi, it looks quite ok, my only concern is for the backword compatibility, i saw that you added some microbenchmark using the openjdk microbenchmark, till now we guarantee the compatibility to java 1.6, does it work in that version ?

be aware that contribution that change a lot of files are really hard to review, in this case i did, it look quite good and checked the backward compatibility issue can be merged, but in general bigger is the number of file changed and smaller is the chance to get the pull request merged.

machak:
Hi,
I'll make sure next pull requests are smaller. 
About benchmark stuff: I bumped JMH version as well, so that could be a problem, but feel free to exclude (revert) those changes.

tglman:
Li said before is too big, closing this.

Issue #3706, opened 2015-03-05
Move to standardized logging framework (logback) or document logging classes


krichter722:
Currently logging is managed and done with classes in `com.orientechnologies.common.log`, `java.util.logging` and `System.out/err`. It should standardized by using either `com.orientechnologies.common.log` (which definitely requires documenting the classes - otherwise it's very hard for new contributors to figure out their purpose) or a logging framework (switching away from `java.util.logging` to something more configurable and flexible, like `logback` would make sense in the review process).

I'd be willing to contribute in from of a pull request after the `com.orientechnologies.common.log` classes are documented or explained here and declared obsolete.

PR #3722, opened 2015-03-06
updating test classes with main methods

krichter722:
removed main methods in test classes and replaced with @Test; removed System.out statements and added slf4j dependency (doesn't interfere with other logging facilities and works as expected); minor beautification


tglman:
we don't want to add additional logging library, is good remove the System.out.print... but should be replaced with
`OLogManager.instance().warn() //or any other lels`

krichter722:
I'm waiting for input in https://github.com/orientechnologies/orientdb/issues/3706 (it's hard to see through the logging facilities). Maybe you can explain the situation there or even better in javadoc, `HACKING`, `README` or something similar (including links to existing docs) and fix #3706. Then I can open a new pull request reflecting your comment for this one. Thanks for taking a look.

PR #3723, opened 2015-03-06
moved public inner classes and interfaces in OrientDB Test Commons to ap...

krichter722:
moved public inner classes and interfaces in OrientDB Test Commons to appropriate places


tglman:
nothing wrong here, merging :)

PR #3725, opened 2015-03-07
Variable ports

krichter722:
started parameterizing ports in OStorageRemote to fix test on default port later; removed unused private deprecated fields; basic code style for OStorageRemote (braces, @Override annotations, missing javadoc parts, avoiding method parameter assignments, using char instead of string for string search, added null test for network in order to provide a better application feedback rather than leaving the test to the JRE which would fail with NullPointerException, removed star import, removed redundant casts, corrected exception types (including removal of catching Throwable (you don't want to do that))

### DIFF
--- a/client/src/main/java/com/orientechnologies/orient/client/remote/OStorageRemote.java
+++ b/client/src/main/java/com/orientechnologies/orient/client/remote/OStorageRemote.java
@@ -124,12 +124,18 @@ public class OStorageRemote extends OStorageAbstract implements OStorageProxy {
   private Map<String, Object>           connectionOptions;
   private OEngineRemote                 engine;
   private String                        recordFormat;
+  private int                           port                  = DEFAULT_PORT;
+  private int                           sslPort               = DEFAULT_SSL_PORT;
 
   public OStorageRemote(final String iClientId, final String iURL, final String iMode) throws IOException {
     this(iClientId, iURL, iMode, null);
   }
 
   public OStorageRemote(final String iClientId, final String iURL, final String iMode, STATUS status) throws IOException {
+    this(iClientId, iURL, iMode, status, DEFAULT_PORT, DEFAULT_SSL_PORT);
+  }
+
+  public OStorageRemote(final String iClientId, final String iURL, final String iMode, STATUS status, int port, int sslPort) throws IOException {
     super(iURL, iURL, iMode, 0); // NO TIMEOUT @SINCE 1.5
     if (status != null)
       this.status = status;
@@ -146,6 +152,8 @@ public class OStorageRemote extends OStorageAbstract implements OStorageProxy {
     asynchExecutor = Executors.newSingleThreadScheduledExecutor();
 
     engine = (OEngineRemote) Orient.instance().getEngine(OEngineRemote.NAME);
+    this.port = port;
+    this.sslPort = sslPort;
   }
 
   @Override
@@ -1784,7 +1792,7 @@ public class OStorageRemote extends OStorageAbstract implements OStorageProxy {
     // REGISTER THE REMOTE SERVER+PORT
     if (!host.contains(":"))
       host += ":"
-          + (clientConfiguration.getValueAsBoolean(OGlobalConfiguration.CLIENT_USE_SSL) ? getDefaultSSLPort() : getDefaultPort());
+          + (clientConfiguration.getValueAsBoolean(OGlobalConfiguration.CLIENT_USE_SSL) ? sslPort : port);
 
     if (host.contains("/"))
       host = host.substring(0, host.indexOf("/"));
--- a/client/src/main/java/com/orientechnologies/orient/client/remote/OStorageRemote.java
+++ b/client/src/main/java/com/orientechnologies/orient/client/remote/OStorageRemote.java
@@ -111,10 +111,6 @@ public class OStorageRemote extends OStorageAbstract implements OStorageProxy {
   private int                           networkPoolCursor    = 0;
   private OCluster[]                    clusters             = OCommonConst.EMPTY_CLUSTER_ARRAY;
   private int                           defaultClusterId;
-  @Deprecated
-  private int                           minPool;
-  @Deprecated
-  private int                           maxPool;
   private ORemoteServerEventListener    asynchEventListener;
   private String                        connectionDbType;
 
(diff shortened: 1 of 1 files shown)


PhantomYdn:
@krichter722 I see lots of PRs and changes from you, but let me ask: is there some functional changes or just formatting/performance changes?

I'm not teammember of Orientechnologies, but if you would commit something to my repositories I would scary a lot:))
1) Formatting is something personal: you like "if(...) {return;}", but I like "if(...) return;" - why change to the way how you like it?:)
2) Some of your changes are very unsafe. Even testcases might be OK, but on production that might lead to some unpredictable things. Examples: you changed '} catch (Exception e) {' to 'catch (IOException e)'. But catch can be used for runtime things too.
3) (it's safe, but useless) 345 line: java is passing elemental variables by value, so you can safely change passed iMode instead of creation of new variable. And there are lost of places like this.
4) you are adding check for 'if(network==null)' in a lot of places. But almost everywhere network is internal for that method variable. Why it should be checked if it's not null for sure? And more over: in case of unexpected null for internal variable I would prefer to see NullPointerException instead of IllegalStateException. And more over: in combination with limiting of catch closure: it might lead to interesting effects...
5) you are changing "if (((String) listener.get("protocol")).equals("ONetworkProtocolBinary"))" to "listener.get("protocol").equals("ONetworkProtocolBinary")". But I would prefer first version: because in case of wrong type you will have clear exception about that (ClassCastException) instead of unpredictable behavior.

And so on...

I can't say for OT team, but I would straggle that types of changes...

krichter722:
Thank you for looking so deeply into the changes. They're not functional changes, although I think they have potential to reveal some issues.
> 1) Formatting is something personal: you like "if(...) {return;}", but I like "if(...) return;" - why change to the way how you like it?:)

Due to else-ambiguity there's no doubt that adding braces has certain advantages. I have to admit that I thought a warning to be produced by the compiler as well, but is was just the NetBeans parser, so it's not "officially" discouraged. It is not a simple question of taste.
> 2) Some of your changes are very unsafe. Even testcases might be OK, but on production that might lead to some unpredictable things. Examples: you changed '} catch (Exception e) {' to 'catch (IOException e)'. But catch can be used for runtime things too.

That depends what you define as "save". Catching an exception without any statement in the catch block can cause the application to enter in an undefined state in which it might be better to crash than to continue running, so I wouldn't speak of "safety" per se. If you don't trust your test cases, you shouldn't merge. Furthermore, Imagine what catching `Exception` or even `Throwable` means. It is practically impossible IMO because writing a catch block which can handle all exceptions would be a life tasks and doing so in the main method/web server entry method/web beans/... is the only possible location for such drastic action.

> 3) (it's safe, but useless) 345 line: java is passing elemental variables by value, so you can safely change passed iMode instead of creation of new variable. And there are lost of places like this.

I think avoiding assignments to method parameter increases readability and shouldn't have any disadvantage in comparison to what the compiler does with the code.
> 4) you are adding check for 'if(network==null)' in a lot of places. 

`NullPointerException` with an explicit message is fine as well. I prefer `IllegalStateException` because it indicates that the class got out of a valid state at some undefined point. The tradeoff in the compiled code should be zero as well.
> 5) you are changing "if (((String) listener.get("protocol")).equals("ONetworkProtocolBinary"))" to "listener.get("protocol").equals("ONetworkProtocolBinary")". But I would prefer first version: because in case of wrong type you will have clear exception about that (ClassCastException) instead of unpredictable behavior.

I'd suggest to either change the class of `listener` to return the protocol as a `String` or test different things (`instanceof String` and equality) in different tests, e.g. add

    if(!(listener.get("protocol") instanceof String)) {
      throw new IllegalStateException("protocol property of listener isn't a String");
    }

before the critical equality test.

I know that it's a lot of changes and some might not appear to have a low gain comparing the effort to merge and improvements, but I managed to reveal some architectural deficies in another branch I'm working on currently with a different approach.

PhantomYdn:
Let me comment just some of the points:) 

0) Question of taste: imho such things should follow original taste of authors.
2) You are almost right. It's ok if you write code from scratch. But I hope that you do agree, that such changes: that's not safe change in logic. If catch was wide it might lead to unpredictable errors if you make it more specific.
3) Imho original code more readable. And more over: several variable of the same nature might lead to user confusion (which var should I use?) and subsequent mistake (I used var instead of var0 which was changes). In case when it's important do not change original value: arg should be marked by "final".
4)  Almost every exception in java that's a sign that something going wrong... And it's bestpractice to through specific exception instead of using IllegalStateException in every unknown situation. Please look into SDK source code: why they throwing in a lot of places NullPointerException and etc. instead of IllegalStateException? But I do agree: that in this particular case that's just a question of taste.
5) The same as for 4. Why use IllegalStateException instead of more specific one: ClassCastException?

lvca:
Hi @krichter722, we appreciate all your contribution. In orde to understand what you have changed, please don't mix formatting things otherwise takes 1h to analyze all your changes line by line.

Please could you send us individual PR focused on changes? In this case the support for ports.

krichter722:
> 2) You are almost right. It's ok if you write code from scratch. But I hope that you do agree, that such changes: that's not safe change in logic. If catch was wide it might lead to unpredictable errors if you make it more specific.

I guess we're not on the same page whether an exception which causes the application to crash is more secure than it becoming non-deterministic after a too-wide exception has been caught and ignored. I'd like to stress the strong difference between safety and stability which I see.

> 4) Almost every exception in java that's a sign that something going wrong... And it's bestpractice to through specific exception instead of using IllegalStateException in every unknown situation. Please look into SDK source code: why they throwing in a lot of places NullPointerException and etc. instead of IllegalStateException? But I do agree: that in this particular case that's just a question of taste.
5) The same as for 4. Why use IllegalStateException instead of more specific one: ClassCastException?

Because in both cases the `null` pointer and the cast are not the reason for the illegal state of the application. Long before something else than a `String` got into the property map and the method which initializes `network` returned `null` (because network connection was dropped wrong code got injected at runtime or else) something went wrong - at the point the exception is thrown we don't know the reason and the fact that a capsuled variable is `null` doesn't help the caller. In the SDK you can throw `NullPointerException` because it's a stateless library routine and the `null` pointer is definitely the (only) cause for the exception. If the (real) cause is unspecific the exception should be unspecific afaik. Moreover throwing `NullPointerException` or `ClassCastException` explicitly obliges the caller to catch those which decreases maintainability.

krichter722:
@lvca I will do that. Seems like getting a feedback in an issue report before coding and sending PRs is a good idea, too, because you're quite responsive as well ;)

PhantomYdn:
@krichter722 Regarding to NullPointerException, ClassCastException and so on. It seems that we are not on the same page about that.  I think that you know about a "contract" approach in OOP: if you are using some method which encapsulate some internal lifecycle - support of that lifecycle should be their own and of illegal state is possible - it should be thrown within that methods, but not externally: checking of a state outside of a classes involved into lifecycle - that's not good. And btw: there is no obligation to catch NullPointerException or ClassCastException... they are runtime and show that something especially wrong with the code. But if you are doing if(...) throw - you are responsible for proper handling of those situations.

lvca:
@krichter722 It's hard to check what you changed, because you used a different formatter where {} are added at if with 1 line only. Could you please disable such formatting ad update this PR?

lvca:
Closing this because it's hard to understand the changes (see above).

PR #3732, opened 2015-03-08
Feature/correct history location

bougyman:
without the exec it's forked by sh and sending a term to the sh parent does not send a term to the running java process. This corrects that so it can be properly supervised by systemd, runit, daemontools, etc.


lvca:
@tglman please could you test it's ok with all the rest?

lvca:
@bougyman Thanks for the PR, may you do it against "develop" branch? We use master only for releasing. Thanks.

PR #3738, opened 2015-03-09
Added a test case(distributed) and a fix

matanshukry:
### The Problem (User perspective)
The problem is an edge case of a distributed record update in a transaction.
Consider the following series of events:
 * 0. Create new document with value 2L
 * 1. Node 0 starts a transaction, and loads document with version 0 and value 2
 * 2. Node 1 starts a transaction, loads the document with version 0 and value 2, updates it to value 4, and commit(v1)
 * 3*. Node 1 starts a transaction, loads the document with version 1 and value 4, updates it to value 5, and commit(v2)
 * 4. Node 0 update his document (v0, 2) to value 5, and commit.

Note: event 3 isn't necessary, it just shows the versions don't have to be consecutive nor have any relation. They have no impact on the result what so ever.

At the last commit of node 0, an exception should be raised. Even though the content is the same, the versions in the beginning of the transactions aren't, and we do not know the intention of the transaction.

For example, consider the use case of two people trying to increment the same field, starting at 0. The result should always be 2, even if the above scenario happens, which currently, it isn't.

### The problem (code perspective)
When we try to commit a new update to a document it is distributed to the nodes.
Each node will attempt to merge the new document with the old one, and if the _content_ is different, it will set it as dirty.
If the result document is not dirty, it won't even check the versions.

### The solution
When a node will attempt to merge a new document with an old one, on a distributed update task, it will now also check if the versions match. If not, it will set it as dirty.

Not sure how 'correct' it is, but it definitely works, and not a performance hit either.

### Extra: a test case
I also added a test case, which you can run, and see it fails (don't forget to comment out the fix part).

### DIFF
--- a/distributed/src/test/java/com/orientechnologies/orient/server/distributed/AbstractServerClusterMergeUpdateTest.java
+++ b/distributed/src/test/java/com/orientechnologies/orient/server/distributed/AbstractServerClusterMergeUpdateTest.java
@@ -1,0 +1,99 @@
+package com.orientechnologies.orient.server.distributed;
+
+import com.orientechnologies.orient.core.db.OPartitionedDatabasePoolFactory;
+import com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx;
+import com.orientechnologies.orient.core.exception.OConcurrentModificationException;
+import com.orientechnologies.orient.core.id.ORID;
+import com.orientechnologies.orient.core.record.impl.ODocument;
+import org.junit.Assert;
+
+/**
+ * @author Matan Shukry (matanshukry@gmail.com)
+ * @since 3/9/2015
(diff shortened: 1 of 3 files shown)


lvca:
@matanshukry thanks for this PR, it's high quality contribution with a test case and rich comment about it. Thanks, I'm reviewing it asap.

lvca:
@matanshukry yesterday I pushed a fix on a problem close to this. Please could you check that 2.0.5 (just released) still has this problem? if not, please close this PR.

matanshukry:
I'm afraid not, this PR is still required. I did merged the new commits though, so a merge can be made again without conflicts.

lvca:
Merged, thanks!

PR #3740, opened 2015-03-10
Added an error callback, improved success callback and added asynchronous querying

DanielRuthardt:
I am fully aware, that the change in line 279 might and probably will break existing code. The success callback however doesn't make any sense, at least I can't think of any situation where it would be needed to set a success callback, if the code is blocked anyways until success.

If there is a success callback, there probably should be an error callback as well.

To make the success callback useful, I am making the request an asynchronous one if the success callback is provided, additionally the success and error callback get a little more useful adding the data or error message as first parameter.


lvca:
Makes sense. Thanks ;-)

Issue #2939, opened 2014-10-14
Traits


mattaylor:
This is proposal to support a simple form of multiple inheritance through the concept of 'traits', which would allow more easily reusable composable behaviors in classes.

Suppose I have a class 'tag' and I would like to define a set of behaviors around 'taggable' classes ,e.g
```
create class tagable extends OTrait;
create property tagable.tags LINKSET tag;
alter class tagable onBeforeUpdate beforeUpdateTagable; #checks that tags are all valid links
create index tagable.tags NOTUNIQUE;
```
Now suppose I want to defined another set of behaviours for a classes that implmnent 'prev' and 'next' links so I can traverse through ordered records that share a common 'type' eg..
```
create class chainable extends OTrait;
create property chainable.prev LINK chainable;
create property chainable.next LINK chainable;
create property chainable.type STRING; # the common key for determining prev and next links;
create property created DATETIME;
alter class chainable onBeforeCreate beforeCreateChainable; # update prev/next links when a new record is created
create index chainable.typeTime on chainable(type, time) UNIQUE;
```

If I want to create a class that is both tagable and chainable then I could do something like

```
create class listItems implements tagable,chainable
```` 
Now listItems will have all the properties of both chainable and tagable traits and will execute the hooks associated with each in the order defined by the implments attribute. It would also use the same indexes that those traits defined. 

This would be also a very usefull way of composing new classes that might extend built in super classes like ORestricted, vertex and edge.

**Comment by mattaylor** [comment]
Im thinking that this could be more simply (and transparently) implemented without having to extend OTrait, (or similar) by using a 'mixin' keyword eg..

```
create class myClass mixin super1,super2
```

This would simply compose the class myClass by copying over properties and chaining hooks from each superclass in the order specified. 

Not clear however what the appropriate behaviour would be if the properties of each superclass were later modified..

PhantomYdn:
I disagree with copying of properties...

But guys: please help me understand, what the problem with an ability to have multiple superClasses?
There should be only one check: absence of cycles. Or oven more correct: even cycles might be allowed, but runtime collection of properties for a particular class should stop in case of presence of already introspected superclass.

lvca:
I'd love having this feature, but impacts could be huge. The good part is that inheritance is always executed at run-time, so no property is copied during creation of sub-class.

lvca:
@PhantomYdn would you like to create a POC?

PhantomYdn:
Sure - will try

[...]

Issue #3629, opened 2015-02-19
Renaming Property


vitorenesduarte:
**SNAPSHOT**

Hi,

``` sql
create class User
create property User.propA string
create index User.propA unique

insert into User set propA = 'valueA'

alter property User.propA name propB

select from User
```

![screen shot 2015-02-19 at 16 02 35](https://cloud.githubusercontent.com/assets/4520516/6270255/7ac0cbe6-b850-11e4-82cc-a308452a1015.png)

propA and it's value are kept in the record. Is this supposed to happen?
Besides that, I was also hoping that the index would be updated:

``` sql
insert into User set propA = 'valueA' # this should work (doesn't)
insert into User set propB = 'valueA' # this shouldn't work (does)
```

Let me know if I did something wrong. 
Thanks

PhantomYdn:
Let me subscribe: I know about this issue...

Could you please clarify what kind of engine are you using? Remote, plocal or memory?

vitorenesduarte:
A noticed this using Studio.

hohwille:
I just downloaded latest OrientDB (community 2.1.9) and logged into studio to play around.
When I rename properties a lot of strange things happen. E.g. I renamed a property only changing the Case. Then I technically had the same property duplicated (name and Name) what I could see when I query data but in schema when I list properties only the renamed property (Name) is shown. The more I tried to fix this by dropping property and or renaming back the worse it got.
To me it seems that the feature to rename properties is currently broken and the handling of properties in different views of studio is inconsistent.

Anyhow, thanks for creating OrientDB. From a conceptual point of view it looks really cool and I like the first impression. I am planning to use it for an big open-source project and might consider it also in commercial projects. However, if things are buggy (and I just hit the surface) this is of course a showstopper... Keep going and let me know if you need further input how to reproduce the problem.

smolinari:
I can confirm. Renaming properties doesn't seem to work properly or rather as expected in Studio.

Working with 2.1.9.

Scott

hohwille:
As I am new to OrientDB I did some anaylsis and investigation. To sum up this is not a critical bug as there are simple workaround to get the intended result.

[...]

Issue #2674, opened 2014-08-11
drop property should allow to choice from schemaless data or delete data


tglman:
add to delete property the option for remove the data:
ex:

``` sql
drop property Person.name delete data
```

Issue #2670, opened 2014-08-11
Alter property allows expression for data refactor.


tglman:
Create a new property refactor operation for change property types from not directly castable types
Ex if i've a String field and I want to convert to a DATE field:

```
ALTER PROPERTY Person.birth TYPE DATE WITH date(birth,'yyyy-MM-dd')
```

this will create a temporary property where it'll copy all the converted value and at the and will swap the property names and delete the old values.

In case of convertion failure all the changes done should be reverted

PR #3802, opened 2015-03-24
Multiple Inheritance for OrientDB

PhantomYdn:
Guys,

This patch allow to have multiple super classes for a class. 
It's related to this Issue #2939  

To take a look into testcases please navigate to: https://github.com/PhantomYdn/orientdb/blob/multi-super-classes/core/src/test/java/com/orientechnologies/orient/core/metadata/schema/TestMultiSuperClasses.java

IMPORTANT: I understand - it's big change for OrientDB and many things can be impacted. Please fill free to start discussion there. 

P.S. There are a couple of things that should be fixed as well (for example class description and etc.). Those things will be commited into this branch as well.


tglman:
hi,
multiple inheritance have some design issue, like what happen if you extend two class that define the same property with two different types?  did you already solved this issue ?
same problem with all the settings(like validation) of the property.
a solution could be check all the extension tree for this problem and throw an error, but that have to be done for every change in hierarchy or property.

we have to take care also of the backward compatibility, as today we have to guarantee the current schema structure for older client, that's mean we can just add things. i think we should improve now the backward design to allow this changes in future.

lvca:
Hey @PhantomYdn, thanks so much for this PR. By a quick look seems very well done. I saw you also took care about retro compatibility. Maybe not everywhere, we should double check it.

@tglman questions about conflicts with names could be resolved by applying the following limitations:
- OClass.createProperty() -> will check if the field already exists on current and recursively all subclasses
- OClass.addSuperClass() -> will check that super class has no fields in conflict with current one

lvca:
@PhantomYdn I saw you removed some methods, like `OClass.setSuperClass()`. Actually we should preserve compatibility with code that runs on 2.0.x, so this would brake it. I'd suggest to don't remove any method, but just deprecate it and let's do what was supposed to do on 2.0. For example, the `OClass.setSuperClass()` method could set that class as only superclass: exactly what 2.0.x did. WDYT?

PhantomYdn:
@lvca I didn't change remove methods: setSuperClass() is there. Might be due to some reformatting it was moved.

Regarding properties conflict: I do especially use "List" for super classes. Idea is following: first super classes has priority over the rest. If there is "prop" field on A and B classes and we derived C from A&B: priority will be on those from A. But if you derive from B&A: priority will be on B.  Btw, this behaviour is not tested...

PhantomYdn:
And regarding backward compatibility: I kept that as well. The logic is following:
getSuperClass will return first of your super classes.
setSuperClass will replace whole set of your super classes by this one.

Also SQL changed. So know following are possible:

```sql
create class C extends A, B;
alter class C superclasses A, B;
alter class C superclass +A;   --Add A class
alter class C superclass -B;   --Remove B class
```

PhantomYdn:
And let me comment also this one:

```
OClass.createProperty() -> will check if the field already exists on current and recursively all subclasses
```

I think - it's not so useful behavior of OrientDB... If there are classes A and B and "B extends A": it will be really cool to be able to override some properties from A on B. What do you think? So: imho, check is not  required.

lvca:
@PhantomYdn The only concern is with binary protocol. @tglman WDYT about allowing redefinition of fields and using the first one found in a breadth-first algorithm starting from current class?

PhantomYdn:
Guys, please do not apply this PR once I will know.

But I have few things to discuss:
I found out that there are lots of code duplication between OClassImpl and OImmutableClass. To remove that duplication I have idea to introduce new class OAbstractClass. But with that class there are also problems: almost all fields in OImmutableClass are final - so it will be not so streight forward... WDYT?

Also registration of OClass from JavaClass take into account only direct super class: without interfaces and so on. Should we add registering of interfaces as well?

lvca:
@PhantomYdn Forget about OImmutableClass: fields should remain final for now, we could think about it later.

What do you means for "OClass from JavaClass"? The Object Database API?

PhantomYdn:
There are methods like createClass(Class<?> clazz) in OSchemaShared.

@lvca please let me know what do you think about this my comment? https://github.com/orientechnologies/orientdb/pull/3802#issuecomment-85715700

PhantomYdn:
Added few testcases: looks pretty stable.

Still working on moving from deprecated method getSuperClass in a set of OrientDB functionalities.

tglman:
the idea of override fields following the order of extension is good, but we have to provide a way to choose this order.

for binary compatibility we have a quite huge issue, the schema is stored as document, and is sent on the network as a document, so we have to guarantee that all the field present in the document of the 2.0.* version will be there for all the 2.*

one more thing is if we include this functionality now all the old clients could start to send data that is not valid, we should evaluate well if this is fine.

In my opinion we are doing well to add the basic structure for handle this now, but we should enable this feature only in 3.0

PhantomYdn:
@tglman actually there is no deletion of existing fields: just added a new one.
And from code perspective: case when there is only class with old structure also supported as well. Imho it should be OK. In any case: please refer me to some particular case: I will check. For now: all existing testcases are passing.

lvca:
@PhantomYdn the worse case is when you override a field changing type. Now the type is stored once in the schema, so it should execute a massive update changing the type. However this is already supported in some fashion by @tglman when you manipulate class properties. So, I'd like to see this feature in 2.1, but the risk that breaks something is quite high. WDYT @PhantomYdn about inserting some constraints now and then try to remove them? The constraints could be what I mentioned above, namely the fact you can't redefine a type. You could just put a constraint we could remove later. WDYT?

lvca:
@tglman will write some notes about impacts of this change.

PhantomYdn:
@lvca , @tglman , just let me know what exact cases should be checked... 

And btw, what's better from you point of view:
This
```java
public List<OClass> getSuperClasses();
public OClass setSuperClasses(List<OClass> superClasses);
```
or this:
```java
public OClass[] getSuperClasses();
public OClass setSuperClasses(OClass... superClasses);
```

I see that there is no common approach in the code...

tglman:
ok, we have a few things that already are not perfect that i would like to fix before get new features in:
`ALTER PROPERY AClass.aproperty set type STRING `
this should trigger the migration of data, this is done today, but it's not safe in case of shutdown or crash, data could be migrated partially.
plus this: https://github.com/orientechnologies/orientdb/issues/2670 and this: https://github.com/orientechnologies/orientdb/issues/2674, and the index management: https://github.com/orientechnologies/orientdb/issues/3629.
this is just what is missed now without the concept of multiple class, we have some design already there, some to do from scratch.

going to the main topic of this PR the issue i see are:
Conflicts in case of multiple definition of property:
` CREATE PROPERTY A.name STRING`
` CREATE PROPERTY B.name LONG`
`CREATE CLASS C extends A,B` // this make the property A.name win on B.name
` INSERT INTO C SET name = "something" ` so this is correct
`SELECT name FROM B ` // this return a STRING when i expect just a list of LONG this is wrong.

The solution for this is just not allow two property with the same name in the hierarchy, that's my suggested solution, this example case is just one wrong case of many that could happen.

for remote should be fine if we guarantee the old document structure, but this could cause some validation issue client side, but the validation should be triggered also server side so not a big issue, but we need to test it properly.

The reason of the double structure: OSchemaShared and OImmutableSchema is due to some partial refactoring, in long term we will remain only with OImmutableSchema so no need to do an OAbtractSchema.

The API i prefere ` public List<OClass> getSuperClasses();  `

PhantomYdn:
Thanks @tglman ! Now I got your concerns and fully agree... Let me add required check during class creation...

lvca:
Hey @PhantomYdn we'd like to release tomorrow 2.1-rc1. What's the status of your contrib?

PhantomYdn:
Thanks for the note!:) Will complete mentioned check today.

PhantomYdn:
@lvca Check is added a long with testcases for good and bad scenarious. Please consider to take into 2.1-rc1 for early adoption...

mattaylor:
How are hooks resolved on classes that extend from multiple parents? Is this even supported?

lvca:
Do you mean hooks that extend ODocumentHookAbstract? Yes, this is already supported by this PR because call `OClass.isSubClassOf()`.

mattaylor:
I mean if I have a class a, which extends class b and class c, and both class b and c have triggers associated with them, would any or all of these hooks be.applied.to.instances of the subclass a?

lvca:
Yes

PhantomYdn:
Guys, any progress on this? If you see something to do: just let me know.

mattaylor:
2.1?

mattaylor:
Is this pr going to make 2.1 rc1?

lvca:
@PhantomYdn I'm trying to merge it manually with last "develop" but remote tests won't pass. Any clue on this?

PhantomYdn:
That's strange. It passed for me. Are you refering to "mvn test" or "ant test" testcases? Can you point me some particular unit test class?

lvca:
ant clean test

the second phase, the remote tests, won't pass. I did also:

    git merge develop

first.

PhantomYdn:
Yea - there are some problems with tests appear. I need a day to fix.

lvca:
Great, thank you @PhantomYdn. I'll move forward the release day of 2.1-rc1 because this feature is a badass ;-)

lvca:
Great, Merged! Thanks @PhantomYdn !

PR #3824, opened 2015-03-27
Fixing bug when using a token issued for a user in subclass of OUser

stuartking:
Pull request into develop as requested by @lvca on #3821 - first PR for project :)
-----
If a user record is contained with a subclass of OUser, then an authorization token can be issued however a NullPointerException is returned when an attempt is made to use it.

This is down to a condition in the JsonWebToken class which only returns an OUser object if the RID contained in the token is in the OUser class itself and doesn't take into consideration classes that may extend OUser. This pull request addresses this issue.


lvca:
Thanks for the PR!

Issue #3846, opened 2015-04-01
Count is broken when adding a vertex when count is used in the same transaction


wouterv:
Related to issue #3807

When adding a vertex, OrientBaseGraph.countVertices gives the total before the vertex is added.

New unit test code to reproduce:

```
package com.tinkerpop.blueprints.impls.orient;

import org.junit.After;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.JUnit4;

import static org.junit.Assert.assertEquals;

@RunWith(JUnit4.class)
public class OrientCountVertexTest {

  @After
  public void tearDown() {
    OrientGraph g = createGraph();
    g.drop();
  }

  @Test
  public void countVertexShouldWorkWithinOrOutsideTransactions() {

    // Create a node v1 with at least two edges
    OrientGraph g = createGraph();

    assertEquals(0, g.countVertices());
    g.addVertex("class:V1");
    g.addVertex("class:V2");
    g.addVertex("class:V2");

    g.shutdown();

    g = createGraph();

    long allCount = g.countVertices();
    assertEquals(3, allCount);
    long v1Count = g.countVertices("V1");
    assertEquals(1, v1Count);
    assertEquals(2, g.countVertices("V2"));


    g.addVertex("class:V1");
    assertEquals(allCount + 1, g.countVertices());
    assertEquals(v1Count + 1, g.countVertices("V1"));

    g.shutdown();
  }


  private OrientGraph createGraph() {
    return (OrientGraph) new OrientGraph("memory:" + OrientCountVertexTest.class.getSimpleName()).setWarnOnForceClosingTx(false);
  }

}

```

wouterv:
I've seen what you did to fix the other issue, so I patched it on my machine the same way and that fixed it, so I will create a pull request

lvca:
Thanks.

PR #3847, opened 2015-04-01
Fix for #3846. Count was broken when counting new vertices in same transaction

wouterv:
Just added upon your earlier change.


lvca:
Thanks!

lvca:
Sorry, I forgot to port this to 2.1.x and develop.

Issue #3862, opened 2015-04-02
cannot unlock a never acquired lock - distributed dserver


anmoldeep0123:
Issue : 

In distributed mode , running a query in locked mode in a multi-threaded environment , fails with the following exception.

OCommandExecutorSQLUpdate#209

    if (lockStrategy.equals("RECORD"))
      query.getContext().setVariable("$locking", OStorage.LOCKING_STRATEGY.KEEP_EXCLUSIVE_LOCK);

applies the lock ... 

and 

OTransactionAbstract 
 final ORID rid = iRecord.getIdentity();

    final OStorage.LOCKING_STRATEGY lock = locks.remove(rid);

    if (lock == null)
      throw new OLockException("Cannot unlock a never acquired lock");

removes the lock and seems like the hashmap.remove returns null , so either it was not populated with the lock or was removed with the lock at some earlier stage.

Steps to recreate  -
1. create class osequence extends V
2. create vertex osequence set classname='Test',value=1
3. run the query on the studio in distributed mode -
update #rid:clusterid increment value = 1 RETURN after @this lock record
run and there you go the exception

Exception trace -
Apr 02, 2015 5:47:08 PM com.orientechnologies.common.log.OLogManager log
INFO: OrientDB auto-config DISKCACHE=894MB (heap=841MB os=3,783MB disk=395,410MB)
#11:24
HERE myThread - 0
Exception in thread "myThread - 0" com.orientechnologies.orient.server.distributed.ODistributedException: Error on execution distributed COMMAND
	at com.orientechnologies.orient.server.distributed.ODistributedStorage.command(ODistributedStorage.java:397)
	at com.orientechnologies.orient.core.command.OCommandRequestTextAbstract.execute(OCommandRequestTextAbstract.java:63)
	at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.command(ONetworkProtocolBinary.java:1178)
	at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.executeRequest(ONetworkProtocolBinary.java:385)
	at com.orientechnologies.orient.server.network.protocol.binary.OBinaryNetworkProtocolAbstract.execute(OBinaryNetworkProtocolAbstract.java:220)
	at com.orientechnologies.common.thread.OSoftThread.run(OSoftThread.java:69)
Caused by: com.orientechnologies.common.concur.lock.OLockException: Cannot unlock a never acquired lock
[...]

PR #3870, opened 2015-04-04
Fix for #3862

anmoldeep0123:
Fix : 

While unlocking the record we are not checking whether the record is locked or not , we are just called a unlock in finally block in /OCommandExecutorSQLSelect.java , adding a new method to interfaces OIdentifiable.java
 and  
OTransaction.java

Tested for ant clean install 
ant clean test


lvca:
Perfect, thanks!

Issue #3887, opened 2015-04-07
Limitate the characters of the Class name to an alphanumeric set


tglman:
Today is possible to create class with the dot '.' in the name or with some not Latin Characters that create problem in query parsing.

lvca:
Duplicated of #3879

PR #3896, opened 2015-04-09
Fix for #3887

anmoldeep0123:
Regex for a valid class :
"^[_\\p{L}][_\\p{L}\\p{N}]*";

starts with underscore / UNICODE letter  followed by 0 or more (underscore or UNICODE letters or digits) 

if pattern fails to match , throw an exception as code was doing previously

tested for 

ant clean test 
ant clean install


lvca:
Great, thanks!

lvca:
Maven tests doesn't pass:

    mvn clean test

Reverting it.

PhantomYdn:
Imho it's better to use old approach: it's much faster than regexp.

lvca:
I already included the check for '.' char in last snapshots.

PR #3912, opened 2015-04-10
Aggregative concat function with ability to specify separator

PhantomYdn:
Possible usages:

```sql
--There are 3 objects with names = a, b and c
SELECT concat(name) from MyClass order by name;  --returns 'abc'
SELECT concat(name, ', ') from MyClass order by name;  --returns 'a, b, c'
```


lvca:
Cool function, thanks!

vitorenesduarte:
Hi @PhantomYdn. You have an __order by__ in those queries but they have no effect right?
To actually order the concat inputs I'll have to, e.g.:
```sql
select concat(name) from (select name from ConCat order by name desc)
```

PhantomYdn:
Hmm.. interesting  - it works for me, but that might be due to native sort is the same. Will check.

Issue #3604, opened 2015-02-17
Schema generation for reference to Serializable classes makes a CUSTOM type property instead of LINK


wouterv:
Lets say I have the following classes:

```
public class Foo implements Serializable{
//ids, fields, getters/setters etc
}

public class Bar {
   private Foo foo;
   //ids, getters/setters etc
}
```

When I let the schema be generated, the type of Bar.foo becomes CUSTOM and any data becomes an embedded object. 
What I would except is a LINK. 
When removing Serializable from Foo, it creates a LINK.

The culprit is in OType:
```
  CUSTOM("Custom", 20, OSerializableStream.class, new Class<?>[] { OSerializableStream.class, Serializable.class }),
```

So whenever a class implements Serializable it becomes CUSTOM, while it is more logical to make it a LINK.

Orient version 2.0.2

PR #3924, opened 2015-04-14
Fix schema generation for serializable fields that point to other orient entity/schema classes

wouterv:


wouterv:
Fixed #3604

lvca:
It's a shame I was already releasing 2.0.7 :-( Thanks, this will be part of 2.0.8 and 2.1

lvca:
Thanks

Issue #3684, opened 2015-03-03
OrientDB ORecordHookAbstract is firing onRecordAfterRead twice if no index is present


ardias:
I'm using the hook functionality in OrientDB to implement encryption of some document fields "transparently" to the caller. Basically the process is:

1 : when onRecordBeforeCreate or onRecordBeforeUpdate event fires we apply encryption to some data fields and change the document before is created or updated

    byte[] data = doc.field("data");
    byte[] encrypted = encrypt(data);
    doc.field("data", encrypted);

2 : when onRecordAfterRead fires we get the encrypted data from the document fields, decrypt them, and change the document fields again with the decrypted data.

    byte[] encrypted = doc.field("data");
    byte[] decrypted = decrypt(encrypted);
    doc.field("data", decrypted);

The problem is that the event onRecordAfterRead is firing twice and in the first time the data decrypts correctly (because is encrypted) but on the second time the decryption fails because we already had decrypted it, and so the document "load" fails.

It happens if the query that i execute to load the document uses some field of the document in the filter (where clause). So for example, the following query does not trigger the problem:

    select count from Data;

but the following query triggers the problem:

    select from Data where status ="processed";

This is because i don't have an index on the status field. If i add an index then it fires the event only once. So this is related with the use or not of the indexes. Should the events be fired if OrientDB is "scanning" the documents when executing a query? Shouldn't it only fire when the matching documents are in fact loaded? Is there a way around this?

It's very easy to reproduce: Create a database, register a hook, save a document, and query the database by some field with an index and a fields without an index. You will notice that onRecordAfterRead will be raised once for the query with the index and twice for the query without the index

ardias:
You can check a sample that reproduces the bug here: https://github.com/ardias/snippets

tglman:
fixed there was a wrong call.

lvca:
Great.

lvca:
Can you back port it in 2.0.5?

ardias:
Thanks! It would be really great if you could port it back 2.0.x

tglman:
ported.

ardias:
Tks, that's great!

PR #3942, opened 2015-04-17
Fix DESC iterator for classes with multiple clusters

mcculls:
The change made in 18740e01cf46c882053aa542bd09e1ed38d7d1bb for #3684 has indirectly led to a bug in `ORecordIteratorClassDescendentOrder` when dealing with classes that have multiple clusters.

Before #3684 the `begin` method was always called in the `config` method of `ORecordIteratorClass`. Since `ORecordIteratorClassDescendentOrder` overrides `begin` to instead call `last` this meant that the `currentClusterIdx` was correctly set to the last cluster before the `setRange` call.

After #3684 the `begin` method is only called for the shorthand constructors - the longer form which takes a locking strategy doesn't call `begin`. Instead it expects that the `setRange` method will subsequently be called - and that this will call `begin` and complete initialization of the iterator, as done in:

https://github.com/orientechnologies/orientdb/blob/develop/core/src/main/java/com/orientechnologies/orient/core/sql/OCommandExecutorSQLResultsetAbstract.java#L447

Unfortunately because `begin` is only called at the end of the `setRange` method the `currentClusterIdx` will be left set to the first cluster (0) when we first enter `setRange`, which causes all sorts of interesting issues depending on the number of clusters attached to the class and their relative bounds. For example the attached test will hang using the latest code in the develop branch.

The simplest solution is to ensure that `currentClusterIdx` is always set to the last cluster at the end of the `ORecordIteratorClassDescendentOrder` constructor which then means that the right bounds will be checked in the `setRange` method.


lvca:
Makes sense. Thanks for this PR and for the test case. I'll merge it asap.

lvca:
Thanks!

PR #3965, opened 2015-04-19
GraphSON

Blondysinger:
New Interface IGraphSONIterable

The change allows GraphSONUtility to use any kind of type that implements IGraphSONIterable. In this way not only List, Map, Array or Element, but all the wrapper that implements IGraphSONIterable, such as OrientElementIterable, will return a valid value from the jsonFromElement method.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,6 @@ out
 *.iml
 .idea/
 
+*.DS_Store
+
 tests/src/test/java/com/orientechnologies/orient/test/database/auto/_*.xml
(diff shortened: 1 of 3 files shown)


lvca:
How can GraphSONUtility use this interface if it's brand new ?

Blondysinger:
Yes, sorry! You are right!
I've also modified blueprints libs in the method createJSONMap (GraphSONUtility) so that it too can recognize the interface.
Therefore you can benefit from this feature only if this interface will be accepted via the PR I've made in tinkerpop.

Blondysinger:
I have modified the PR. OrientIterable doesn't need any change.

lvca:
I don't see any changes. Can I close it?

PR #3983, opened 2015-04-21
Fixes #3982

### LINKED ISSUE #3982: including return { f: $val } in SQL script does not return a record with key / value pairs using binary protocol
When using the binary protocol, the following SQL script:

```
let res = INSERT INTO V SET name = 'bill' RETURN @rid
return { result: $res }
```

The expected result should be a record type with a key of `result` and a value of type `RID`, instead it simply returns a list of RIDs

However, using the HTTP protocol does return the expected result.

stuartcarnie:
Ideally, OCommandExecutorScript would return ODocument, so there
was no special casing required in either ONetworkProtocolBinary of the
HTTP protocol. I am not familiar enough with the reasoning, so I emulated
 the behavior from the HTTP protocol

### DIFF
--- a/server/src/main/java/com/orientechnologies/orient/server/network/protocol/binary/ONetworkProtocolBinary.java
+++ b/server/src/main/java/com/orientechnologies/orient/server/network/protocol/binary/ONetworkProtocolBinary.java
@@ -23,13 +23,8 @@ import java.io.IOException;
 import java.io.ObjectOutputStream;
 import java.net.Socket;
 import java.net.SocketException;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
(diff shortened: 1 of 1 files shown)


stuartcarnie:
This logic was duplicated from the [HTTP protocol implementation](https://github.com/orientechnologies/orientdb/blob/master/server/src/main/java/com/orientechnologies/orient/server/network/protocol/http/OHttpResponse.java#L202-L208)

stuartcarnie:
@lvca this is the pull request :)

lvca:
@stuartcarnie You're right. I opened it from another URL and I cannot see as is. Thanks!

lvca:
Thanks!

tglman:
is missing the test case, would be cool get another pull with a proof of use case.

PR #3998, opened 2015-04-22
Bug fixes: @Transient fix on inherited fields, backup file names with spaces

e-d:
I feel the community would benefit from both of these simple fixes that address critical issues we've run into, so I'm creating a pull request.


lvca:
@e-d Thanks for the PR. While the first is clear, what's the purpose of the second one?. Another thing is using the branch "develop". We use "master" only for releases.

lvca:
Please could you resent it against a "develop" aligned branch?

PR #3999, opened 2015-04-22
1.7.11

matanshukry:
The following are two sections, each for the documentation regarding the 2 new commands,
create user and drop user. They are simply followed one by another in this page, but they are supposed to be separated and put each in his own file.

# SQL - CREATE USER 

Creates a user in the current database, with the specified password and an optional role.
Note that if a role isn't specified, a default role ("writer") is used.

Note this is simply a wrapper around OUser and ORole tables; Look up [Security](Security) for more details.

## Syntax

```sql
CREATE USER <user name> IDENTIFIED BY <password> [ROLE <role name>]
```

### Creates a new user called 'Foo', with password 'bar', and role 'admin'

```sql
create user Foo identified by bar role admin
```

### Creates a new user called 'Bar', with password 'Foo'
```sql
crete user Bar identified by Foo
```

To know more about other SQL commands look at [SQL commands](SQL).

# SQL - DROP USER 

Drops a user from the current database.

Note this is simply a wrapper around OUser table; Look up [Security](Security) for more details.

## Syntax

```sql
DROP USER <user name>
```

### Drop a user called 'Foo'

```sql
drop user Foo
```

To know more about other SQL commands look at [SQL commands](SQL).

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -5,7 +5,7 @@
     <property environment="env"/>
     <property name="vendor" value="Orient Technologies Ltd"/>
     <property name="product" value="OrientDB"/>
-    <property name="version" value="1.7.10"/>
+    <property name="version" value="1.7.11-SNAPSHOT"/>
     <condition property="community.release" value="${releaseHome}/orientdb-community-${version}"
                else="../releases/orientdb-community-${version}">
         <isset property="releaseHome"/>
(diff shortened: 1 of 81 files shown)


matanshukry:
Wrong branch, closing this and opening a correct one.

PR #4000, opened 2015-04-22
SQL Commands - Create/Drop User

matanshukry:
The following are two sections, each for the documentation regarding the 2 new commands,
create user and drop user. They are simply followed one by another in this page, but they are supposed to be separated and put each in his own file.

# SQL - CREATE USER 

Creates a user in the current database, with the specified password and an optional role.
Note that if a role isn't specified, a default role ("writer") is used.

Note this is simply a wrapper around OUser and ORole tables; Look up [Security](Security) for more details.

## Syntax

```sql
CREATE USER <user name> IDENTIFIED BY <password> [ROLE <role name>]
```

### Creates a new user called 'Foo', with password 'bar', and role 'admin'

```sql
create user Foo identified by bar role admin
```

### Creates a new user called 'Bar', with password 'Foo'
```sql
crete user Bar identified by Foo
```

To know more about other SQL commands look at [SQL commands](SQL).

# SQL - DROP USER 

Drops a user from the current database.

Note this is simply a wrapper around OUser table; Look up [Security](Security) for more details.

## Syntax

```sql
DROP USER <user name>
```

### Drop a user called 'Foo'

```sql
drop user Foo
```

To know more about other SQL commands look at [SQL commands](SQL).


lvca:
@matanshukry Thanks for this PR! We're in RC phase, I'll merge it for sure as soon as we'll switch to 2.2-SNAPSHOT.

matanshukry:
lvca - There seem to be more conflicts. Please comment again when the project has moved to 2.2-SNAPSHOT, I'll fix it and push again then (it will send me an email notification).

lvca:
@matanshukry Now develop is 2.2-SNAPSHOT ;-) About this what do you think on supporting multiple roles syntax like:

    CREATE USER <user name> IDENTIFIED BY <password> [ROLES <role name>, <role name>*]

Example:

    CREATE USER jay IDENTIFIED BY miner ROLES genius, creative

or

    CREATE USER <user name> IDENTIFIED BY <password> [ROLE [<role name>, <role name>*] ]

Example:

    CREATE USER jay IDENTIFIED BY miner ROLE [genius, creative]

matanshukry:
I stayed with a single keyword, seems more consistent.
You can either do ```ROLE [admin, reader, etc]``` or ```ROLE admin```

Also, merged again with develop.

lvca:
Thanks!

lvca:
I saw test cases are missing, I'm going to write them.

Issue #89, opened 2012-12-10
Encryption of Data at rest [moved]


lvca:
I would like to use OrientDB for a healthcare application in the U.S. One of the requirements is that the data be encrypted at rest. This could probably be done through writing drivers for a particular OS. We are currently using MySQL and have a third party driver to accomplish encryption at rest. It would be great if OrientDB had a built in capability to do this.

mandrachek:
@lvca this should probably be re-opened. This is not about encryption-on-the-wire (SSL), but rather storage encryption for data at rest. (e.g., on disk, or perhaps even in memory). See the mongodb docs @ http://docs.mongodb.org/manual/core/security-introduction/ - "To encrypt document or field level data, write custom encryption and decryption routines or use a commercial solutions such as the Vormetric Data Security Platform."  I think that's more or less what this particular ticket is after.

edwardkay:
@lvca I agree with @mandrachek - this issue should probably be reopened.

We're using Baasbox (http://www.baasbox.com/), which incorporates OrientDB, on a project. The stipulated ISMS requirements from our client at that data is encrypted at rest.

This is different from either encryption on the wire (SSL) and filesystem encryption. See http://www.teamshatter.com/topics/general/team-shatter-exclusive/encrypting-data-at-rest/ for some further reasons why it's important.

What do you think?

lvca:
Reopened. Guys, this is quite easy to support but we'd need a few days. Any chance to receive a sponsorship to develop it?

edwardkay:
Thanks @lvca. Would love to offer sponsorship if we could but sadly there's no capacity in our project budget :(

giastfader:
@lvca we are interested in this and depending by the necessary effort, we could develop it, if you will give us the right advices.
Please drop me an email....

lvca:
@giastfader it would be great: a new feature by BaasBox ;-)
You can write an implementation of com.orientechnologies.orient.core.compression.OCompression interface and register into `OCompressionFactory` singleton. Example, you call your implementation as "encrypted". In order to activate it on a new cluster use:

    ALTER CLUSTER MyCluster COMPRESSION encrypted

We store the database compression inside storage configuration, but we never exposed it on ALTER DATABASE command.

giastfader:
Where do we store the key?

edwardkay:
@giastfader Thanks for your reply. As @lvca says, direct BaasBox contributions to OrientDB would be great and is very encouraging from our point of view.

[...]

PR #4018, opened 2015-04-24
Issue 89, AES and DES encryption at rest

giastfader:
This PR is related to https://github.com/orientechnologies/orientdb/issues/89
Here are two new cluster compression methods that use AES and DES to encrypt data.
Keys have to be declared into the configuration file, or in any other way supported by OrientDB (command line, direct set on OGlobalConfiguration enumerator).
There is of course room for improvements, consider this as a first WORKING implementation.

usage:
```
create class aes
alter cluster aes compression aes-encrypted

create class des
 alter cluster des compression des-encrypted
```


lvca:
We're in RC phase, I'll merge this as soon as we start the 2.2-SNAPSHOT.

lvca:
Hey @giastfader I was merging this in develop, but unfortunately you didn't sign the clahub agreement (for the Apache2 license). Please could you do that? Thanks in advance.

lvca:
Waiting for you to electronic signing the CLA, I saw you use 2 new global settings:
```java
STORAGE_ENCRYPTION_DES_KEY("encryption.des_key","The simmetric key to use to encrypt/descript data at rest using the DES alghorithm, stored in BASE64. The key must be 64 bits long. Default is \"T1JJRU5UREI=\" (ORIENTDB).",String.class,"T1JJRU5UREI="),
 STORAGE_ENCRYPTION_AES_KEY("encryption.aes_key","The simmetric key to use to encrypt/descript data at rest using the AES alghorithm, stored in BASE64. The key must be 128 or 256 bits. Default is \"T1JJRU5UREJfSVNfQ09PTA==\" (ORIENTDB_IS_COOL).",String.class,"T1JJRU5UREJfSVNfQ09PTA=="),
```

I think it would be much better don't setting any password by default and check if password is null, an exception is raised. In this way the user can't do mistakes. WDYT?

giastfader:
Yes, I agree, if you want to force the admin to explicitly set a key.

lvca:
Cool, I'm merging it right now

lvca:
Done, thanks!

Issue #4032, opened 2015-04-26
Add support for FetchType.LAZY on @OneToOne relations 


wouterv:
We have some object graphs that when fully loaded from orient can be quite large. When running attached mode, this is no problem, because we can steer the loading using fetch plans.

However when detaching using detachAll(object, true), the whole object graph will be loaded. I propose that adding @OneToOne(fetch = FetchType.LAZY) will have an detachAll skip those fields, inserting only a placeholder object with the id inserted, so the user can manually load the record when needed. So that would not be the responsibility of orient, as the object inserted is not a proxy.

lvca:
Hey @wouterv is this already in your merged PR?

wouterv:
Yes it is.

PR #4033, opened 2015-04-26
Support for one-to-one lazy fetch on detachAll. Issue #4032

wouterv:
This adds support for @OneToOne(lazy=FetchType.LAZY) when calling detachAll(object, true).

I would be happy to add info to the manual, but I couldn't find how to contribute to that.


lvca:
Cool PR, thanks! However this is a new feature and shouldn't go as hotfix (2.0.x) but rather in 2.1 in "develop" branch. Please could you resynch it against "develop" branch and send a PR again? Thanks in advanc.

wouterv:
New one: #4061

PR #4039, opened 2015-04-27
Fix for found record in transaction with the same RID warning. 

### LINKED ISSUE #4038: Warning: Found record in transaction with the same RID #2132:5 but different instance.
I get this warning a lot when loading via the Object API multiple times during a transaction in our code:

WARNING: {db=test} Found record in transaction with the same RID #2132:5 but different instance. Probably the record has been loaded from another transaction and reused on the current one: reload it from current transaction before to update or delete it

However somehow I cannot recreate this with a unit test in the orient source, but I did manage to find the problem and will provide a pull request with the fix.

wouterv:
Fixes #4038. This removed the warning for us and everything worked just fine as it did before.


lvca:
Thanks!

PR #4040, opened 2015-04-27
Email alert for server down

martingg88:
can we have an event to alert system admin if sever down?


lvca:
@martingg88 I can't see your changes because 152 files were changed. Furthermore we use "master" branch only for releases while all new implementations go to "develop" branch. Please could you rebase your branch against "develop" and send a new PR to be reviewed?

PR #4054, opened 2015-04-29
Allow backups for file paths with spaces and fix @Transient inherited fields

e-d:
1. You can now run a backup to a file that has one or more spaces in the file path.
2. There is a bug with inheriting @Transient fields where the field loses it's transient property in the subclass. For example, given the following classes if you create an Apple and save it to the database, it will persist both sweetness and color.

class Fruit {
  @Transient protected String color;
}

class Apple extends Fruit {
  private int sweetness;
}

This becomes an important issue because we can't use the transient Java keyword because we still need to be able to serialize these objects to send them over the network, we just don't want them to be persisted. That is why, I assume, OrientDB went to the trouble to support the JPA @Transient annotation. This small one variable change fixes the inheritance of these transient fields.


lvca:
Thanks!

tglman:
would be cool also have a test case to prove the fix ;)

Issue #4060, opened 2015-04-30
Cannot create class with multiple inheritance.


maggiolo00:
I got this error but the class is created without superclasses.

{"errors":[{"code":500,"reason":500,"content":"com.orientechnologies.orient.core.exception.OSchemaException: Can't add base class 'Test45', because of recursion"}]}

To Replicate

```
alter class V superclasses ORestricted
create class Dummy1 extends V,ORestricted
create class Dummy2
create class Dummy3 extends Dummy1, Dummy2
```

The class Dummy3 is created with no superclasses

PhantomYdn:
Thank you for the testcase! Fixed in PR #4062 . As a workaround for 2.1-rc1 you can use the following:

```sql
alter class V superclasses ORestricted
create class Dummy1 extends V
create class Dummy2
create class Dummy3 extends Dummy1, Dummy2
```

tglman:
merged the PR closing

PR #4062, opened 2015-04-30
Fix issue #4060: wrong recursion detection

PhantomYdn:
Issue #4060


tglman:
seems all right merging, thanks for the quick fix :)

Issue #4202, opened 2015-05-21
Can not create UNIQUE/NONUNIQUE index in distributed mode 


dblaber:
In  #4177 , the issue with distributed+hash index is fixed, however it effects more than just hash index (or it is another issue). This is orientdb 2.1 rc2, no hashindex. As of right now, it seems there is no way to use a index in distributed mode. 
I am creating a Schema with Unique Index and NOTUNIQUE index and am receiving this exception:
Exception in thread "main" com.orientechnologies.orient.core.exception.OTransactionException: Error on committing distributed transaction
	at com.orientechnologies.orient.server.distributed.ODistributedStorage.commit(ODistributedStorage.java:967)
	at com.orientechnologies.orient.core.tx.OTransactionOptimistic.doCommit(OTransactionOptimistic.java:491)
	at com.orientechnologies.orient.core.tx.OTransactionOptimistic.commit(OTransactionOptimistic.java:148)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.commit(ODatabaseDocumentTx.java:2412)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.commit(ODatabaseDocumentTx.java:2382)
	at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.commit(ONetworkProtocolBinary.java:1074)
	at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.executeRequest(ONetworkProtocolBinary.java:376)
	at com.orientechnologies.orient.server.network.protocol.binary.OBinaryNetworkProtocolAbstract.execute(OBinaryNetworkProtocolAbstract.java:221)
	at com.orientechnologies.common.thread.OSoftThread.run(OSoftThread.java:71)
Caused by: com.orientechnologies.orient.core.index.OIndexException: Error during insertion of key in index
	at com.orientechnologies.orient.core.index.OIndexMultiValues.put(OIndexMultiValues.java:141)
	at com.orientechnologies.orient.core.index.OIndexMultiValues.put(OIndexMultiValues.java:52)
	at com.orientechnologies.orient.core.index.OIndexAbstract.putInSnapshot(OIndexAbstract.java:871)
	at com.orientechnologies.orient.core.index.OIndexAbstract.applyIndexTxEntry(OIndexAbstract.java:1051)
[...]

PR #4205, opened 2015-05-21
Fix Possibly #4202 in distributed mode

dblaber:
Made this local change, worked for me.


lvca:
The fix makes sense. Unfortunately we use "master" only to release. Could you do that against "develop" or "2.0.x" ? Remember to synch your repository with the branch before to send it.

lvca:
No problem, I can do that.

Issue #4214, opened 2015-05-23
BROWSE CLASS throws a NullPointerException on fresh install


wayspurrchen:
Hi, I've been following along with the tutorial on the OrientDB website and I'm at the [Classes](http://orientdb.com/docs/last/Tutorial-Classes.html) step. Trying to use `browse class [class]` against any of the classes in the GratefulDeadConcerts test database on a new community release (orientdb-community-2.0.9) throws a NullPointerException error. I can't seem to figure out why. I'm using Mac OS X 10.10.

```
OrientDB console v.2.0.9 (build UNKNOWN@r; 2015-05-14 21:57:10+0000) www.orientechnologies.com
Type 'help' to display all the supported commands.
Installing extensions for GREMLIN language v.2.6.0

orientdb> connect remote:localhost/GratefulDeadConcerts root [password]

Connecting to database [remote:localhost/GratefulDeadConcerts] with user 'root'...OK

orientdb {db=GratefulDeadConcerts}> classes


CLASSES
----------------------------------------------+------------------------------------+------------+----------------+
 NAME                                         | SUPERCLASS                         | CLUSTERS   | RECORDS        |
----------------------------------------------+------------------------------------+------------+----------------+
 E                                            |                                    | 10         |              0 |
 followed_by                                  | E                                  | 11         |           7047 |
 OFunction                                    |                                    | 7          |              0 |
 OIdentity                                    |                                    | -          |              0 |
 ORestricted                                  |                                    | -          |              0 |
 ORIDs                                        |                                    | 6          |              0 |
 ORole                                        | OIdentity                          | 4          |              3 |
 OSchedule                                    |                                    | 8          |              0 |
 OTriggered                                   |                                    | -          |              0 |
 OUser                                        | OIdentity                          | 5          |              3 |
 V                                            |                                    | 9          |            809 |
----------------------------------------------+------------------------------------+------------+----------------+
 TOTAL = 11                                                                                                 7862 |
----------------------------------------------+------------------------------------+------------+----------------+
orientdb {db=GratefulDeadConcerts}> browse class V

Error: java.lang.NullPointerException

orientdb {db=GratefulDeadConcerts}> browse class E

Error: java.lang.NullPointerException

orientdb {db=GratefulDeadConcerts}> browse class ORole

Error: java.lang.NullPointerException
```

ManiacalAce:
I'm facing this as well - I was using the 2.1rc3 version.

luigidellaquila:
[...]

PR #4216, opened 2015-05-24
fix for #4214

fabriziofortino:
I have modified just two lines (2280, 2283) but the file looks entirely changed.


lvca:
You're right: everything seems changed. To be more safe I merged your 2 lines manually. Thanks!

PR #4238, opened 2015-05-28
NPE on GET request to /server

matanshukry:
# The Problem

When using a minimalist configuration file ( orientdb-server-config.xml ), and trying to query server information using the http protocol, and the path "/server", the result is a null pointer expection message in the server log:

```
2015-05-28 23:48:39:375 SEVERE Internal server error:
java.lang.NullPointerException [ONetworkProtocolHttpDb]
```

and an internal server error for the client.

# Cause
The error is caused when trying to iterate over a null array of OServerEntryConfiguration .

# The Solution
Simply check if the array exists before iterating over it


lvca:
Thanks a lot!

PR #4296, opened 2015-06-04
REST - CreateDatabase - AlreadyExists returns 500

matanshukry:
When trying to create a new database that already exists,
the result is http status code 500.

The result for a specific checked case should be specific,
rather than the generalized 500 code which should be used for runtime exception.

The code and the text can obviously be changed to anything.


lvca:
I like, it: +1.

Issue #4303, opened 2015-06-05
java.util.ConcurrentModificationException on commit in remote for document with LINKSET


wouterv:
When saving a document containing a LINKSET with 2 new other documents, the server throws an exception on commit:

````
{db=beeyondmemory} Sent run-time exception to the client /127.0.0.1:53164: com.orientechnologies.orient.core.exception.OStorageException: Error during transaction commit.
com.orientechnologies.orient.core.exception.OStorageException: Error during transaction commit.
	at com.orientechnologies.orient.core.storage.impl.local.OAbstractPaginatedStorage.commit(OAbstractPaginatedStorage.java:941)
	at com.orientechnologies.orient.core.tx.OTransactionOptimistic.doCommit(OTransactionOptimistic.java:488)
	at com.orientechnologies.orient.core.tx.OTransactionOptimistic.commit(OTransactionOptimistic.java:147)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.commit(ODatabaseDocumentTx.java:2427)
	at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.commit(ODatabaseDocumentTx.java:2397)
	at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.commit(ONetworkProtocolBinary.java:1088)
	at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.executeRequest(ONetworkProtocolBinary.java:390)
	at com.orientechnologies.orient.server.network.protocol.binary.OBinaryNetworkProtocolAbstract.execute(OBinaryNetworkProtocolAbstract.java:217)
	at com.orientechnologies.common.thread.OSoftThread.run(OSoftThread.java:69)
Caused by: java.util.ConcurrentModificationException
	at java.util.HashMap$HashIterator.nextEntry(HashMap.java:922)
	at java.util.HashMap$EntryIterator.next(HashMap.java:962)
	at java.util.HashMap$EntryIterator.next(HashMap.java:960)
	at com.orientechnologies.orient.core.db.record.ORecordLazySet$1.next(ORecordLazySet.java:98)
	at com.orientechnologies.orient.core.db.record.ORecordLazySet$1.next(ORecordLazySet.java:75)
	at com.orientechnologies.orient.core.db.record.OLazyRecordIterator.next(OLazyRecordIterator.java:68)
[...]

PR #4321, opened 2015-06-08
Fix for 4303, fix for records in lazy sets with ids that are not persisted

wouterv:
I couldnt reproduce this with a unit test, but with our own code I could see that in, the ORecordLazySet.add method, a temporary id was put in the set, that was later changed by another txEntry in the same commit, but the set wasnt updated, because no change listener was recorded.

So this fix also checks for those kind of elements in the set. This fix works for us and the orient build runs nicely.


wouterv:
Fixes #4303

lvca:
@tglman WDYT about this PR?

tglman:
the test pass i'm going to merge it and cherry-pick it to develop,

@wouterv can you provide a pull request that prove the fix ? 

thank you

wouterv:
@tglman by "prove the fix" you mean a unit test or something? I tried to recreate it with a unit test, but couldn't, I got this with our running -very complex- application and debugged a running orient server, that is how I found the problem and the fix.

PR #4359, opened 2015-06-13
One fix for "com.orientechnologies.orient.core.exception.OSerializationException: Type String cannot be serialized because is not part of registered entities. To fix this error register this class"

aheintz:
I'm frequently running into the error "com.orientechnologies.orient.core.exception.OSerializationException: Type String cannot be serialized because is not part of registered entities. To fix this error register this class" which is caused by a OObjectLazyMap being created when the map only contains java types. I have found that the cause for the errors occurring to me is that an OObjectLazyMap is created when the fields have a lower version number than the document version number. It's likely that the application is badly coded, I'm however porting it from a working JPA implementation.

The fix is by adding a !OReflectionHelper.isJavaType(genericMultiValueType) check which solves the problem (my problem at least).

Had a few issues with my lack of experience of creating pull requests (I'm old, this new fancy git stuff :-) and the code formatter (which by the way reformats recently cloned code) so there's a few formatting issues. The unit test is ugly, but I really haven't time to create a really nicely documented one, it do what it should, proves the error and my change).

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/conflict/OAutoMergeRecordConflictStrategy.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/conflict/OAutoMergeRecordConflictStrategy.java
@@ -20,11 +20,15 @@
 
 package com.orientechnologies.orient.core.conflict;
 
+import com.orientechnologies.orient.core.db.ODatabaseDocumentInternal;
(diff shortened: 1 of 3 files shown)


tglman:
nice pull request, checking it

tglman:
look good and test passes, merging.

PR #4364, opened 2015-06-15
Uninstall signal handlers on shutdown

mcculls:
To help with embedding...


tglman:
it make sense, i run test with it and i had no problem merging

Issue #4285, opened 2015-06-03
cannot create index of linkset on distributed mode


henryzhao81:
tried on 2.0.11-SNAPSHOT
I have a class called user contains a field called roles which type is a linkset and link to ORole
```
PROPERTIES
-------------------------------+-------------+-------------------------------+-----------+----------+----------+-----------+-----------+----------+
 NAME                          | TYPE        | LINKED TYPE/CLASS             | MANDATORY | READONLY | NOT NULL |    MIN    |    MAX    | COLLATE  |
-------------------------------+-------------+-------------------------------+-----------+----------+----------+-----------+-----------+----------+
roles                         | LINKSET     | ORole                         | false     | false    | false    |           |           | default  |
...
```
But when i try to create index under distributed mode, i got following exception 

orientdb {db=pharos}> create index user.roles NOTUNIQUE_HASH_INDEX

Creating index...
Error: com.orientechnologies.orient.server.distributed.ODistributedException: Error on execution distributed COMMAND

Error: com.orientechnologies.orient.core.index.OIndexException: Cannot create the index 'user.roles'
```
Error: java.lang.ClassCastException: com.orientechnologies.orient.server.distributed.ODistributedStorage cannot be cast to com.orientechnologies.orient.core.storage.impl.local.OAbstractPaginatedStorage
```
Tried on single mode, it works fine.

henryzhao81:
(OAbstractPaginatedStorage) db.getStorage()
in distributed mode, in OStreamSerializerSBTreeIndexRIDContainer db.getStorage() will return ODistributedStorgae
public class ODistributedStorage implements OStorage, OFreezableStorage, OAutoshardedStorage
ODistributedStorage not extends from OAbstractStorage, then throw class cast exception

lvca:
Fixed.

lvca:
(test cases added only to 2.1)

PR #4368, opened 2015-06-15
Issue #4285 propagate the fix to 1.7.x

henryzhao81:
for issue https://github.com/orientechnologies/orientdb/issues/4285 propagate the fix to 1.7.x


lvca:
Thanks!

PR #4400, opened 2015-06-19
Auto-create databases for distributed-mode OrientDB

thanethomson:
It seems as though, due to this bit of code in the OServer class, distributed-mode OrientDB doesn't automatically create databases defined in your `orientdb-server-config.xml` file. I've reordered the code so that storages are created prior to firing up plugins (e.g. Hazelcast).

With my particular application, I most certainly can't afford to create databases programmatically or via the console. Without this fix, distributed OrientDB crashes on startup and never gets to the point where it should create locally specified storages.

I've personally built this against the `develop` branch and while building `v2.0.10`, and all tests are successful.


lvca:
Thanks vey much for such contribution, merging it now.

lvca:
Back ported also to 2.0.11 just before releasing it. Thanks again.

PR #4408, opened 2015-06-20
Change to ensure list order on save

aheintz:
Currently, when adding an object to a list using add(int, Object), the item is, in the database and on retrieval, added to the end of the list. This change fixes this problem. Unit test provided.


tglman:
well made pull request, tested, merging

PR #4446, opened 2015-06-23
Provide Hazelcast with orient-distributed bundle classloader.

ylopotun:
Provide Hazelcast with orient-distributed bundle classloader, which is aware of ODocument class.


lvca:
Thanks!

PR #4493, opened 2015-06-30
Allow signal handler to be uninstalled separately from the shutdown hook

mcculls:
... as you might want to replace the shutdown hook, but keep the signal handling

(this is an improvement to https://github.com/orientechnologies/orientdb/pull/4364)


mcculls:
For example, the server currently removes the core shutdown hook as it wants to use its own:

https://github.com/orientechnologies/orientdb/blob/master/server/src/main/java/com/orientechnologies/orient/server/OServerShutdownHook.java#L29

but ideally this should keep the signal handlers installed, since it doesn't install its own handlers.

lvca:
Thanks.

PR #4502, opened 2015-07-02
Unit test that demonstrates the need of reload.

ylopotun:
Unit test that demonstrates the need of reload when getting vertex edges.

### DIFF
--- a/distributed/src/test/java/com/orientechnologies/orient/server/distributed/asynch/TestGetVertexEdgesNoReload.java
+++ b/distributed/src/test/java/com/orientechnologies/orient/server/distributed/asynch/TestGetVertexEdgesNoReload.java
@@ -1,0 +1,94 @@
+package com.orientechnologies.orient.server.distributed.asynch;
+
+import java.util.Iterator;
+
+import com.tinkerpop.blueprints.Direction;
+import com.tinkerpop.blueprints.Edge;
+import com.tinkerpop.blueprints.impls.orient.OrientBaseGraph;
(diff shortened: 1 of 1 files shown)


lvca:
Hi guys,
My question is: what's the reason you don't manage concurrent updates in the suggested way (optimistic approach = retry)? Take a look at:

http://orientdb.com/docs/last/Graph-Database-Tinkerpop.html#transactions

for (int retry = 0; retry < maxRetries; ++retry) {
    try {
        // LOOKUP FOR THE INVOICE VERTEX
        Vertex invoice = graph.getVertices("invoiceId", 2323);
        // CREATE A NEW ITEM
        Vertex invoiceItem = graph.addVertex("class:InvoiceItem");
        invoiceItem.field("price", 1000);
        // ADD IT TO THE INVOICE
        invoice.addEdge(invoiceItem);
        graph.commit();

        // OK, EXIT FROM RETRY LOOP
        break;
    } catch( ONeedRetryException e ) {
        // SOMEONE HAVE UPDATE THE INVOICE VERTEX AT THE SAME TIME, RETRY IT
    }
}
So in case of concurrent modification, the OConcurrentModificationException extends the ONeedRetryException, so it's caught, and the thread retry to update it. This technique (CAS) is very common on concurrent systems. Take a look at: http://stackoverflow.com/questions/2664172/java-concurrency-cas-vs-locking. Also Java internally uses CAS (http://java.dzone.com/articles/how-cas-compare-and-swap-java).

Issue #4535, opened 2015-07-09
Default value and creation of new document


PhantomYdn:
Guys,

From common logic perspective `new ODocument(classA)` should already have all fields with default values prepopulated, but OrientDB requires storing of ODocument for population of default values. And more over: this done in strange for this taks method ODocument.validateField().

I suggest:
1) Refactor ODocument.validateField() (might be in parallel with #3429)
2) Allow to prepopulate ODocument by default values on creation, but not onSave

So, what do you think?

luigidellaquila:
IMHO pre-populating ODocument with default values is a good solution. I just have to check how it will affect backward compatibility.

lvca:
:+1

PhantomYdn:
Guys, please consider PR. All tests are OK.

lvca:
The call to the populateDefaultValues() should be done also on setClass() and setClassname() right?

PhantomYdn:
My idea was that once document created we can do with the object what ever we want, but fields values remains the same. Other wise there are lots of additional questions like: what to do with the a prepopulated field if super class was changed? Clean it or keep it? What to do if new class has the same property by with some default value? And so on.

lvca:
The problem is that new ODocument(<class>) is semantically equivalent to: new ODocument(<class>)

PhantomYdn:
My idea was that in case 

```java
new ODocument().setClassname(<class>);
```

document will not have default values, because class was defined after creation of a document, so population of default values should be explicitly invoked.

In case 

```java
new ODocument("ClassA").setClassname("ClassB");
```
document will have default values from ClassA and changing to ClassB will preserve already set values.

If you disagree: just let me know proper way to resolve that.

PhantomYdn:
Seems to be fixed in #4552. Closing?

PR #4552, opened 2015-07-11
Default values prepopulation. For issue #4535

PhantomYdn:


lvca:
WDYT about pre-populate them on setClass() and setClassname() ?

PhantomYdn:
In this case I personally don't like a side effect: if class was set twice - class will have default values also from first class populated...

lvca:
You're right, but semantically these 2 calls should produce the same effect:
```java
new ODocument("Person");
new ODocument().setClassname("Person");
```

PhantomYdn:
@lvca agreed! PR has been updated with recent code. Please take a look into testcase for all supported cases.

PhantomYdn:
Guys? Build is failing due to timeout. Local test is fine.

lvca:
Perfect, merging it now. Thanks.

Issue #3427, opened 2015-01-20
Vertex.getProperty on an embedded list of documents doesn't work right


countfloyd:
I've created a vertex with an embedded list of documents like so:

List<ODocument> l = ....
v.setProperty("rules", l);

Trying to read them back is a problem. If I try:

v.getProperty("rules")

I get an OrientElementIterable containing one ORecordLazyList which contains ODocument's. I haven't been successful converting this to a List.

If I try:

((OrientVertex)v).getRecord().field("rules")

I get an ORecordLazyList of ODocuments which I can then treat like a list and everything is fine. The Vertex.getProperty seems to insert an extra iterable at the top.

countfloyd:
The bigger problem is if I try to iterate on the OrientElementIterable returned from Vertex.getProperty, I get an exception:

```java
[java.lang.IllegalArgumentException: Cannot determine the graph element type because the document class is null. Probably this is a projection, use the EXPAND() function] with root cause
java.lang.IllegalArgumentException: Cannot determine the graph element type because the document class is null. Probably this is a projection, use the EXPAND() function
	at com.tinkerpop.blueprints.impls.orient.OrientElementIterator.next(OrientElementIterator.java:70)
	at com.tinkerpop.blueprints.impls.orient.OrientElementIterator.next(OrientElementIterator.java:34)
	at com.guildone.synergy.server.rest.elements.FiltersResource.getFilterFromVertex(FiltersResource.java:201)
	at com.guildone.synergy.server.rest.elements.FiltersResource.lambda$getFiltersForUser$8(FiltersResource.java:185)
	at com.guildone.synergy.server.rest.elements.FiltersResource$$Lambda$11/732670953.apply(Unknown Source)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:512)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:502)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #4556, opened 2015-07-12
fix for #3427. Added test case.

fabriziofortino:


fabriziofortino:
@lvca can this be merged?

lvca:
It's impossible to see what you changed in ORecordLazyList.java. Is it a formatting problem?

fabriziofortino:
I think so. I have just removed an unused import.

lvca:
Ok, merging anche checking it right after.

lvca:
Thanks, as always ;-)

nagarajasr:
@lvca my 2c, in case you are using the github web interface to see the diff, add ?w=1 at the end of the URL to ignore whitespace in the comparison, using that, i can clearly see that there was only 1 line change to remove the unused import.

lvca:
`?w=1` works, thanks!

PR #4563, opened 2015-07-13
Mark junit and orientdb-graph as scope=test in orientdb-distributed module

jdillon:
junit and orientdb-graph are only used by src/test/ bits, no reason to expose these as scope=runtime, but instead should be scope=test.


lvca:
It's all ok, thanks.

Issue #4558, opened 2015-07-12
Expose configuration for plugins directory


jdillon:
ATM the `plugins` directory is hard coded to:

```
private void updatePlugins() {
    final File pluginsDirectory = new File(OSystemVariableResolver.resolveSystemVariables("${ORIENTDB_HOME}", ".") + "/plugins/");
```

I'd like to be able to define a property in the server configuration to specify the location of this instead.

Should help with embedding where the runtime environment is different from the shrink-wrap distribution.

jdillon:
It would also be kinda nice if you could install a plugin from a URL, so we can embed a plugin into an OSGI bundle and then install it based on the resource url?  And avoid the entire plugins/ directory completely?

lvca:
- Exposed Plugin Manager instance in OServer class
- new "plugin.loadAtStartup" to decide if load plugins at startup
- new "plugin.directory" settings to assign a different directory

PR #4565, opened 2015-07-13
Expose the location of the plugins directory in server config

jdillon:
Maybe a solution to https://github.com/orientechnologies/orientdb/issues/4558


jdillon:
Looks like the underlying code actually makes URL to function so exposing the URL-based configuration to install/update plugins may not be too much of a stretch:

```
  private void installDynamicPlugin(final File pluginFile) {
    String pluginName = pluginFile.getName();

    final OServerPluginInfo currentPluginData;
    OLogManager.instance().info(this, "Installing dynamic plugin '%s'...", pluginName);

    URLClassLoader pluginClassLoader = null;
    try {
      final URL url = pluginFile.toURI().toURL();
```

lvca:
We could expose the Plugin manager outside.

PR #4576, opened 2015-07-15
Fix for NPE in  ODocument.undo(field)

PhantomYdn:
First commit is the main. Second one is just to optimize other undo


PhantomYdn:
Guys?
Imho, it's quite safe PR

laa:
Thank you !

PR #4580, opened 2015-07-15
Avoid direct dependency on com.sun.management classes

jdillon:
This prevents problems with OSGI environments not having wiring information for `com.sun.management`:

```
Unable to resolve 154.0: missing requirement [154.0] osgi.wiring.package; (&(osgi.wiring.package=com.orientechnologies.common.console)(version>=2.1.0)) [caused by: Unable to resolve 164.0: missing requirement [164.0] osgi.wiring.package; (osgi.wiring.package=com.sun.management)]
```

... and should still allow HeapDumper.dumpHeap(String,boolean) to function.

### DIFF
--- a/core/src/main/java/com/orientechnologies/common/util/HeapDumper.java
+++ b/core/src/main/java/com/orientechnologies/common/util/HeapDumper.java
@@ -1,48 +1,25 @@
 package com.orientechnologies.common.util;
 
-import com.sun.management.HotSpotDiagnosticMXBean;
-
 import javax.management.MBeanServer;
+import javax.management.ObjectName;
+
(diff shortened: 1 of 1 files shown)


jdillon:
Also this class is not used directly, only referenced by `com.orientechnologies.orient.core.Orient` by import  but not actually used.

The static helper may still be useful to invoke via scripting, but you might want to remove the import, and defs want to apply this to avoid leaking in the `com.sun.management` bits, or if this was only for testing remove the HeapDumper class.  Seems kinda useful so I'd suggest if you need this in the future you apply this PR to avoid direct deps on the classes, while still exposing the functionality you want.

laa:
Thank you !

PR #4594, opened 2015-07-16
Expose extension class-loader to be used when loading classes from config

jdillon:
This should allow embedded use-case to configure the appropriate class-loader to resolve handler, protocol and socket extension impls.

I considered making a helper for this as there are current **~41** uses of `Class.forName` in the code base, and you folks may want to consider something like that for now however I opted for a localized change to allow OServer to load classes in OSGI/embedded environments better.

`Class.forName` really should be avoided generally, but fully fixing that is a much larger change and needs some insight from you guys about intended use of dynamic class loading to properly craft a solution.

Note I marked some things with `/*@Nullable*/` as the findbugs annotations where not on class-path, I may followup to fix that.

Some refs for reading about why Class.forName is bad and how to do the same thing in safer/portable ways:

* http://blog.bjhargrave.com/2007/07/why-do-classforname-and.html
* http://blog.bjhargrave.com/2007/09/classforname-caches-defined-class-in.html
* http://njbartlett.name/2010/08/30/osgi-readiness-loading-classes.html

### DIFF
--- a/server/src/main/java/com/orientechnologies/orient/server/OServer.java
+++ b/server/src/main/java/com/orientechnologies/orient/server/OServer.java
@@ -107,6 +107,8 @@ public class OServer {
   private String                                           databaseDirectory;
   private final boolean                                    shutdownEngineOnExit;
 
+  private ClassLoader extensionClassLoader;
+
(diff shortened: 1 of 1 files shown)


lvca:
Merging this PR, and created issue #4597. Thanks.

PR #4601, opened 2015-07-17
Fir for setting the same class name

PhantomYdn:
I understand, that setting the same name to a OClass is pretty strange, but system throwing exception in this case.

Btw, OClassImpl allow to have several OClasses with the same shortname. Is this a feature or bug?


lvca:
Setting classes with the same shortname is a bug. Did you already try it?

PhantomYdn:
Yes. And that works - so but is there...

PR #4604, opened 2015-07-18
Use scope=test for dependencies used for testing

jdillon:
Trivial pom change which marks newly added fluent-hc dependency ( from https://github.com/orientechnologies/orientdb/commit/923183a18bf26bd9cae5ed0e70d446c934fda1b7 ) as scope=test.


lvca:
Makes sense. Thanks.

PR #4609, opened 2015-07-20
Widen InterruptedException cause check to cover other exceptions.

mcculls:
For example when running distributed OrientDB under OSGi the worker might throw a runtime exception whose cause is InterruptedException if it happens to be interrupted when loading classes/resources.

Also include a safety-check during cleanup due to https://github.com/hazelcast/hazelcast/issues/5526


lvca:
Thanks!

Issue #4611, opened 2015-07-21
Property default values don't seem to survive server restart


caturner81:
So, I'm pretty new to OrientDB, so I could be missing something.  I like the feature set and performance, but thus far the documentation is very unclear with many things being outdated, dead links, or suggesting things that turn out to be deprecated versions ago.

All that said, I'm trying to use the new default value functionality and no matter how I set a default value (SQL through the studio, programmatically via the Graph API, or programatically via TinkerPop), the default value setting does not survive restarting the server.  I think I'm doing things correctly given that setting other property values like read only or mandatory work as expected and persist fine.  Looking at your tests it looks like you just have an in memory test where you create a class with a couple properties with defaults and then drop it, but never test actually persisting the schema change.

nagarajasr:
which version of orientdb did you try?

caturner81:
Oh blast it, sorry about that, 2.1 RC5.

nagarajasr:
I am able to reproduce the issue on 2.1-RC5, using a database with plocal storage.
the schema modification commands were run from console.
looks like a bug to me.

in addition, the default value attribute is not displayed when i query the metadata:schema
in the below example schema i had defined a default value 123 for the 'maxid' property.

```sql
select expand(properties) from (select expand(classes) from metadata:schema) where name='testclass' limit -1

----+------+----------+----+--------+---------+--------+-------+----+----+------+--------------+-------
#   |@CLASS|name      |type|globalId|mandatory|readonly|notNull|min |max |regexp|customFields  |collate
----+------+----------+----+--------+---------+--------+-------+----+----+------+--------------+-------
0   |null  |out_photos|23  |77      |false    |false   |false  |null|null|null  |{ordered=true}|default
1   |null  |maxid     |1   |80      |false    |false   |false  |null|null|null  |null          |default
----+------+----------+----+--------+---------+--------+-------+----+----+------+--------------+-------
```

nagarajasr:
i think i've found a fix for this issue. will submit a PR in a while.

nagarajasr:
submitted PR #4621 against develop branch

lvca:
Merged 2 months ago, closing it. Thanks @nagarajasr for the PR.

PR #4621, opened 2015-07-21
Attribute 'Default Value' should be persisted along with other attributes. Fixes issue #4611

nagarajasr:


nagarajasr:
@lvca any chance this can be merged before the 2.1 GA? Considering that 'default values' is one of the new features in 2.1 release, the issue #4611 must be resolved before GA i think.

lvca:
You're absolutely right. Merged. Thanks.

PR #4661, opened 2015-07-27
Add configuration/profile for invocation of the animalsniffer

jdillon:
This will help you ensure that your codebase does not refer to newer JDK API.

You may need to add some annotations to mark things are not really problems, but this should help you ensure you don't accidentally leak in incompatible JDK references in the future.

ATM this will cause your build to fail if you enable the profile (but that is its job to prevent you from leaking in incompatible api), as there are several places where you are using Java7 api, and also a handful of places that use Unsafe.  Probably needs someone with understanding of why/how these are used to apply the IgnoreJRERequirement annotation as appropriate.

https://github.com/mojohaus/animal-sniffer/blob/master/animal-sniffer-annotations/src/main/java/org/codehaus/mojo/animal_sniffer/IgnoreJRERequirement.java

Profile is disabled by default, recommend you update your CI automation to add `-Panimalsniffer` to enable the check.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -241,6 +241,23 @@
                         <arguments>-Prelease</arguments>
                     </configuration>
                 </plugin>
+
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>animal-sniffer-maven-plugin</artifactId>
+                    <version>1.13</version>
+                    <configuration>
+                        <!--
+                        Keep aligned with the maven-compiler-plugins source/target, ATM this is 1.6,
+                        so we use the java16 signature.
+                        -->
+                        <signature>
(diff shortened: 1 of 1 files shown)


lvca:
I didn't know it, merged. Thanks. @tglman Could you please add the -Panimalsniffer to Jenkins?

tglman:
before enable it on jenkins we have to configure it/fix the code, there are a few problems like unsafe.

lvca:
Does animalsniffer fights with Unsafe?

jdillon:
@lvca doesn't fight with it, but complains about its improper usage (which pretty much every use is improper according to the original authors, but that is another story).

You can/will need to add a few @IgnoreJRERequirement annotations (best with some inline comments why) to mark things that you are aware of being non-standard/problematic.

PR #4662, opened 2015-07-27
Use new InternalError().initCause(e) for < java8 compat

jdillon:
This should restore buildability with java7.

Highly recommend you look this related PR to help avoid brining in incompatible JDK usage in the future:

https://github.com/orientechnologies/orientdb/pull/4661


lvca:
@jdillon Thanks again.

PR #4679, opened 2015-07-29
Added iDatabaseName to dropDatabase and existsDatabase

alonsod86:
Added the possibility to drop or check for existence of a database specifying the database name. This functionality existed in createDatabase before, but for some reason it was not included in drop and exists.


lvca:
The PR seems ok, but you did it against "master" branch. We use "master" only for releasing. Please, could you do the same but to "develop" branch?

alonsod86:
For sure! My bad..

Issue #4497, opened 2015-07-01
dijkstra function ignores <weightEdgeFieldName>


mbhulin:
Consider this simple graph: 
V has nodes #20:0, #20:1 and #20:2
E has these connections
from #20:0 to #20:2 with w1 = 3 and w2 = 3
from #20:0 to #20:1 with w1 = 1 and w2 = 1
from #20:1 to #20:2 with w1 = 1, no w2 field
The result of SELECT dijkstra(#20:0, #20:2, 'w1') is [#20:0, #20:1, #20:2] which is the expected result because the total weight 1 + 1 = 2 is less than the direct connection from #20:0 to #20:2 with weight 3.

The result of SELECT dijkstra(#20:0, #20:2, 'w2') is [#20:0, #20:1, #20:2] which is wrong because the only connection from #20:0 to #20:2 following edges with a w2-field is the direct connection. Hence the result should be [#20:0, #20:2]

luigidellaquila:
Hi Martin,

in current implementation, if an edge does not have a valid value on the weight property, then 0 is assigned by default.
I think it could be a good idea adding another parameter to the function for one of the following:
* default weight, to be assigned when property does not exist or
* boolean flag to ignore edges where the property does not exist

WDYT

Luigi

mbhulin:
Hi Luigi,
both suggested solutions can solve the problem. Personally I prefer the second one but of course you can get the same result with the first solution by setting the default value to Float.MAX_VALUE.

However I'm still not completely satisfied. An application might use different subclasses of E as edges with the same weight field. I suggest another parameter with the edge type. If this parameter is set only edges of this class are considered. Of course you could avoid this problem by different names for the edge classes e. g. distance1 and distance2 instead of distance. However such a work around might decrease readability.

I have reimplemented the dijkstra algorithm of OrientDB with few changes in my client program:
- additional parameter edgeType: only edges of this subtype of E are considered
- Only edges with the weight field are considered for the best path
- The output of my dijkstra function is not only the list of vertices but also the weight of the path and the list of edges. This is relevant if you may have more than one edge between two verteces e. g. two different streets between two cross points. Then you want to know which of the streets is on the optimal path.

Are you interested in my version? However the last change is not suitable for the dijkstra function in SQL because in SQL you only want to get one result (i. e. the list of vertices) not a structured result with a list of vertices, a list of edges and a float value. Perhaps OrientDB could offer a method
[...]

PR #4694, opened 2015-07-31
Another version of the dijkstra function: dijkstra2

mbhulin:
As discussed in issue #4497 dijkstra2 is a copy of the original dijkstra function with the following changes:
* dijkstra2 has 4 or 5 parameters, dijkstra has 3 or 4 parameters 
* The third parameter is the edge type i. e. the subclass of E considered for routing. Other edges are neglected in dijkstra2.
* The fourth parameter is the name of the property that represents the 'weight' or 'cost'. In the original dijkstra function this is the third parameter.
* The optional direction parameter is the fifth parameter in dijkstra2 but the fourth parameter in dijkstra.
* If an edge has no weight field or the weight field is null this edge is not considered for routing, becouse the weight is set to Float.MAX_VALUE. In contrast the original dijkstra function sets the weight to zero in this case.
* If source and destination coincide return a path with only one vertex regardless whether a self-loop exists. This feature of dijkstra2 can be discussed: On one hand in a routing task if source and destination are the same you can reach the destination without effort. On the other hand there is no path with edges from source to destination.

There is also a new JUnit test OSQLFunctionDijkstraTest2 for the dijkstra2 function copied from OSQLFunctionDijkstraTest with additional test cases. All tests succeed.


lvca:
Thanks @mbhulin, @luigidellaquila will review it. I'd prefer to change the original dijkstra() function adding custom parameters in form o JSON. Example:

```js
dijkstra({sourceVertex:#2323:32, destinationVertex:#232:23, edgeType:"Knows", weightEdgeFieldName:"since", direction:"OUT"})
```

In this way we can add feature to functions without break the signature. WDYT?

mbhulin:
Hi Luca, 
I considered to have only one dijkstra funktion, the original one with 3 or 4 parameters and overload it with the new one with 5 parameters. The 5th argument then would be the edgeType. However a user might be confused by the different handling of edges which have no weight-field: In the original version an edge without the weight field is the ideal connection with zero cost, in the overloaded version an edge without the weight field is the worst connection with Float.MAX_VALUE as cost.

In order not to hide this difference I suggested two versions. Of course your suggested solution is possible, too. It has the advantage that you don't have a lot of similar functions dijkstra, dijkstra2, dijkstra3, ... I think it's your decision.

lvca:
@mbhulin I really don't like the idea of having different dijkstra functions with a different number to differentiate them. If the original dijkstra() considered no weight as 0 (the best), it's a bug IMHO. So I'd prefer to have one-more-powerful dijkstra() function ;-) The problem is managing optional params in better way. We could pass a JSON, but I'm not sure the SQL parser would eat it correctly.

w3aponx:
Other variants I could think of would be to have a list of allowed edge
types (instead of a single one), possibly with different cost depending on
edge class, and with the ability to allow sub-classes of the specified edge
classes or not.

On Mon, Aug 24, 2015 at 12:27 PM, Luca Garulli <notifications@github.com>
wrote:

> @mbhulin <https://github.com/mbhulin> I really don't like the idea of
> having different dijkstra functions with a different number to
> differentiate them. If the original dijkstra() considered no weight as 0
> (the best), it's a bug IMHO. So I'd prefer to have one-more-powerful
> dijkstra() function ;-) The problem is managing optional params in better
> way. We could pass a JSON, but I'm not sure the SQL parser would eat it
> correctly.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/orientechnologies/orientdb/pull/4694#issuecomment-134289625>
> .
>



-- 
_____________
Mario Cormier

gg4u:
Hi, as a user, I would also be interested in using djikstra with reversed weight, that is, instead of cheapest, most expensive weight.

Is the new function already based in master? Is documentation updated?
I'd like to try it.
Also, customed parameter 'weight' is a good thought.

mbhulin:
@lvca If you think that the behaviour of the original dijkstra function to set "no weight" = "weight 0" is a bug the solution is easy. I also regarded this behaviour as bug but in the description it is explicitly specified: "* If property is not defined in edge or is null, distance between vertexes are 0." The question is whether there are already users who use the dijkstra funktion and rely on the specified behaviour.
Solution, if original dijkstra function may be changed:
* dijkstra(sourceVertex, destinationVertex, weightEdgeFieldName)
All edges with "weightEdgeFieldName" are considered with default direction OUT. If this field is missing or NULL the weight is set to Float.MAX_VALUE (worst possible weight).
* dijkstra(sourceVertex, destinationVertex, weightEdgeFieldName, direction)
All edges with "weightEdgeFieldName" are considered in the specified "direction". If "weightEdgeFieldName" is missing or NULL the weight is set to Float.MAX_VALUE (worst possible weight).
* dijkstra(sourceVertex, destinationVertex, weightEdgeFieldName, direction, edgeType)
New version; only edges of the specified "edgeType" in the specified "direction" are considered.

Shall I change the original dijkstra function in my branch and update this pull request?

@w3aponx There could be more than one edgeType in the new version:
* dijkstra(sourceVertex, destinationVertex, weightEdgeFieldName, direction, edgeType [, edgeType-n]*)

mbhulin:
@luigidellaquila In issue #4946 Luca suggested to reimplement the dijkstra function to increase performance. If you do that, you could consider the discussion above. Or shall I try to implement a new version of the dijkstra function based on the actual implementation of shortest path?

mbhulin:
**dijkstra2 function** as extra function additionally to the original dijkstra function **deleted**.
**dijkstra3 function** added instead which could **replace** the original **dijkstra function**.
Differences of dijkstra3 to dijkstra:
* If source and destination coincide return the trivial path with only one vertex instead of an empty path
* If an edge has no 'weight' field or the 'weight' field is null this edge is not considered for routing (gets the maximal possible weight)
* The result is List<ORID> instead of LinkedList<OrientVertex>
* If no path exists the result is an empty list instead of null
* Additional optional fifth parameter:  the edge type i. e. the subclass of E considered for routing; default is E (all edges)
* Implementation uses a priority queue

lvca:
Hi @mbhulin thanks for the contribution. I think dijkstra3 function could replace current dijkstra function. @luigidellaquila WDYT?

saeedtabrizi:
Hi , @mbhulin, @lvca , @luigidellaquila 
Based On http://theory.stanford.edu/~amitp/GameProgramming/Heuristics.html if we return a fixed number in heuristic part of A Star function our result must be equal to Dijkstra’s function . 
I suggest use the a star implementation with zero heuristic value to achieve more functionality in Dijkstra function . 

Thanks 
Saeed Tabrizi

luigidellaquila:
@saeedtabrizi +1, I'm rewriting current Dijkstra implementation to delegate to A* algorithm

Issue #4706, opened 2015-08-02
Disable 'root' user 


matanshukry:
As a security measure, I would like to disable the 'root' user completely.
Before starting up orient service for the first time, I set only one user, not root, in the config file.
However, orient creates a root user anyway, with a random password.

I noticed the use of root password is used as constant SRV_ROOT_ADMIN in OServerConfiguration.java, which is only used in OServerShutdownMain.java and OServer.java.

in OServer.java, it is only used to check if it's the first time the service is run. We can simply choose a place and use a boolean config, something like 'isFirstTime'. Seems to clarify the behavior too, in my opinion anyway.

in OServrShutdownMain.java , it is used to connect to a server. We can simply get a user argument from command line, much like the other arguments (and use 'root' as default user).

WDYT?

nagarajasr:
:+1:

lvca:
:+1

lvca:
Merged, thanks @matanshukry

tglman:
reopening due to revert for failing tests.

tglman:
fixed now closing

matanshukry:
Hi, this is a bit confusing;
as far as I can tell, you create PR to revert this issue, merged it, and then reverted the issue that reverted the issue. I'm not sure about it though.

Anyway, what is the problem? if you fixed it, what was it?

tglman:
Hi @matanshukry,

Now is merged, the problem was that in the test suite the are many test that rely on the root password written in the xml, i fixed them adding in the configuration the flag for not ask the password.

PR #4715, opened 2015-08-03
Disable 'root' user

matanshukry:
PR for #4706

Note:
As far as I can tell, the line endings are CRLF, yet some of the files didn't seem consistent with that, and IntelliJ automatically changed it. Same for line spacing.

I hope that's ok.


lvca:
Very cool, thanks!

tglman:
hi, this pull request actually make fail the server block on start when the root password is configured in the orientdb-server-config.xml, in our test suite we have tests that block due to this patch see the end of this: http://helios.orientdb.com/job/orient-maven-csv-serialization/566/console, i've to revert it.

Issue #4725, opened 2015-08-04
Query functions over collections [+using field aliases]


matanshukry:
When running the following query:

`SELECT id, name, In(Friend).name as friend_name, In(Friend).name.append(" Test
") as friend_name_test FROM User where id=1`

The result is:

```
[{"@type":"d","@rid":"#-2:0","@version":0,"id":1,"name":"Don Kingsly","friend_name":["","Samantha Bensly"],"friend_name_test":"[, Samantha Bensly] Test","@fieldTypes":"id=l"}]
```

As you can see, the field 'friend_name_test' is a single value string, which is probably the collection .toString(), and concat "test". The intention, although, was to use .append() on every item in the collection.

Also, since on the subject, would be nice if I could refer to aliases fields, so that the query above would become:
`SELECT id, name, In(Friend).name as friend_name, friend_name.append(" Test") a
s friend_name_test FROM User where id=1`
(The result of this doesn't include any friend_name_test fields, I'm guessing since it doesn't recognize the friend_name).

In my opinion, any function over a collection should apply to each item in the collection.
If one wants to use the entire collection, he may use .toString() or .size(), and then what he wants.

Your opinions?

matanshukry:
Took the initiative and started implementing it (hopefully I didn't do it for nothing, and it's not implemented yet and I just missed it).

When doing so, I found it was problematic, since you couldn't choose between applying the function to each item in a collection, or to the collection as a whole.

As a result, I added an .each() method. That way, the following query should work:
```SELECT id, name, In(Friend).name as friend_name, In(Friend).name.each().append(" Test ") as friend_name_test FROM user where id=1```

So, what do you guys think?

lvca:
Not relevant anymore. Closing it.

PR #4728, opened 2015-08-04
Functions over collections

matanshukry:
PR for #4725


nagarajasr:
i think context variables are passed on to sub-queries so if each() is used once, the context variable is set for the entire query and sub-queries, so it will not be possible for sub-queries to apply the method to whole collection. however i have not tested to confirm if this is indeed the case.

matanshukry:
@nagarajasr  I added a paremeter (true or false), so one can disable it.

The following now works:
```
SELECT id, name, friend_name_test.each(false).append("Parent") FROM ( SELECT id, name, In(Friend).name as friend_name, In(Friend).name.each().append(" Test ") as friend_name_test FROM user where id=1)
```

The result:
``
"id":1,
"name":"Joe",
"friend_name_test":"[Mat Test , Sam Test ]Parent"}]`
```
I know it looks weird, but it's correct (appended " test " to each, then "parent" to both).
This specific case won't make sense, but you can combine .each() as you like

p.s.
The true/false parameter also works for multiple fields inside the same query level.

matanshukry:
Added a fix, all tests are passing now.

PR #4732, opened 2015-08-05
Better exception when you try to update a deleted distributed record

mcculls:
Use a more accurate exception when a distributed record couldn't be updated because it has been deleted, so clients can react accordingly. Otherwise it's hard to discern this from other instances of `OTransactionException` which typically indicate a problem with the internal transaction state.

This change also makes it more consistent with the exception used in `OUpdateRecordTask`.


lvca:
:+1:

PR #4736, opened 2015-08-05
Fix node exclusion in distributed commands

mcculls:
A small number of commands were incorrectly re-creating command instances instead of using the variable which already had the local node excluded. This caused the stack to overflow when these commands were issued on distributed storage.


lvca:
Perfect, thanks

PR #4738, opened 2015-08-05
Unit test for issue #4913

ylopotun:
Added unit test that demonstrates issue:
http://www.prjhub.com/#/issues/4913


lvca:
Merged manually against 2.1.x and majors, thanks.

PR #4701, opened 2015-08-01
added initial .travis.yml

krichter722:

PR #4746, opened 2015-08-06
added initial .travis.yml

krichter722:
see #4701

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+language: java


lvca:
Thanks.

krichter722:
If you register orientechnologies at travis-ci.org, you can activate checks for every PR and all members.

lvca:
We didn't use Travis anymore because we have Jenkins at helios.orientdb.com. However, I've re-activated Travis support because the PR ;-) Thanks.

krichter722:
In case it ever falls behind the requirements of `orientdb` build and test process or support is re-dropped for another reason, the content of `.travis.yml` can be changed to `exit 1` and a comment could be added in order to indicate that the service has been taken into consideration.

Issue #4718, opened 2015-08-03
Console return collections as array length rather than content


matanshukry:
Console return array length rather than array content.

Is there any reason it would be the intended response, or is it simply a bug?

p.s.
If anyone managed to see the old comment & post, I deleted it because it's not relevant anymore.

matanshukry:
A bit of follow up (hence the extra comment rather than edit),
It seems when I'm querying the data using the HTTP API I am receiving it as expected (array with results).

The result shown as an array length is only displayed while in the console.
I did not go over the code, but I believe it is due to console side code.

Hence I am not closing this issue, since it needs to be fixed

PR #4758, opened 2015-08-08
Console return collections as array length rather than content [PR]

matanshukry:
PR for #4718


nagarajasr:
IMO it would be prudent to put a limit on the number of values that should be printed. e.g., if the number of values exceeds some pre-defined limit, say 20, then either fall back to the earlier behaviour or print only the first 20 values with an ellipsis indicating more values are present.

lvca:
@matanshukry Thanks again. @nagarajasr I like you idea, implementing it after merge

lvca:
Done in develop.

PR #4761, opened 2015-08-09
[HTTP API] Exception when creating property with linkset

matanshukry:
Running the following
```
curl -XPOST http://localhost:2481/property/test/User/places1/LINK/place -u root:root
```

Will result in:
```
{
  "errors": [{
      "code": 500,
      "reason": 500,
      "content": "java.lang.IllegalArgumentException: No enum constant com.orientechnologies.orient.core.metadata.schema.OType.place"
      }
  ]
}
```

this PR fixes it.

Notes:
1. The problem was there was a call to OType.valueOf(), which threw the above exception. The intended behavior, at least as can be seen by the code, was to match either an OType or a custom class. Hence it is now silently ignored by surrounded in a try catch.
2. I rearranged the calls a bit, so that if OType.valueOf() succeed, there will be no lookup for a class with the same name. One less call!


lvca:
Thanks.

Issue #4727, opened 2015-08-04
Can't Set Array Value using SQL


mmacfadden:
Attempting to set the value of an array via SQL does not work.  We are currently using the Document API from Java.  

Given the following document: 
```javascript
{
    "someArray": [
        10,
        20,
        30
    ]
}
```
The following query returns successful, but does nothing

```sql
UPDATE model SET someArray[1] = 25
```

The expected result is:

```javascript
{
    someArray: [
        10,
        25,
        30
    ]
}
```

The current work around requires loading the entire array and treading it as a OTrackedList which can then be manipulated and then set back as an entirely new value for "someValue".  This is not practical for large arrays.  The above example was greatly simplified.  The actual use case involves larger arrays that have large objects as their values.

Also, if two users were editing the data such that each was attempting to edit a different value in the array, the approach of getting the whole array, modifying it, and setting it back would create an artificial conflict where none really occurs.  

This issue was first discussed in the group here:

https://groups.google.com/forum/#!topic/orient-database/YCSyS7sQKTo

mmacfadden:
To add more information, we came across another use case that makes this feature more important.  If one has nested arrays, then the amount of data that needs to be loaded is even more, and the efficientcy drops rapidly with the current functionality.

```javascript
{
    "array": [
        [
           10,
           20,
           30
        ],
        [
           "A",
           "B",
           "C"
        ]
    ]
}
```

What we would like to do is to update the 20 to be a 25.  If this was just a JSON node I would do the following:

```javascript
doc[0][1] = 25
```

Our current workaround for the original use case above was something to the effect of:

```Java
String queryString = "SELECT array FROM model";
OSQLSynchQuery<ODocument> query = new OSQLSynchQuery<ODocument>(queryString);
List<ODocument> result = db.command(query).execute(params);

OTrackedList<Object> array = result.get(0).field(FIELD_DATA);
array.set(1, 25);

String updateString = "UPDATE model SET array = :value ";
OCommandSQL updateCommand = new OCommandSQL(updateString);
params.put("value", array);
db.command(updateCommand).execute(params);    
```

This worked.  However, this work around DOES NOT work with nested arrays.  The path to the nested array is "array[0]".  Therefore the code would wind up looking like this:

```Java
String queryString = "SELECT array[0] FROM model";
OSQLSynchQuery<ODocument> query = new OSQLSynchQuery<ODocument>(queryString);
List<ODocument> result = db.command(query).execute(params);

[...]

PR #4777, opened 2015-08-12
Fix for setting values directly in array subscripts in ODocument

mmacfadden:
This fixes an issue where it is currently impossible to set the value at a position in an embedded list using the ODocument.field(String, Object) method.  The details of the issue are outlined in issue #4727.


mmacfadden:
Close.  Will resubmit to fix some issues.

Issue #4272, opened 2015-06-02
Classes lose custom fields after import / export operations


Ivanov-Yuriy:
I use 
OrientDB Server v2.1-rc3 (build UNKNOWN@r${buildNumber}; 2015-05-22 16:53:13+0000) 

orientdb> **create database remote:localhost/test root password plocal**
Creating database [remote:localhost/test] using the storage type [plocal]...
Connecting to database [remote:localhost/test] with user 'admin'...OK
Database created successfully.

Current database is: remote:localhost/test

orientdb {db=test}> **create class testCust**

Class created successfully. Total classes in database now: 11

orientdb {db=test}> **alter class testCust custom myCustomField = tro-lo-lo**

Class updated successfully

orientdb {db=test}> **select customFields from (select expand(classes) from metadata:schema) where name = 'testCust'**

----+------+-------------------------
\#  | @ CLASS|customFields
----+------+-------------------------
0   |null  |{myCustomField=tro-lo-lo}
----+------+-------------------------

1 item(s) found. Query executed in 0.017 sec(s).

orientdb {db=test}> **export database D:/test**
Exporting current database to: database D:/test in GZipped JSON format ...

Started export of database 'test' to D:/test.json.gz...
Exporting database info...OK
Exporting clusters...OK (12 clusters)
Exporting schema...OK (11 classes)
Exporting records...
- Cluster 'internal' (id=0)...OK (records=3/3)
- Cluster 'index' (id=1)...OK (records=5/5)
- Cluster 'manindex' (id=2)...OK (records=1/1)
- Cluster 'default' (id=3)...OK (records=0/0)
- Cluster 'orole' (id=4)...OK (records=3/3)
- Cluster 'ouser' (id=5)...OK (records=3/3)
- Cluster 'ofunction' (id=6)...OK (records=0/0)
- Cluster 'oschedule' (id=7)...OK (records=0/0)
- Cluster 'orids' (id=8)...OK (records=0/0)
- Cluster 'v' (id=9)...OK (records=0/0)
- Cluster 'e' (id=10)...OK (records=0/0)
- Cluster 'testcust' (id=11)...OK (records=0/0)

Done. Exported 15 of total 15 records

Exporting index info...
- Index OUser.name...OK
- Index dictionary...OK
- Index ORole.name...OK
OK (3 indexes)
Exporting manual indexes content...
- Exporting index dictionary ...OK (entries=0)
OK (1 manual indexes)

Database export completed in 135ms

[...]

Issue #4505, opened 2015-07-03
New method: ODocument.eval()


lvca:
We're deprecating `ODocument.setAllowChainedAccess(true)` to avoid `ODocument.field()` is able to parse paths, but rather single fields. We could take this chance to introduce the new method `ODocument.eval(String expression)` that evaluate a SQL statement against current document. Example:

```java
String countryName = doc.eval("city.country.name");
```

```java
Collection<String> bestFriendNames = doc.eval("out_Friends[0-9].name");
```

lvca:
Related to #4236

PR #4800, opened 2015-08-15
Fixes setting values in arrays with SQL Updates (#4272)

mmacfadden:
Fixes a bug where the following SQL would fail:
```SQL
UPDATE test SET data.array[0] = 5
```

See issue #4727. For details.

### DIFF
--- a/core/src/main/java/com/orientechnologies/common/collection/OMultiValue.java
+++ b/core/src/main/java/com/orientechnologies/common/collection/OMultiValue.java
@@ -220,6 +220,26 @@ public class OMultiValue {
     }
     return null;
   }
+  
+  /**
+   * Sets the value of the Multi-value object (array or collection) at iIndex 
+   * 
+   * @param iObject
+   *          Multi-value object (array, collection)
+   * @param iValue
+   *          The value to set at this specified index.      
+   * @param iIndex
+   *          integer as the position requested
+   */
+  public static void setValue(final Object iObject, final Object iValue, final int iIndex) {
+    if (iObject instanceof List<?>) {
+      ((List<Object>)iObject).set(iIndex, iValue);
+    } else if (iObject.getClass().isArray()) {
+      Array.set(iObject, iIndex, iValue);
+    } else {
+      throw new IllegalArgumentException("Can only set positional indices for Lists and Arrays");
+    }
+  }
 
   /**
    * Returns an <code>Iterable<Object></code> object to browse the multi-value instance (array, collection or map)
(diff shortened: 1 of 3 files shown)


lvca:
Great, thanks!

nagarajasr:
@mmacfadden you might want to consider the following points in your PR:

1. the syntax doc.field('a.b.c') is being deprecated in favour of the new eval() method. see #4505
2. IMO stronger type checking is needed in OMultiValue.java to better handle different types of lists

mmacfadden:
@nagarajasr  Thanks for the comments.  I have  a couple questions:

1.  

Is the new eval field implemented yet?  I took a look at the issues, and I pulled the code.  I didn't see the eval method in the ODocument class.  I wanted to reiterate that this fix was not really intended to fix an issue with ODocument.  Currently the SQL Update Executors uses the field method on ODocument when executing a set (and other commands) when executing the SQL command.  It was this I was trying to fix.  Therefore, if and when the eval method replaces the field command, we will just need to make sure the executors are updated to still work.  Perhaps I should add a unit test from the SQL perspective (which I actually already have) to make sure this happens properly.

2.
With regards to OMultiValue.  Are you talking about my new method specifically, or are you referencing the class in general.  What I did was take a look at the getValue() method, and saw what it handled.  In terms of list(ish) type things it used java.util.List and Arrays.  The only other things were maps, and iterators, which are things you can't "set" positionally with a square bracket index notation.  What do you feel are the issues?  I am new to the codebase, I can take a stab if I understand what is needed.

Thanks again.

nagarajasr:
1. you're right, it doesn't look like the eval has been implemented yet. but i noticed that @lvca had tagged this issue as 2.2 which is also the release that eval is planned to be implemented eventually. so i'm a bit surprised that a certain usage of a method is being deprecated for an upcoming release and at the same time PRs using that method in deprecated way are being merged for the same release.

2. Sorry for not being more specific. My comment was regarding setValue(), i was concerned about schema-full properties where there is a linked type for the list elements. Perhaps that type checking is being performed in the parser/ executor and does not need explicit handling in OMultiValue.java. Maybe you could add some more test cases to cover updating schema-full embedded list properties?

mmacfadden:
For number 1.  That makes sense.  I suppose this could be back ported into a 2.1.1 release.  That is actually where we would need it for now, since that is where the bug exists.  When we move to eval, we will have to change they way the SQL Executor does its job any way as you point out.  So really this might be a hotfix for the 2.1.x branch.  How do we go about getting it applied as such?

For number 2.  Ok.  I will add a few more unit test in that regard.

nagarajasr:
@lvca could you tag this PR for 2.1.1 milestone ?

mmacfadden:
@lvca any thoughts on a 2.1.1 backport?

lvca:
Yes, I can do that.

lvca:
Done ;-)

PR #4850, opened 2015-08-25
dynamicImport added to avoid ClassCastException in OSGi mode

ylopotun:
dynamicImport directive added to avoid ClassCastException when running in OSGi mode for the classes loaded dynamically via ServiceLoader. Package export added.


lvca:
@jdillon WDYT about this?

PR #4860, opened 2015-08-25
Fix #4247

### LINKED ISSUE #4247: Inconsistent query results depending on order of RIDs in array
I have encountered a bug both in orientdb-server v.2.0.5 and in v.2.0.10. It has to do with querying an edge depending on the order of vertices in an array. In the example below:

- Query 1 is correct. Vertex has the specified edge.
- Query 2 is correct. Vertex does not have an edge.
- Query 3 is correct. When querying multiple vertices via an array one has an edge and the other doesn't so one shows null values and the other one has populated values.
- Query 4 is wrong. The result set shows populated edges for both vertices even though one vertex has no edge! The only thing that changed in this query was the order of vertices in the array.

```
OrientDB console v.2.0.10 (build UNKNOWN@r${buildNumber}; 2015-03-12 22:59:10+0000) www.orientechnologies.com
Type 'help' to display all the supported commands.
Installing extensions for GREMLIN language v.2.6.0

orientdb {db=river}> select first(out('rProductChannelProduct')).SKU as SKU, first(out('rProductChannelProduct')).Brand as Brand, @rid from #19:244441

----+------+----+----------+----------
#   |@CLASS|SKU |Brand     |rid
----+------+----+----------+----------
0   |null  |1111|BodyGuardz|#19:244441
----+------+----+----------+----------

1 item(s) found. Query executed in 0.003 sec(s).
orientdb {db=river}> select first(out('rProductChannelProduct')).SKU as SKU, first(out('rProductChannelProduct')).Brand as Brand, @rid from #19:326589

----+------+----------
#   |@CLASS|rid
----+------+----------
0   |null  |#19:326589
----+------+----------

1 item(s) found. Query executed in 0.002 sec(s).
orientdb {db=river}> select first(out('rProductChannelProduct')).SKU as SKU, first(out('rProductChannelProduct')).Brand as Brand, @rid from [#19:326589, #19:244441]

----+------+----------+----+----------
#   |@CLASS|rid       |SKU |Brand
----+------+----------+----+----------
0   |null  |#19:326589|null|null
1   |null  |#19:244441|1111|BodyGuardz
----+------+----------+----+----------

2 item(s) found. Query executed in 0.003 sec(s).
orientdb {db=river}> select first(out('rProductChannelProduct')).SKU as SKU, first(out('rProductChannelProduct')).Brand as Brand, @rid from [#19:244441, #19:326589]

----+------+----+----------+----------
#   |@CLASS|SKU |Brand     |rid
----+------+----+----------+----------
0   |null  |1111|BodyGuardz|#19:244441
1   |null  |1111|BodyGuardz|#19:326589
----+------+----+----------+----------

2 item(s) found. Query executed in 0.002 sec(s).
orientdb {db=river}>
```

lelonek1:
This fixes #4247, which was caused by incomplete removal of aggregation support in 5ca7fce7c6f3200bb0dee78f3082ddd4f093e1c9.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/functions/coll/OSQLFunctionFirst.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/functions/coll/OSQLFunctionFirst.java
@@ -34,7 +34,7 @@ import com.orientechnologies.orient.core.sql.functions.OSQLFunctionConfigurableA
 public class OSQLFunctionFirst extends OSQLFunctionConfigurableAbstract {
   public static final String NAME  = "first";
 
-  private Object             first = this;
+  private Object             first;
 
   public OSQLFunctionFirst() {
     super(NAME, 1, 1);
@@ -50,9 +50,7 @@ public class OSQLFunctionFirst extends OSQLFunctionConfigurableAbstract {
     if (OMultiValue.isMultiValue(value))
       value = OMultiValue.getFirstValue(value);
 
-    if (first == this)
-      // ONLY THE FIRST TIME
-      first = value;
+    first = value;
 
     return value;
   }


core/src/main/java/com/orientechnologies/orient/core/sql/functions/coll/OSQLFunctionFirst.java
```diff
@@ -50,9 +50,7 @@ public Object execute(Object iThis, final OIdentifiable iCurrentRecord, Object i
     if (OMultiValue.isMultiValue(value))
       value = OMultiValue.getFirstValue(value);
 
-    if (first == this)
-      // ONLY THE FIRST TIME
-      first = value;
+    first = value;
```
lvca: This line assign each value to the "first" variable. So it will have the last value.

lvca:
Thanks for the PR, but I found the implementation was an hybrid between inline and aggregate. Fixed in https://github.com/orientechnologies/orientdb/commit/0cfaa58d80b3649b6ce9089e8e6bd0930e885094.

Issue #4885, opened 2015-08-30
SQL UPDATE SET fails when Strings contains quotes


mmacfadden:
SQL UPDATE SET statements are currently failing in 2.1.  

```Java
OCommandSQL command = new OCommandSQL("UPDATE test SET text = :text");
Map<String, Object> params = new HashMap<>();
params.put("text", "single \"");    
db.command(command).execute(params);
```

A string with a single quote produces an error: 
- "java.lang.IllegalStateException: Missing closed string character: '"'

A string with a quoted string in it produces the error: 
- Error on parsing command at position `#32`: Invalid keyword VALUE" STRING". Use UPDATE <class>|cluster:<cluster>> [SET|ADD|PUT|REMOVE|INCREMENT|CONTENT {<JSON>}|MERGE {<JSON>}] [[,] <field-name> = <expression>|<sub-command>]* [LOCK <NONE|RECORD>] [UPSERT] [RETURN <COUNT|BEFORE|AFTER>] [WHERE <conditions>]

I assume the issue is in the parser, but I haven't debugged much yet.  I have uploaded a unit test to a gist for review.

https://gist.github.com/mmacfadden/2be681114850ef30a6d0

mmacfadden:
So far I have determined that the OUpdateStatement.replaceParameters method is calling OUpdateOperations.replaceParameters which is turning 

```
SET text = :text
```
into this:
```
SET text = "quoted "value" string"
```
The resulting update statement then looks like this:
```
UPDATE test SET text = "quoted "value" string"
```

which is problematic.  You wind up with SET text = "quoted" followed by what appears to be a keyword called 'value'.

I am still digging.

mmacfadden:
It appears that the issue may lie in the OInputParameter.toParsedTree.  If the bound parameter is a string, it is just returns the string directly, where is probably needs to escape the quotes.  I am going to test out a fix in the morning and submit a PR if it works.

mmacfadden:
Upon further investigation.  It appears that this is fixed in the current develop branch.  I am going to do a PR that just adds the unit tests to the OCommandExecutorSQLUpdateTest class.

lvca:
@luigidellaquila Please could you merge this in 2.1.x branch so it will be available on the 2.1.1 release today?

luigidellaquila:
yes, I'm merging it right now

mmacfadden:
[...]

PR #4888, opened 2015-08-30
Adds unit tests for named parameters containing strings with quotes.

mmacfadden:
Issue #4885 occurs in 2.1.  Named parameters that are set to Strings with quotes in them can be problematic.  This issue has already been fixed in the develop branch.  This pull requests simply adds 2 units tests that currently fail in 2.1, but that pass in 2.2.0-SNAPSHOT.  This will help ensure there are no regressions in this area going forward.


luigidellaquila:
Merged, thanks a lot!

PR #4956, opened 2015-09-11
fix dependency on sun.misc that prevents deployment in eclipse

b4mse:
by making the import on sun.misc optional, the orientdb-tools may be
used in eclipse without dependency problems. 
This change was already made on core/pom.xml but is missing in tools/pom.xml


laa:
Thank you !

lvca:
Ported also on "develop".

PR #4970, opened 2015-09-15
HTTP API - can't create a property with LINK without linked class

matanshukry:
Although I couldn't find it in the documentation, it seems possible to create a property with a LINK type, without the link (parameter) type. This is also useful to me, allowing me to have a field that can be a link to different classes, dependent on the row.

However, using the HTTP api, I get an exception saying the linked (parameter) type is missing.
This PR fixes it, only checking and using the linked (parameter) type if provided.

Also, the documentation needs a bit of fixing:
1. in ```http://orientdb.com/docs/2.0/orientdb.wiki/SQL-Create-Property.html``` , linked-class is declared as "the contained class in containers (see above).". However, it is also the linked class in link types.
2. in ```http://orientdb.com/docs/last/OrientDB-REST.html#property``` , The syntax is both wrong (and the examples are correct).
Currently it is:
```
/property/<database>/<class-name>/<property-name>/[<property-type>]
```

It should be though (the correct syntax is also used in the examples, by the way):
```
/class/<class-name>/<property-name>/[<property-type>]/[<linked-type>]
```


lvca:
We're fixing that test that relies on a timeout... Your PR is perfect. Thanks!

PR #4998, opened 2015-09-22
Add test to cover concurrent transaction usage

Jotschi:
I created these test cases while researching how to use orientdb transactions in a concurrent context. I was unable to find any testcase that covers those cases when using the blueprint API. Lastly i was not able to find a testcase that tries to provoke a OConcurrentTransactionException. Instead i only found testcases that assure no OConcurrentTransactionException occur.

I noticed that it is mandatory to load the OrientVertex within the scope of the transaction. I was unable to 
reuse the OrientVertex that was created in the main thread from within one the spawned threads. 
#L43: ```OrientVertex vertex = mainTx.addVertex(null, PROPERTY_NAME, "init");```

Instead i had to load the OrientVertex by id.
#L46 ```OrientVertex secondVertexHandle = tx.getVertex(recordId);```

My questions: 

* Is the OrientVertex Class Threadsafe?
* Should it be possible to reuse OrientVertex objects across multiple threads? If so how?
* Afaik OrientGraph is not Threadsafe but in my code calls to ```graphFactory.getTx();``` are placed within the Threads Runnable. Is is the correct way to create new Transactions or must i synchronize the access to the factory manually?

### DIFF
--- a/graphdb/src/test/java/com/orientechnologies/orient/graph/blueprints/ConcurrentTxTest.java
+++ b/graphdb/src/test/java/com/orientechnologies/orient/graph/blueprints/ConcurrentTxTest.java
@@ -1,0 +1,186 @@
+package com.orientechnologies.orient.graph.blueprints;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.orientechnologies.orient.core.exception.OConcurrentModificationException;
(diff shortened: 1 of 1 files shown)


laa:
Hi,
Graph elements are not thread safe so you can not share them between threads.
About your last question - yes all is correct.

About your test, I will merge it, but honestly I would use future instead of tracking exception in atomic variables.

laa:
Fixed few minor issues and merged. Thank you !!

Jotschi:
@laa 
I see. I was just wondering whether it would be possible to detach the graph element from the used graph and reattach it within the thread. I tried this but it failed.
Your last answer is not 100% clear to me. Do i need to synchronize manually?

This means that ```OrientVertex secondVertexHandle = tx.getVertex(recordId);``` is the prefered way of loading a known Orientvertex in order to (re)use it within a thread?

PR #5020, opened 2015-09-25
Allow document inside an embedded

matanshukry:
The problem can be demonstrated in the following scenario:

```
CREATE CLASS User EXTENDS V
INSERT INTO User set id=0, name="Joe"
INSERT INTO User set id=1, name="Mat"

UPDATE User SET extra={"tag": (SELECT FROM User LIMIT 1)}

SELECT FROM User
```

And the result is:

```
----+------+------+----+----+----------------------------------
#   |@RID  |@CLASS|id  |name|extra
----+------+------+----+----+----------------------------------
0   |#11:9 |user  |0   |Joe |{tag=sql.SELECT FROM User LIMIT 1}
1   |#11:10|user  |1   |Mat |{tag=sql.SELECT FROM User LIMIT 1}
----+------+------+----+----+----------------------------------
```

this PR fixes it.

Also, I'm not 100% sure if this is a bug fix or a feature; That is, I'm not sure there was ever an intention to allow records inside a map. Either way, I hope it's accepted.

p.s.
I am having some troubles running the tests, so please make sure this doesn't make any problems; If it does, tell me which test and I'll fix the PR (I seems to be able to run any individual test alone, just not all of them).


luigidellaquila:
Hi @matanshukry 

Thank you very much.
This is definitely a bug, I'll review your PR in short and I'll merge it

Thanks

Luigi

lvca:
Hey @matanshukry,
Since there isn't a test case I was trying your branch but it's very old and not aligned to develop. Please could you resend a more updated PR ? If you can't write a test case I can do that.

matanshukry:
The travis fails at orientdb-jdbc vm crash. All tests are running fine on my pc.

matanshukry:
Merged, and also added a test case

Issue #4979, opened 2015-09-17
[Feature Request] Field alias and/or view


matanshukry:
I was wondering if there is any desire for a field alias or a view feature.
#### Field Alias

The field alias I was thinking about is a simple second name for a field. Unlike the current aliases in the query, the field alias I am referring to can be used to query a field with a different name. Something like

```
CREATE CLASS User EXTENDS V
CREATE PROPERTY User.id INTEGER
CREATE PROPERTY User.username STRING
CREATE PROPERTY ALIAS User.key_field id

CREATE VERTEX User SET id=0, username="Mat";

SELECT key_field FROM User
// Will select key_field with value 0
```
#### View

I am talking about the same video as most RDBMS have these days; That is, a simple object that stores a (probably optimized-ish version of) query. Something like:

```
CREATE PROPERTY User.id INTEGER
CREATE PROPERTY User.username STRING
CREATE PROPERTY ALIAS User.key_field id

CREATE VIEW UserBasic AS Select id as key_field from USER
```
#### Reasoning

My specific use case for any of the above is I have a field of type LINK, that can point to different classes, dependent on the row. When querying the field though, I would like to query some small information, displaying a short text (name) and linking it to the object page.

Currently, I am simply querying everything, and checking what isn't null at the application level.
Something like (assuming 'user' has a field 'friend' that can point to either 'user' or 'animal')

```
SELECT user_id,username, friend.user_id, friend.username, friend.animal_id, friend.animal_name FROM User
```

p.s.
I don't mind working on this feature, assuming it will be merged and I won't work for nothing.

luigidellaquila:
Hi @matanshukry 

I think views would be nice to have. I have some more doubts about aliases, even though I understand your use case.
My only concern is about record identities in views: in relational databases the problem does not arise, because join operations are based on property values, but in OrientDB relationships rely on immutable RIDs, so how should a view manage its RIDs? 
In particular in the general case, when the view is a projection/aggregation of data, how should RIDs be assigned and re-calculated when the query result changes? 

About the merge, of course we do not like to waste contributors resources, so any piece of code that makes sense and that is well implemented will eventually be merged. 
The problem is "when"...
What I mean is that the roadmap for 2.2 is well defined and very rich, so it's unlikely that we will be able to merge this feature in that release. Maybe it will be merged in 3.0, more likely in 3.1, so take it into consideration in your decision.

Thanks

Luigi

matanshukry:
[...]

PR #5031, opened 2015-09-29
View object

matanshukry:
PR for view object. ( talked about at #4979 )
### Creating a view

```
CREATE VIEW <view-name> AS <query>
```
### Dropping a view

```
DROP VIEW <view-name>
```
### Alter a view

```
ALTER VIEW <view-name> [NAME <new-name>] [QUERY <new-query>]
```
### Listing all views (console command)

```
list views
```

Example result:

```
VIEWS
-----------------+----------------------------------------------------------------------------------+
 NAME            | QUERY                                                                             
-----------------+----------------------------------------------------------------------------------+
 userDisplay     | SELECT name, color.append(" ").append(name) as display_name FROM User            |
-----------------+----------------------------------------------------------------------------------+
 TOTAL = 1                                                                                          |
-----------------+----------------------------------------------------------------------------------+
```
### Example usage:

```
CREATE CLASS User
INSERT INTO User SET name="Mat", color="Blue"
INSERT INTO User SET name="Jason", color="Green"
INSERT INTO User SET name="Samantha", color="Red"

CREATE VIEW userDisplay AS SELECT name, color.append(" ").append(name) as display_name FROM User

SELECT FROM userDisplay WHERE name in ["Mat", "Samantha"]
```

result:

```
----+------+--------+------------
#   |@CLASS|name    |display_name
----+------+--------+------------
0   |null  |Mat     |Blue Mat    
1   |null  |Samantha|Red Samantha
----+------+--------+------------
```
### What's left
1. Some tests for the view; Would be nice if someone else can implement them!
2. Currently there is no check for the query value. That is, you can enter any value you would like in the query, and it will be saved. The error will only show up when actually querying the view. To be honest though I didn't really look at it, due to some time constraints. I'll try to look at it in the next few days.


PhantomYdn:
Cool:)
1) But is it persistent view or dynamic? 
2) Is there support for views usage in queries?

If answers are 1) not persistent and 2) no : it looks like just registry for queries, but not true views.

matanshukry:
I'm not entirely sure what are you referring to; What is persistent/dynamic view? what is your definition for complex query?

Anyway, I think what you're asking if this is a view or a materialized view; If so, this is a simple view, not even insert-able.

"My" definition for view/materialized view, just to be clear:
https://en.wikipedia.org/wiki/View_(SQL)
https://en.wikipedia.org/wiki/Materialized_view

PR #5035, opened 2015-09-30
Fixed OIOUtils.getTodayWithTime so it parses using 24hr clock.

allanmoso:
OIOUtils.getTodayWithTime was parsing the time with format 'HH:mm:ss' but was converting it to a 12 hour clock.

So 17:00:00 was converted to 5:00 AM instead of 5:00 PM.


lvca:
Sorry I reverted it because you did the PR against "master" instead of "develop". Take a look at: http://orientdb.com/docs/last/Contribute-to-OrientDB.html

allanmoso:
I am sorry about that. I will resubmit a PR request on develop branch.

PR #5042, opened 2015-10-01
Fixed OIOUtils.getTodayWithTime so it parses using 24hr clock.

allanmoso:
OIOUtils.getTodayWithTime was parsing the time with format 'HH:mm:ss' but was converting it to a 12 hour clock.

So 17:00:00 was converted to 5:00 AM instead of 5:00 PM.


lvca:
Perfect, thanks!

allanmoso:
@lvca Is this fix going to be on the next hotfix release? When is the next hotfix release?

This issue affects the way OAutomaticBackup works in that the 'firstTime' behaves differently depending on what time of the day the orientdb server is started (since the parameter discards the AM/PM aspect of the time).

PR #5104, opened 2015-10-12
added $matches and $group context vars

micha-nerdlichter:
Adds context vars ```$matches``` and ```$group``` for ```MATCHES``` operator.
Example in GreatfulDeadConcerts:
```sql
select name, $matches, $groups from V where name matches "(Garcia.*|JA.*?)(ROE)?"
```
outputs information about the group that matched:

image

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/operator/OQueryOperatorMatches.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/operator/OQueryOperatorMatches.java
@@ -19,7 +19,10 @@
   */
 package com.orientechnologies.orient.core.sql.operator;
 
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.List;
+import java.util.ArrayList;
 
 import com.orientechnologies.orient.core.command.OCommandContext;
 import com.orientechnologies.orient.core.db.record.OIdentifiable;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/orientechnologies/orient/core/sql/operator/OQueryOperatorMatches.java
```diff
@@ -66,6 +69,21 @@ private boolean matches(String iValue, String iRegex, OCommandContext iContext)
...
+			List<String> iMatches = new ArrayList<String>();
+			List<Integer> iGroups = new ArrayList<Integer>();
+			for(int i = 1; i <= m.groupCount(); i++) {
+				String match = m.group(i);
+				if (match != null) {
+					iMatches.add(match);
+					iGroups.add(i-1);
+				}
+			}
+			iContext.setVariable("matches", iMatches);
```
nagarajasr: the context variable may already have an existing value, in which case you might want to add to the existing list instead of losing the existing context variable values.

nagarajasr: also while we're at it, we should consider to refactor all of the available context variable names into a constants file so that it becomes easier to document what context variables are available. it should probably be a new issue though.

lvca:
@micha-nerdlichter Thanks for the PR, but I'm forced to discard it because you used master branch, while develop should be used. Take a look at: http://orientdb.com/docs/last/Contribute-to-OrientDB.html. Please could you resent the PR to develop after you have pulled last version?

Small note: could you call the context variable with a `matches` prefix? Like: `$matches_key` and `$matches_group` ? I'm not sure for `$matches_key` name, if you have something better let's propose it.

PR #5142, opened 2015-10-17
Corrected grammar in settings

smolinari:
I tried to also get some standardization in the settings. Rules I set and followed.

1. All settings end with a period, despite the setting description being a full sentence or not.
2. All references to the defaults or setting values are in parenthesis. Like (in seconds), (in ms), (default is set to true), etc.

I am also not certain about the correctness of some of my changes. Please check lines

101 for correctness. "doesn't allow /allows"?
117 for correctness. Change of the storage configuration?
119 no correction made on my part, just a note. "nothing" isn't really the right word for the setting. "None" would be better, but I realize that means the code would need to change too.
122 the same note for line 119 goes for this setting too.
149 not sure what the intent was with "Can be more or less.", so I just removed it. I would assume the setting controls the max size of the WAL.
266 Says "false by default", but is set to true, I think?
294 for correctness.
325 for correctness.
493-501 descriptions aren't very descriptive.

Hope I could help. 

Scott

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/config/OGlobalConfiguration.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/config/OGlobalConfiguration.java
@@ -1,75 +1,75 @@
-    /*
-        *
-        *  *  Copyright 2014 Orient Technologies LTD (info(at)orientechnologies.com)
-        *  *
-        *  *  Licensed under the Apache License, Version 2.0 (the "License");
-        *  *  you may not use this file except in compliance with the License.
-        *  *  You may obtain a copy of the License at
-        *  *
-        *  *       http://www.apache.org/licenses/LICENSE-2.0
-        *  *
-        *  *  Unless required by applicable law or agreed to in writing, software
-        *  *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


lvca:
Hey @smolinari the formatting is completely gone and it's impossible for me compare what you changed. I have to be sure you didn't change settings names.

smolinari:
I am pretty sure, I didn't change anything outside of the settings text.

Scott

smolinari:
Wait, I see what you mean. Hmm...I think that is because I cut and pasted from the Github editor and my editor zapped the formatting. Let me correct it. 

Scott

smolinari:
Ok. I've reformatted everything. 

I hope it is ok that I used 4 spaces for all the indents. Is it really a standard in the ODB code to use 4 spaces for some indents and 2 for others? Do you have a coding standard written somewhere?

To be honest, the whole file still looks sloppy to me. For instance, why do some "if" statements use brackets and others don't? Why do some configs have comments saying the same thing as the description text and others don't (I'd get rid of the comments, since they add no benefit)? I'd also make it standard that the next line for every configuration setting starts with the description text. Some do, some don't. 

If you want me to go through and "standardize" the file, let me know.

Scott

lvca:
Merged, thanks!

PR #5234, opened 2015-10-27
Changed to use hazelcast and hazelcast-cloud instead of hazelcast-all

w3aponx:
As discussed with lvca, changed to use `hazelcast` and `hazelcast-cloud` instead of `hazelcast-all` to reduce the number of dependencies being pulled in unnecessarily.  Also reduces the risk of conflicts with other projects using Hazelcast but adding only dependencies on individual modules.
Unit tests passed successfully.  Consider back-porting to 2.1.x if there is going to be another release before 2.2.0 is released.

### DIFF
--- a/distributed/pom.xml
+++ b/distributed/pom.xml
@@ -118,7 +118,12 @@
         </dependency>
         <dependency>
             <groupId>com.hazelcast</groupId>
-            <artifactId>hazelcast-all</artifactId>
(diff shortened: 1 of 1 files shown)


w3aponx:
Looks like the CI build failures have nothing to do with the pull requests, but rather the JDBC Driver is failing to build for some reason.  Correct?

lvca:
Thanks.

Issue #5238, opened 2015-10-28
Fix for count issue within transaction wasn't merged into develop and/or 2.1.x


wouterv:
The following pull request was approved and merged into 2.0.x. But now we are moving to 2.1.x client, we are seeing the errors that made me create this fix.

I looked into the code, and the fix simply isn't there, so I believe @lvca forgot to merge the fix into develop and/or the 2.1.x branch.

See pull request: https://github.com/orientechnologies/orientdb/pull/3847
And issue: https://github.com/orientechnologies/orientdb/issues/3846

wouterv:
Created pull request: #5242, please also merge into develop

lvca:
Sorry, I forgot to port this to 2.1.x and develop.

PR #5242, opened 2015-10-29
Fix for #5238. Count was broken when counting new vertices in same tr…

wouterv:
Please also merge into develop!


lvca:
I didn't see this. Merged the old one, should be the same.

wouterv:
Yep, new one was just a cherrypick of the old one.

Issue #1114, opened 2012-12-10
UPDATE EDGE statement [moved]


lvca:
This is [Issue 1114](http://code.google.com/p/orient/issues/detail?id=1114) moved from a Google Code project.
Added by 2012-10-14T22:09:41.000Z by [infected...@gmail.com](http://code.google.com/u/105221022043430137887/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Enhancement, Priority-Low, v2.1.0

### Original description

    Having an UPDATE EDGE statement would be nice because there is less overhead then deleting and recreating the edge.
    
    Discussion here: https://groups.google.com/forum/#!topic/orient-database/Z5ZElaxVSZ0

lucasxteixeira:
+1

thedanielhanke:
+1

luigidellaquila:
Same for UPDATE VERTEX

AlexFiverr:
+1

yujinlim:
+1

iforwms:
+1

AlexFiverr:
hey, any updates on the version this feature will be in? 
thanks!

pisaruk:
+1

StarpTech:
+1

liorn:
+1

FGRibreau:
:+1:

sebastiandev:
+1

smolinari:
:+1:

matanshukry:
As far as I can tell, update works (including in/out):
```
UPDATE FRIEND SET in=#11:5 WHERE in.name="Samantha" and out.name="Mat"
```

Although, using a query inside the set clause for in/out will cause it to return array, which is not the desired behavior. Example:
```
UPDATE Friend SET in=(SELECT FROM User WHERE name="Jason") WHERE in.name="Samantha" and out.name="Mat"
```
When later querying the table Friend, we can see column "in" will contain [#11:4], emphasis on the brackets [], which means an array.


Upsert also works, although when creating a new vertex it doesn't set the in/out properties.
You could set the in/out properties your self, like this:
```
UPDATE Friend SET comment="Wazzzaup", in=#11:5, out=#11:6 UPSERT WHERE in.name="Jason" AND out.name="Mat"
```
It will result in a longer query when using sub-query for ```in=``` and ```out=```, but at least it works (sub-query has same problem as above).

To summarize, the only problem I can see is using a sub-query for in/out. So:

[...]

PR #5318, opened 2015-11-14
Issue update edge

matanshukry:
Related to #1114


matanshukry:
Travis shows a build error, but I successfully ran all maven tests, and I couldn't figure out the real problem from the log, the test just seems to crash (worked fine on my pc)

matanshukry:
Closing this PR, we don't want to allow people to directly change in/out properties.

Issue #5328, opened 2015-11-16
Drop database but preserve files on disk


grossvater:
There is a way to achieve this?

laa:
What do you mean, database is files on the disk and nothing more.

grossvater:
Detach, unregister, but do not delete, that's it, a way to tell the server forget about this database.

smolinari:
How about just moving the related database files out of the source directory? Wouldn't that work?

Scott

grossvater:
@smolinari I need to detach while the server is running

I'll try to explain my scenario:
- I have a relative big database (~20G) that I want to update
- since using remote connections proved to be slow and sometimes causes out of memory when the transaction is big, I decided to update the database via a local connection
- to achieve this, I need a way to tell the server to somehow release the database (drop is not working here since it deletes the files).
- stopping the server is also not an option, because the update must be "live", being part of the normal functioning, not a maintenance procedure
- besides that, I would like to keep the files in the same physical location during update, avoiding thus to copy / move 20G.

smolinari:
Ah, ok. Now your requirements are a bit clearer.

What is is the current version of your database and to what new version do you want to upgrade to?

Usually, if the version numbers aren't too far apart, the update process is through a drop in process and is pretty quick to do. Theoretically, you could do it within seconds. However, you still have to stop the server for that time to do it. 

I think the only way you can hot-swap upgrade is with a distributed system, which means, you need extra nodes with the data, which means copying the data. 

I am not that experienced with ODB, so don't take my word for it either. Maybe @laa can also give you some tips.

Scott

grossvater:
@smolinari update != upgrade, but update = massive delete, insert, update

I run community edition v2.1.5.

grossvater:
Meanwhile I try myself to find an workaround by freezing the database, hard linking all the files into another (snapshot) directory, then drop the database. At the end of the process I expect to find the files intact into snapshot directory, but this doesn't happen, it seems that before dropping the database, the server truncates some files, which obviously affects the hard linked files.

The lines to be blamed seems to be inside `com.orientechnologies.orient.core.storage.cache.local`:

```
  private String doDeleteFile(int fileId) throws IOException {
    if (isOpen(fileId)) {
        truncateFile(fileId);
    }
    ...
 }
```

smolinari:
Ah, yes. Sorry about that. Totally misunderstood. 

Have you tried doing the updates with batch?

[...]

PR #5345, opened 2015-11-18
wow cache: fix truncate on delete

grossvater:
I would like to have it integrated also on 2.1.x branch.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
@@ -1226,8 +1226,10 @@ public class OWOWCache extends OAbstractWriteCache implements OWriteCache, OCach
   }
 
   private String doDeleteFile(int fileId) throws IOException {
-    if (isOpen(fileId))
-      truncateFile(fileId);
+    if (isOpen(fileId)) {
+    	flush(fileId);
+        close(fileId, false);
+     }
 
     final OFileClassic fileClassic = files.remove(fileId);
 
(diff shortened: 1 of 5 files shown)


lvca:
@laa could you please review it?

laa:
Hi sorry but I do not see a point in truncate in a case of delete.

grossvater:
@laa Please take a look at this issue: https://github.com/orientechnologies/orientdb/issues/5328
and see which is my scenario: when the files are hard linked, deleting them doesn't actually delete the data if there are other links to the file. If we truncate at delete, the other links will point to empty / corrupted files and this is what I try to avoid. I really need this feature, because I do not want to maintain a fork just to have this. Thanks.

laa:
@grossvater I have read the issue. But files do not reflect the real state of the database and update them outside of server control will lead to data corruption.

grossvater:
@laa Regarding the modification outside of the server control, I never do this (I always update the database through local or remote connections), maybe you could rephrase a bit, cause I do no understand what you mean. 
Anyway, all the magic with hard linked files was an workaround to preserve the files after database is dropped. What I actually needed was a way to detach the database (unregister the database and leave the files unchanged), but this feature seems to need more work. So for a while, skipping file truncation will help me a lot, and as far as I remember, all tests passed.

laa:
@grossvater I mean if you access the same files from different JVMs you will have problems with data consistency, because the state of the file on disk does not reflect a current state of a database system.

grossvater:
@laa I do not access the files from different JVM at the same time

laa:
@grossvater OK you persuaded me. I will provide change in next 2.2 snapshot

grossvater:
@laa Great. Thank you!

laa:
@grossvater I just checked , I have already done that in 2.2.x .So you may safely update to 2.2.x and use this feature out of the box.

grossvater:
@laa Super! Thanks!

Issue #4436, opened 2015-06-23
Allow UPSERT of Edges


Bandes:
Currently, UPSERT can only handle vertices. For my purposes, I would like/need to be able to upsert edges as well.

bomgar:
:+1:  would also like to see this

justinkillen:
+1

amyth:
+1

kubz:
+1

strohni:
+1

suyogdilipkale:
I am able to upsert for edge class with following syntax:

UPDATE LikeComment SET out = #21:21, in = #20:20 UPSERT WHERE out = #21:21 and in = #20:20

But When I select source vertex like SELECT FROM #21:21 it does not show me that out edge in result.

karolmie1:
This, if implemented, would support queries like this (multiple edges creation): 

`UPDATE EDGE hasAssignee FROM ( SELECT FROM Project WHERE id=:id ) TO ( SELECT FROM User WHERE login in :login ) UPSERT`

Am I right?

oger-p:
+1

zeeshan2709:
hey,
I am using
 ```update has SET out=#20:16, in=#14:18 UPSERT WHERE out = #14:18 and in = #20:16```
It does not create the edge if not present. Is this feature not yet implemented. If not, can you suggest any workaround for this.
usecase is to create edge only if it does not exist.

karolmie1:
I guess in meantime the good ol' if statement have to suffice

grindhold:
+1

lightjiao:
+1

undecided2013:
+1, this would be great

micha-nerdlichter:
+1

shake10861:
+1

timobanion:
+1

gauravssFlipkart:
+1

triketora:
+1

WalterInSH:
+1

feiYH:
+1

grzegorz-aniol:
+1
I'm really surprised this feature is not available yet in orientdb. How people deal with this problem during cyclic update/import data??

bernierm:
Going on three years now. I personally dealt with this by using something instead of OrientDb.

david-sitsky:
+1.  This is a really important feature for us if OrientDB is to be used in a real-world / real-time mass graph database application.  

[...]

PR #5384, opened 2015-11-26
Allow update edge in/out properties with select query

matanshukry:
Related to #1114 / #4436

p.s.
Currently there seems to be a memory leak in the branch, regardless of my commit.
I'll merge, fix if needed, commit and push when possible.


luigidellaquila:
Hi @matanshukry 

thank you very much, the PR is absolutely correct. There is just a problem with the build, we are fixing it right now.
I'll re-run the build and will merge the PR as soon as the problem is fixed.

Thanks

Luigi

matanshukry:
Merged with develop again, all checks seems to pass now

lvca:
Thank you!

PR #5396, opened 2015-11-29
Implementation of an numerical absolute value function.

mmacfadden:
This PR implements a basic 'abs' function for OrientDB.  It can be use as follows:

```SQL
SELECT abs(-5)
SELECT abs(property) FROM MyClass
```

The function supports the following types:
* null
* Integer
* Long
* Short
* Double
* Float
* BigInteger
* BigDecimal

I believe this covers the main numeric types that OrientDB Supports.  One could argue that Byte could be implemented as well, but I have left that off for now.  If need be that can be added later.

The function as well as the unit tests are supplied.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/functions/ODefaultSQLFunctionFactory.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/functions/ODefaultSQLFunctionFactory.java
@@ -19,6 +19,7 @@ import com.orientechnologies.common.exception.OException;
 import com.orientechnologies.orient.core.exception.OCommandExecutionException;
 import com.orientechnologies.orient.core.sql.functions.coll.*;
 import com.orientechnologies.orient.core.sql.functions.geo.OSQLFunctionDistance;
+import com.orientechnologies.orient.core.sql.functions.math.OSQLFunctionAbsoluteValue;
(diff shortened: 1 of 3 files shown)


lvca:
Thanks for the contribution!

mmacfadden:
Thanks for the quick turn around!  Keep up the good work!

lvca:
Docs updated too.

mmacfadden:
For things like this, I'd be happy to help update the docs as well.  Is any info on how to do that?

luigidellaquila:
Hi @mmacfadden 

to contribute to the docs, you just have to submit PRs to this project

https://github.com/orientechnologies/orientdb-docs/

Thanks

Luigi

mmacfadden:
Thanks!  I saw this repo.  Is there anything that needs to be done to "build" the docs?

luigidellaquila:
Nothing particular to just edit the docs, but if you want to build GitBook I think you'll find answers here:
https://github.com/orientechnologies/orientdb-docs/issues/66
Thanks

Luigi

lvca:
Having also contributions on docs would be even better. This time I did it: https://github.com/orientechnologies/orientdb-docs/blob/master/SQL-Functions.md#abs ;-)

mmacfadden:
@lvca Great.  I can update the docs from now on.

Issue #5443, opened 2015-12-09
Performance can be improved with adding simple caches for reflection calls in Object API


wouterv:
When profiling I found out that several methods used in the Object API use a lot of reflection and that this can cause slowing down because a lot of these calls are slow.

This can be remedied by caching results of these calls. It can be cached for operations for which the result never changes, for instance the Fields of a Class, getting the annotations of a class, etc. 

I propose to implement this at key places in the Object API. I'm preparing a pull request with some simple changes that resulted in a 50% performance gain in some calls for us, especially when the caches have been warmed up.

luigidellaquila:
Hi @wouterv 

It's definitely a good idea. I'm marking it as an enhancement.
A PR will be very appreciated

@tglman I'm assigning this issue to you in case support is needed.

Thanks

Luigi

wouterv:
The PR is already created: #5444

wouterv:
New PR after comments: #5462

wouterv:
PR was merged into 2.1.x

wouterv:
@luigidellaquila Shouldnt this get label 2.1.8?

luigidellaquila:
Hi @wouterv 

Yes, definitely, I missed it because it was not flagged as 2.1.x

@tglman was it also ported to develop?

Thanks

Luigi

tglman:
yes is also in develop.

PR #5444, opened 2015-12-09
object api performance improvements case #5443

wouterv:


lvca:
Hey @wouterv, did you check all tests pass?

wouterv:
Up to server they did, but there was something wrong in the server module
that seemed unrelated to me..
Op 9 dec. 2015 9:56 p.m. schreef "Luca Garulli" <notifications@github.com>:

wouterv:
@lvca I just did a clean checkout of the 2.1.x branch (without these changes) and the same failures occur, so the problem is probably not in my changes.

wouterv:
@lvca ok it looks like the server module just builds using travis, but if you look at the log, somehow the jvm got killed while running orient-tests, I tried to run from the orient-tests module locally with my changes (first did mvn -DskipTests clean install, then mvn clean install -rf :orient-tests), and it doesn't get killed, it builds fine.

lvca:
Yes, Travis kills all the build with exec time >20mins. We're replacing it with CloudBee.

wouterv:
@lvca is there anything else you need for this pull request

tglman:
Hi @wouterv,

This solution use static variable, will be cool if you can move this implementation in some instance linked to OEntityManager ?

wouterv:
@tglman I don't understand, the info stored is static, class information like fields and so on, why do you want to link it to an instance?

tglman:
In some environment like inside an application server or an OSGI environment use static fields can bring to problems, we have some configuration for handle some of this cases for OSGI, so is better not introduce additional static variable and use the existent one, and anyway one of the responsibility of OEntityManager is to handle class definition.

wouterv:
I get the point about OSGi, I will make a rewrite using the OEntityManager.

wouterv:
@tglman Would using OObjectEntitySerializer.OObjectEntitySerializedSchema a good place to put the caches? Looks like it is already done like that for some other stuff.

tglman:
@wouterv yes

wouterv:
@tglman @lvca created new pull request for this: #5462

PR #5462, opened 2015-12-14
object api performance improvements case #5443

wouterv:


tglman:
looks good merglng

smolinari:
Nice work @wouterv!  :+1: 

Scott

PR #5538, opened 2015-12-30
Enhancements to asList and asSet (fixes #5534)

### LINKED ISSUE #5534: Unexpected behavoir of "asList()" with and EMBEDDED ODocument as an argument
I ran into this issue while trying to work around #5532.  In that issue there is a case where the same query is either returning a HashSet or an ODocument.  I was trying to use the "asList()" method as a workaround to ensure that in either case, the value returned would be a list.  While I did always get a list returned, the values in the list were not as expected when the query returned an ODocument (It did work fine for a HashSet).

The "asList()" function (OSQLMethodAsList.java) seems to be intended to convert various collections and single items to a list.  The documentation states that if a single element is passed to this method, a new list will be created with the single element in it.

On line 44, the class checks to see if the value is already a list.  If so, it just returns it.  On line 50 it checks for null, and returns an empty list.  On line 56, it checks to see if the value is a collection and uses the "addAll" constructor of ArrayList to create a new list with all the elements.  On line 62, the class detects an Iterable, and iterates over the value, adding each element to a new list.  Lastly, on line 73 the class assumes that the value is a single element and adds it to a new list.

The problem here is that ODocument itself implements Iterable<Entry<String, Object>>.  Therefore instead of getting a list with a single ODocument init, the caller winds up with a List of Entry<String, Object>.  I suppose there could be a rare case where this is desired, however I believe that the user would expect a list with a document in it.  For all other types, Dates, Numbers, Strings, and even Sets / Lists of documents, you will wind up with the values intact, just in a list.  Pulling apart the document itself into a list, seems strange.  For example, I would not expect the asList method to turn a String into a list of individual characters.

If this is actually the indented behavior, then I would suggest updating the documentation to explicitly state that the ODocument will be converted into a list of Key Value pairs.

In either case, I would be more than willing to either update the code, or the documentation based on your thoughts.  I didn't want to move ahead with either until I understand what the expected behavior is.

Keep up the good work!

mmacfadden:
This PR implements a fix and tests for #5534.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/method/misc/OSQLMethodAsList.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/method/misc/OSQLMethodAsList.java
@@ -19,9 +19,11 @@ package com.orientechnologies.orient.core.sql.method.misc;
 import com.orientechnologies.common.util.OSizeable;
 import com.orientechnologies.orient.core.command.OCommandContext;
 import com.orientechnologies.orient.core.db.record.OIdentifiable;
+import com.orientechnologies.orient.core.record.impl.ODocument;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


lvca:
Awesome, thank you again, specially for all the test cases you provided.

PR #5539, opened 2015-12-30
Enhances the "asMap()" method to handle converting an ODocument

mmacfadden:
The asMap method turns things in to a map.  The behavior is already that if a map is passed in, it simply returns a map.  An ODocument is already essentially a map from field names to field values.  However, ODocument does not implement a Map interface.  It only implements an Iterable interface.  therefore the asMap method does not work well on it.

For an iterable, the asMap method will take even values as keys and odd values as values.  However, the ODocument's Iterable behavoir is Iterable<Entry<String, Object>>.  This means that the iterator is iterating over entries which are already the key / values of the ODocument.

This means that currently if one had the following:

```java
ODocument doc = new ODocument()
doc.field("f1", 1);
doc.field("f2", 2);
```

Then "asMap" would return a Map like this:

```
Map{ 
  Entry("f1", 1) -> Entry ("f2", 2) 
}
```
That is, the Map would be a somewhat random mapping form one key/value entry to another.

This PR uses the ODocument's asMap method to convert to the ODocument to a map in an intelligent way.  Now calling asMap will result in:

```
Map{ 
  "f1" -> 1,
  "f2" -> 2
}
```

This seems much more desirable.

I have also included a unit test for the existing and new behavior.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/method/misc/OSQLMethodAsMap.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/method/misc/OSQLMethodAsMap.java
@@ -16,15 +16,17 @@
  */
 package com.orientechnologies.orient.core.sql.method.misc;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


lvca:
Thank you for this PR, very appreciated.

Issue #5424, opened 2015-12-04
GraphSail is broken because of invalid edge lable encoding 


fabianmeyer:
When running the example code of the Tinkerpop GraphSail outplementation (https://github.com/tinkerpop/blueprints/wiki/Sail-Ouplementation), the OrientBaseGraph throws a: com.orientechnologies.orient.core.exception.OSchemaException: Invalid class name found. Character '%' cannot be used in class name 'http%3A%2F%2Ftinkerpop.com%23knows'

The OrientBaseGraph uses the java.net.URLEncoder to encode the class name of the label. The encoder in fact maps the IRI "http://tinkerpop.com#knows" to "http%3A%2F%2Ftinkerpop.com%23knows", which is an invalid class name.

Example code:
```java
	OrientGraph graph = new OrientGraph("remote:localhost/test");
		try {
			Sail sail = new GraphSail<>(graph);
			sail.initialize();
			ValueFactory vf = sail.getValueFactory();
			SailConnection sc = sail.getConnection();
			sc.begin();

			sc.addStatement(vf.createURI("http://tinkerpop.com#1"), vf.createURI("http://tinkerpop.com#knows"),
					vf.createURI("http://tinkerpop.com#3"), vf.createURI("http://tinkerpop.com"));
			sc.addStatement(vf.createURI("http://tinkerpop.com#1"), vf.createURI("http://tinkerpop.com#name"),
					vf.createLiteral("marko"), vf.createURI("http://tinkerpop.com"));
			sc.addStatement(vf.createURI("http://tinkerpop.com#3"), vf.createURI("http://tinkerpop.com#name"),
					vf.createLiteral("josh"), vf.createURI("http://tinkerpop.com"));

			System.out.println("get statements: ?s ?p ?o ?g");
			CloseableIteration<? extends Statement, SailException> results = sc.getStatements(null, null, null, false);
			while (results.hasNext()) {
				System.out.println(results.next());
			}

			System.out.println("\nget statements: http://tinkerpop.com#3 ?p ?o ?g");
			results = sc.getStatements(vf.createURI("http://tinkerpop.com#3"), null, null, false);
			while (results.hasNext()) {
				System.out.println(results.next());
			}

			sc.rollback();
			sc.close();
			sail.shutDown();
			graph.shutdown();
		} catch (SailException e) {
			e.printStackTrace();
		} finally {
			graph.shutdown();
		}
```

Full exception:
```
[...]

PR #5551, opened 2016-01-04
Workaround for issue #5424

fabianmeyer:
Workaround for issue #5424

The TinkerPop implementation of OrientDB uses the java.net.URLEncoder/Decoder to encode/decode class names. The default encoding for URLs, as defined in RFC 2396, will not encode the '.' character, which is invalid in class/field names. Furthermore, the default encoding uses a %-encoding for non-alpha-numeric characters, resulting in class/field names, containing '%' characters, which is invalid in OrientDB. 

The workaround extends the OrientBaseGraph.encodeClassName method for %-encoding of '.' and '_' characters and afterwards replaces all '%' characters by '_'. To achieve a bijective mapping, the encoding replaces all '_' charakters by  '%' characters and applies the default URL decoding, which already considers the correct decoding of the encoded '.' and '_' characters.

Example:
"http://tinkerpop.com#knows" is encoded into the valid class name "http_3A_2F_2Ftinkerpop_5Ecom_23name", which decodes in "http://tinkerpop.com#knows"

### DIFF
--- a/graphdb/src/main/java/com/tinkerpop/blueprints/impls/orient/OrientBaseGraph.java
+++ b/graphdb/src/main/java/com/tinkerpop/blueprints/impls/orient/OrientBaseGraph.java
@@ -350,7 +350,10 @@ public abstract class OrientBaseGraph extends OrientConfigurableGraph implements
       iClassName = "-" + iClassName;
 
     try {
-      return URLEncoder.encode(iClassName, "UTF-8");
+        return URLEncoder.encode(iClassName, "UTF-8")    // default URL encoding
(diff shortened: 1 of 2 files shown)


lvca:
Hey @fabianmeyer thanks for this PR. Could you do this change in "develop" branch? We use "master" only for releasing. Look at: http://orientdb.com/docs/last/Contribute-to-OrientDB.html

tglman:
unlikely this break one test, i'm reverting it.

tglman:
sorry wrong issue

PR #5611, opened 2016-01-19
Using built-in constant-time comparison function to avoid timing attacks

obi1kenobi:
In agreement with Luca Garulli, opening a PR to fix a timing attack vulnerability. See here for details on how to protect against such attacks:
http://stackoverflow.com/questions/33626298/messagedigest-isequal-function-use-in-java/33626618

The removed line 109 performed a regular string comparison between the supplied and expected passwords, which was vulnerable to a timing attack. I am replacing that with a constant-time check of the hashes of the two passwords, which is not able to be timed. I am also replacing the custom constant-time comparison function with the Java built-in function `MessageDigest.isEqual` which does the same thing.

As this is a security fix, it would be nice to backport this to all active releases as soon as possible.

Thanks!

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/security/OSecurityManager.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/security/OSecurityManager.java
@@ -106,7 +106,10 @@ public class OSecurityManager {
       return checkPasswordWithSalt(iPassword, s, PBKDF2_SHA256_ALGORITHM);
     }
 
-    return iPassword.equals(iHash);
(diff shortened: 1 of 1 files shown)


lvca:
Thanks you! In v2.1.x the OSecurityManager is very different, I'm checking how we can apply it.

obi1kenobi:
If you'd like, I can open a separate PR against the 2.1.x branch with a suggested fix.

lvca:
It would be fantastic.

PR #5361, opened 2015-11-23
Initial implementation for #5360.  Enhances the CREATE Property syntax.

mmacfadden:
Unit tests seem to be passing for #5360

Issue #5360, opened 2015-11-22
Enable SQL CREATE PROPERTY to set ATTRIBUTES such as NOT NULL,  MANDATORY, and READONLY


mmacfadden:
Presently quite a few commands are required to create a property in a class that is "not null", "mandatory", or "read only".  For example:

``` SQL
CREATE Property User.username STRING
ALTER PROPERTY User.username MANDATORY true
ALTER PROPERTY User.username NOTNULL true
ALTER PROPERTY User.username READONLY true
```

It would be convenient to be able to do the following (or a mix of the attributes):

``` SQL
CREATE Property User.username STRING MANDATORY NOTNULL READONLY
```

These attributes would come at the end of the statement.  For example if this property was a linked element:

``` SQL
CREATE Property Employee.supervisor LINK Employee MANDATORY NOTNULL READONLY
```

This would make defining properties in SQL much more succinct.  Other properties could be defined as well (DEFAULT, MIN, MAX, etc.). It appears that this change could be made in the OCommandExecutorSQLCreateProperty parse and execute methods.

If this seems like a good feature, I would be willing to create a PR.

mmacfadden:
I have an initial stab at this implemented and unit tested.  It's targeted to the develop branch.  I was planning on submitting a PR.  To start, I thought it would be simpler to support just the first three.

mmacfadden:
PR #5361 is most likely read to go.  It is failing CI for an issue that is not related to the PR.  Many other PRs are failing for the same reason now.

lvca:
Thanks for the PR, @luigidellaquila will look at it soon.

mmacfadden:
Thanks.  If this is accepted, I will likely continue it to add a few of the other parameters like DEFAULT, MIN, MAX, REGEX as well.  But wanted to get this incrementally in first since it is the immediate item of interest.

smolinari:
:+1: @mmacfadden 

Scott

mmacfadden:
@lvca Given that the initial code was merged, would you like me to implement some of the other properties I mentioned above, or do you want to close this out for 2.2?

lvca:
@mmacfadden 2.2.0 is closed as features, but you can still work on develop branch so they will be merged for the next release

mmacfadden:
[...]

PR #5639, opened 2016-01-25
Revert "Initial implementation for #5360.  Enhances the CREATE Property syntax."

luigidellaquila:
Reverts orientechnologies/orientdb#5361

Sorry, I have to revert this PR, because it makes the grammar inconsistent. 
Consider the following:
```
CREATE PROPERTY Foo.bar LINK MANDATORY
```

Does the `MANDATORY` refer to the linked class (it is a valid class name) or to the Mandatory attribute?

We need to find a better solution for the syntax.

Thanks

Luigi


smolinari:
Make "Mandatory" a reserved word?

Scott

luigidellaquila:
Yes, it could be a good idea, but we already have a bunch of reserved words, and every time I add a new one someone will have problems with backward compatibility...

I'll see what I can do

Thanks

Luigi

smolinari:
Well, a data model with a class named "mandatory" sounds rare to me. And sometimes BC needs to take a back seat to good new functionality. :wink: 

Scott

smolinari:
Though, things that break BC shouldn't happen in minor updates either. Only major point releases or higher, like in the upcoming 2.2 release.

Scott

PR #5655, opened 2016-01-29
Added default sequence type

matanshukry:
The current syntax should allow one to create syntax without declaring it's type, yet if you try the following:
```
CREATE SEQUENCE idseq
```
an exception will be thrown (NPE specifically).

This fixes it by using a default sequence type (cached).
I used cached as the default sequence type since it's the one that may result in better performance.


lvca:
Thanks!

PR #5675, opened 2016-02-02
Default value info

matanshukry:
Fixed information regarding the default value, including:
* Import/Export
* Binary protocol (OServerCommandGetDatabase)
* Http protocol (OProperty)
* Console


lvca:
Thanks again @matanshukry!

Issue #5703, opened 2016-02-09
UPDATE REMOVE from nested EMBEDDED MAP not working


mmacfadden:
Given the following document (in the class 'test):

``` JSON
{
  "data": {
    "child": {
      "k1": "v1",
      "k2": "v2"
    }
  }
}
```

The FOLLOW SQL Statement will fail:

``` SQL
UPDATE test REMOVE data.child = "k1"
```

The parser will throw and exception complaining about a "." position 23.  Interestingly, the following query does work:

``` SQL
UPDATE test SET data.child.k3 = "v3"
```

This will properly add a new property to the child map.  So the UPDATE SET command handles the dot notation properties, but the UPDATE REMOVE does not.  Also, the following command WILL work:

``` SQL
UPDATE test REMOVE `data.child` = "k1"
```

Adding the back ticks causes the parser to treat the whole thing as a QUOTED_IDENTIFIER.  So this is a good workaround for now.  

I actually have already identified the issue in the grammar file / parser and will submit a pull request.

lvca:
Implemented in 3.0-M2.

PR #5704, opened 2016-02-09
Fix and test for #5703.

mmacfadden:
This PR allows dot notation path to appear in an UPDATE REMOVE command such as:

```SQL
UPDATE REMOVE data.child = "k1"
```

This should fix and test #5703.


lvca:
Thanks @mmacfadden! @luigidellaquila this is yours ;-)

mmacfadden:
@luigidellaquila, I completely forgot about this one.  I noticed it was targeted at develop, but i suspect it should be targeted at 2.2.x at this point.  Thoughts.

PR #5716, opened 2016-02-12
Auto increment

matanshukry:
Added an auto increment attribute.

The possible values for it are:
- a sequence name, used to generate the next value
- true, which will create a sequence and assign that for the auto increment.

Notes:
- The auto increment 'overrides' the default value behavior, and hence when both are present, the default value will be ignored.
- I also added a test case for auto generate

Also fixed:
cached sequence - Used to skip a number when allocating cache (for cache size 20 - numbers would be 17, 18, 19, 21, 22, etc..). Also added a test for this


smolinari:
Nice work. :+1: 

Can you give some examples showing how this would work within a query?

Scott

matanshukry:
Sure,

```
CREATE SEQUENCE mySeq TYPE ORDERED

CREATE CLASS User
CREATE PROPERTY User.id LONG
CREATE PROPERTY user.secondId SHORT
CREATE PROPERTY user.thirdId INTEGER
ALTER PROPERTY User.id AUTOGENERATE true
ALTER PROPERTY User.secondId AUTOGENERATE mySeq
ALTER PROPERTY User.thirdId AUTOGENERATE mySeq

INSERT INTO User SET name="Mat"
INSERT INTO User SET name="Joe"
INSERT INTO User SET name="Sam"
INSERT INTO User SET name="Niki"

SELECT FROM User
```

and the result:
```
----+-----+------+----+--------+-------+----
#   |@RID |@CLASS|name|secondId|thirdId|id  
----+-----+------+----+--------+-------+----
0   |#17:0|User  |Mat |1       |2      |1   
1   |#18:0|User  |Joe |3       |4      |2   
2   |#19:0|User  |Sam |5       |6      |3   
3   |#20:0|User  |Niki|7       |8      |4   
----+-----+------+----+--------+-------+----
```

As you can see, you can also share sequences this way.

smolinari:
That is pretty slick. 

I am no expert and I might be just picky and might also be off in left field too, but I believe the usual way to do auto increment fields in SQL is something like

```
CREATE TABLE User (
Id int NOT NULL AUTO_INCREMENT
);
```
It might be too much to ask for that all in one command with ODB, like 

```
CREATE CLASS User (
Id LONG AUTO_INCREMENT
);
```
Wouldn't that be cool? :smile: 

... but could this be possible?

`ALTER PROPERTY User.id AUTO_INCREMENT true`

That would be more inline with how the RDBMS world does it and lower the learning curve some more. 

And for using mySeq, we could have....

`ALTER PROPERTY User.id  AUTO_INCREMENT mySeq`

Scott

sinfex:
:+1: we really need this

matanshukry:
@smolinari - the attribute name (AUTO_INCREMENT / AUTOGENERATE) can be changed easily; It's up to the Orient team though, in my opinion.

Regarding fields with the create class command - I think it's a good idea, yet doesn't belong here. You should open a new issue for it though.

ckiyici:
:+1:

nagarajasr:
regarding the attribute name AUTOGENERATE or AUTO_INCREMENT, the latter makes it more explicit that the field is a numeric field whereas the former suggests that possibly the logic for generating the field value is more generic e.g., a function. (e.g., field should follow a specific naming convention user0001, user0002 etc.).

@matanshukry what is the behaviour in the following scenario (extending your example):

```

INSERT INTO User SET name="Mat"
INSERT INTO User SET name="Joe"
INSERT INTO User SET name="Sam"
INSERT INTO User SET name="Niki"
UPDATE User SET id=5 where name="Niki"
INSERT INTO User SET name="Doraemon" 

```

what is the behaviour of the last insert statement ?

is there a possibility to prevent such out of sequence updates to the auto_generated field?

smolinari:
@nagarajasr - actually I agree after thinking about it some more. 

If the sequence value (or function) is more than just an incremented integer, then it should be "AUTO_GENERATE". If though, it is a standard increment of an integer (and hopefully a standard feature), I'd like it to be "AUTO_INCREMENT". Notice the snake case too. The function needs to be broken into two words.

So, to my examples. We would have:

`ALTER PROPERTY User.id AUTO_INCREMENT true`

and

`ALTER PROPERTY User.id AUTO_GENERATE mySeq`

Scott

matanshukry:
@nagarajasr - Nope. To be honest though, I don't think it should. Auto increment/generate fields will generate their fields in a certain manner. If you interfere with it, it's up to you to make it work.

Regarding the name, I really think it's up to the Orient team.
Either way, adding both verbs (increment and generate) as two different attributes, in my opinion, is too confusing and should not be done.

smolinari:
@matanshukry - actually, I see your point now after reading the docs on sequences. For some reason, I was under the impression that a user could create her own form of sequences. @nagarajasr 's point about the user defined function also supported my impressions. However, I can't find any user definable sequences. So, if there are none. I agree. It should be AUTO_INCREMENT only. 

Scott

smolinari:
Also, looking at the other ODB commands and the possible attributes in the docs, the attribute would more than likely need to be 

`AUTOINCREMENT` 

as ODB [put the words for attributes together](http://orientdb.com/docs/last/SQL-Alter-Class.html#supported-attributes). 

Scott

nagarajasr:
@matanshukry I merged your commit in my local branch (branched off develop) and observed that the commands "alter property... AUTOGENERATE" fail to work from the console. there are no other local code changes.
i noticed from the github commit files list that there have not been any changes made to the SQL grammar (core/src/main/grammar/OrientSQL.jjt). usually when a new command is being introduced, the grammar must be updated with the new syntax so that the command can be parsed correctly.
unless I am missing something, you must also update the grammar file.
can you verify if the console commands work in your environment, and update the grammar file for the new syntax?

nagarajasr:
@matanshukry i also noticed that the sequence for some reason is getting reset to "1" after a server restart. not sure if this is an issue with your commit or issue with the sequence implementation itself.
this is definitely a bug IMHO.

PR #5717, opened 2016-02-12
[+Issue] Default value now happens on validation (require saving)

matanshukry:
## Problem

Currently, the default value is set when creating the field.
As a result, when deserializing a class, it will first set the default fields, and then try to deserialize the fields. When it encounters a field, it will ignore it since a value is already set (rough workflow).

```
1. CREATE CLASS User
2. CREATE PROPERTY User.name
3. ALTER PROPERTY User.name defaultvalue Joe

4. insert into User set name="Not Joe"
5. select from user
```

As a result, when the above is executed from the console, it will show one record in user, with name set to "Joe" rather than "Not Joe".

I Also added a test case using serializer.
## FIx

The fix I did was to move the default value from when the class initialize, to when the class saves (validation). This fixes the issue, however requires the document will be saved before one can see the default value. Since this is a new feature, I think we can still change it before it will start confusing some people. I also think it's the correct way the feature should work.


lvca:
@tglman WDYT?

PR #5724, opened 2016-02-15
final version of position allocate

tglman:


core/src/main/java/com/orientechnologies/orient/core/tx/OTxListener.java
```diff
@@ -26,8 +26,9 @@
  * Listener interface to catch events during transactions phases.
  * 
  * @author Luca Garulli
- * 
+ * deprecated since 2.2
  */
+@Deprecated
```
lvca: Why have you deprecated it?

Issue #4831, opened 2015-08-21
Create single entry point for storage lifecycle management.


laa:
Storage consists from 2 main components:

1. Data clusters.
2. Data structures for indexes.

Those components are managed by different managers:

1. OStorage 
2. OIndexManager

Which causes issues during:

1. Transaction management
2. Storage backup
3. Incremental backup
4. Storage close

Those issues may potentially lead to data corruption so we need to introduce single access point for both those component types.

laa:
Fixed.

PR #5730, opened 2016-02-15
Used the same OModificationLock object of storage

lvca:
Used the same OModificationLock object of storage

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentTx.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentTx.java
@@ -20,6 +20,13 @@
 
 package com.orientechnologies.orient.core.db.document;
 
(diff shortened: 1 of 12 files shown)


laa:
Luca we can not remove modification lock from index in 2.1
because indexes before 2.2 directly go to data structures and do not participate in storage modification lock it is fixed in https://github.com/orientechnologies/orientdb/issues/4831

lvca:
I didn't remove modification lock in indexes, but I used the same object as for storage.

laa:
Well probably it will work, let me think till tomorrow.

laa:
@lvca could you trace all usages of prohibit modifications from indexes and remove them, so we will leave storage as central point of control, then I think we may merge pull request ?

lvca:
So you mean to completely remove the usage of modification lock, right?

lvca:
In all the indexes?

PR #5731, opened 2016-02-15
Modificationlock

lvca:
Removed all the modification locks in indexes


lvca:
I completely removed all the modification locks in all the indexes. @laa Please review it.

laa:
Well Luca, you should need to remove modification lock on clusters it is cluster level only, needed to import only single cluster to database at runtime.

laa:
I will quickly make the change

PR #5836, opened 2016-03-16
Deprecating untyped indexes

taburet:
Untyped indexes shown to be unstable, deprecating.


laa:
Cool, congratulation, your first bug fix !

Issue #5693, opened 2016-02-05
Transaction propagation is not working 


linus46:
```
function1(){
    OrientGraph db = factory.openDatabase() ; // which will give active graph on current thread
    db.begin();
    function2();
    // save some data 
    db.commit();
}
function2(){
    OrientGraph db = factory.openDatabase() ; // which will give active graph on current thread
    db.begin();
    // save some data 
    db.commit(); // this commit is saving the data into db 
}
```
here function2 commit is saving the data even though its part of function1 . 

sample code : 
https://gist.github.com/anonymous/54cd70a1e53d0381658a

machak:
"factory.openDatabase() // which will give active graph on current thread"
Did you check this?
http://orientdb.com/docs/2.1/Transaction-propagation.html states:

The same instance of database should be used between methods.

linus46:
yes , the db instances are same. inside openDatabase function i am checking for graph instance on thread 

        if(OrientGraph.getActiveGraph()==null || OrientGraph.getActiveGraph().isClosed())
        {
        	db = graphFactory.getTx();	
        }
        else {
        	db = (OrientGraph) OrientGraph.getActiveGraph();
        }

this will give same instance of database

laa:
Fixed

PR #5842, opened 2016-03-17
Transaction propagation is not working #5693

taburet:
Auto-started transaction and user-initiated transactions are now separated.


laa:
Well done :-)

PR #5857, opened 2016-03-18
Fix orientdb-tools OSGi exports

mcculls:
Add 'com.orientechnologies.orient.server.config' to the packages exported from orientdb-tools.
(this package was moved from orientdb-server to orientdb-tools but the exports weren't updated)


lvca:
Thanks!

Issue #5865, opened 2016-03-21
Allow to start manual transaction only if auto transactions are disabled in graph database


laa:
If we run OrientGraph#begin method in many cases auto transaction is already started. As result during call of OrientGraph#commit method transaction is not committed and changes are not stored.

PR #5880, opened 2016-03-24
Allow to start manual transaction only if auto transactions are disab…

taburet:
…led in graph database #5865

### DIFF
--- a/graphdb/src/main/java/com/tinkerpop/blueprints/impls/orient/OrientBaseGraph.java
+++ b/graphdb/src/main/java/com/tinkerpop/blueprints/impls/orient/OrientBaseGraph.java
@@ -1738,7 +1738,7 @@ public abstract class OrientBaseGraph extends OrientConfigurableGraph implements
             "Requested command '%s' must be executed outside active transaction: the transaction will be committed and reopen right after it. To avoid this behavior execute it outside a transaction",
             msg.toString());
       }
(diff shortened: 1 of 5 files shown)


graphdb/src/main/java/com/tinkerpop/blueprints/impls/orient/OrientBaseGraph.java
```diff
@@ -1738,7 +1738,7 @@ public long countEdges(final String iClassName) {
             "Requested command '%s' must be executed outside active transaction: the transaction will be committed and reopen right after it. To avoid this behavior execute it outside a transaction",
             msg.toString());
       }
-      raw.commit();
+      raw.commit(true);
```
taburet: Not sure about this, before nested tx support this was relatively safe operation, now we need to commit all nested transactions too, but we can't restart them.

laa: Why we can not ? We can , merely call begin the same amount of times as amount of nested txs was, was not it ?

taburet: Good point, will fix it.

laa:
Will wait for fixed one.

PR #5899, opened 2016-03-29
Two phase commits

taburet:


lvca:
Thanks for the PR. In order to support this on distributed part, what should I do? I cannot find the `confirmCommit()` method to confirm the transaction after commit.

taburet:
There are two methods `ODatabase#initiateCommit()` and `ODatabase#initiateCommit(boolean)`, which return `OUncompletedCommit`, which has `complete` and `rollback` methods.

PR #5906, opened 2016-03-30
Fixes for speeding up when registering a lot of classes

wouterv:
Throughout the registration code, the schema is reloaded way to many times.
- Added extra method to avoid this. 
- Removed one unnecessary reload
- On OEntityManager initialization, only reload the schema when the last class has been registered

Also in OClassImpl the polymorphic cluster ids were kept in an array, which needed to be sorted over and over again, which is very expensive. Added a helper SortedSet next to it so the sorting is now fast.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/entity/OEntityManager.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/entity/OEntityManager.java
@@ -23,6 +23,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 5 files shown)


wouterv:
Ran mvn clean pakage locally and I had no build issues.

tglman:
hi,

This look really good, the only thing is not the best is that this commit: https://github.com/orientechnologies/orientdb/pull/5906/commits/e07cf58fc0d3e06d8864a942df85aac978644236

introduce a new structure in memory, for the case of schema we prefer to have a smaller memory footprint and use more cpu, would be better a full replacement of the array with a set.

all the rest can be merged semless.

wouterv:
@tglman I was struggling with this. When the array is replaced by a set, public int[] getPolymorphicClusterIds() would also have to be changed, and so all the calling methods, which would open up a big load of changes all over the place, so that is why I opted for this solution.

If we would do change the whole thing, the only memory improvement you get is not having the cluster id array, which is quite small in terms of actual memory consumption, so is it really such a big deal?

tglman:
Ok i understand, can you skip that commit for now we will see how to refactor our api accordly

wouterv:
@tglman new pull request: #5936

PR #5924, opened 2016-04-04
Refactor remote client

tglman:
This is a refactor for clean up a bit the client side storage an fix some issue with session and authentication management.

The changes involve the move of the remote session information from the OStorageRemoteThread to ODatabaseDocumentTx: https://github.com/orientechnologies/orientdb/blob/refactor_remote_client/core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentTx.java#L136

with the remove of all the OStorageRemoteThread structures and of the special behavior of ORemoteEngine.

Was refactored also the ODatabaseCompare to work with database instead of direactly with storage, this was done because the storage don't have anymore the session information.

All the session specific inforation like sessionId tokens and user/pwd are moved into OStorageRemoteSession, this  fix a wide range of issue with use of multiple users from remote.


client/src/main/java/com/orientechnologies/orient/client/remote/OServerAdmin.java
```diff
@@ -101,9 +100,8 @@ public Void execute(OChannelBinaryAsynchClient network) throws IOException {
           String password = iUserPassword;
 
           OCredentialInterceptor ci = OSecurityManager.instance().newCredentialInterceptor();
```
robfrank: Since uname and passwd aren't used, I'll prefer to extract a method "setupCredentialIntercerptor"

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentTx.java
```diff
@@ -133,6 +133,7 @@
   private boolean initialized = false;
   private OTransaction currentTx;
   private boolean keepStorageOpen = false;
+  protected Object sessionMetadata;
```
luigidellaquila: Why Object...? Can't we define an interface on top of OStorageRemoteSession?

PR #5926, opened 2016-04-04
added flag to disable transaction check for move vertex operations

wouterv:


lvca:
@wouterv thanks for this PR. I'd prefer to remove this check at all and leave the responsibility to the user: using OrientGraphNoTx is faster (on atomic operations), but less safe. WDYT about updating this PR by removing the new global setting and just removing the check?

wouterv:
@lvca agreed and done.

lvca:
Perfect, thanks!

luigidellaquila:
@lvca why was the NON_TX_CHECK_MOVE_VERTEX flag removed from the PR????
This is definitely unsafe!!!

Luigi

lvca:
The reason is already here, read my comment above: https://github.com/orientechnologies/orientdb/pull/5926#issuecomment-209845849.

Issue #5898, opened 2016-03-29
OObjectEntitySerializer.registerClass is slow because of unneeded schema reloads


wouterv:
When you have a lot of classes that need to be registered with the object api, starting up can be quite slow. This is because registerClass always calls schema.reload() and most of the time this isnt needed.

Working on a pull request for this.

wouterv:
PR: #5906

lvca:
PR merged, I can close this issue.

PR #5936, opened 2016-04-05
Speedup registering classes2

wouterv:
Same as #5906 but with commit removed as requested by @tglman


tglman:
look good merging

tglman:
ported to develop as well

wouterv:
@tglman you should probably update issue #5898 as well

smolinari:
Thanks for your support @wouterv !

Scott

Issue #5882, opened 2016-03-24
New API to get the RAM used by OrientDB


lvca:
http://andreylomakin.tumblr.com/post/137154835613/how-to-calculate-maximum-amount-of-memory-consumed

laa:
Fixd

laa:
MBean with name com.orientechnologies.common.directmemory:type=OByteBufferPoolMXBean was added.

PR #5943, opened 2016-04-06
New API to get the RAM used by OrientDB #5882

taburet:


lvca:
Thanks @taburet, how can we use this API? For example in console, studio or even server dumps?

taburet:
@lvca, to see the byte buffer pool stats you have to launch `jvisualvm` with MBeans plug-in installed, connect to the target JVM instance and navigate to `com.orientechnologies.common.directmemory/OByteBufferPoolMXBean` bean. Fore sure, you may use any JMX client you like, from the console or from the code using the Java JMX library.

taburet:
@lvca, if you are thinking about the obtaining of the stats from the code inside of the OrientDB instance,  it's easy – `OByteBufferPool` implements `OByteBufferPoolMXBean` interface, so all of the stats maybe fetched from `OByteBufferPool ` instance directly, the global one may be found at `com.orientechnologies.common.directmemory.OByteBufferPool#instance`.

I will provide a detailed description along with my new the pull requests next time ;)

PR #5971, opened 2016-04-12
Refactor remote server

tglman:
Server side refactor for the binary protocol, for remove old not needed code, and make the client connection not associated to the socket thread but provided in the execution context.


maggiolo00:
+1

PR #6002, opened 2016-04-19
A Star function now added to graph functions

saeedtabrizi:
Now A Star function with some test is ready to use .

### DIFF
--- a/graphdb/src/main/java/com/orientechnologies/orient/graph/sql/functions/HeuristicFormula.java
+++ b/graphdb/src/main/java/com/orientechnologies/orient/graph/sql/functions/HeuristicFormula.java
@@ -32,7 +32,7 @@ import java.util.*;
 
 
 /**
- *  heuristic formula enum.
+ *  Heuristic formula enum .
  *
  * @author Saeed Tabrizi (saeed a_t  nowcando.com)
  *
--- a/graphdb/src/main/java/com/orientechnologies/orient/graph/sql/functions/HeuristicFormula.java
+++ b/graphdb/src/main/java/com/orientechnologies/orient/graph/sql/functions/HeuristicFormula.java
@@ -1,0 +1,47 @@
+/*
+  *
+  *  *  Copyright 2014 Orient Technologies LTD (info(at)orientechnologies.com)
+  *  *
+  *  *  Licensed under the Apache License, Version 2.0 (the "License");
+  *  *  you may not use this file except in compliance with the License.
+  *  *  You may obtain a copy of the License at
+  *  *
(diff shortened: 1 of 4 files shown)


lvca:
@saeedtabrizi thanks for this amazing contribution. @luigidellaquila could you review this tomorrow, please?

luigidellaquila:
Doing it now

Thanks

Luigi

luigidellaquila:
Hi @saeedtabrizi 

Thank you very much, the PR seems to be perfect!

Just two small things:

1) I see in the algorithm you do 

```
if (current.getIdentity().equals(goal.getIdentity()) || currentDepth >= paramMaxDepth) {
 
         while (current != null) {
              route.add(0, current);
              current = cameFrom.get(current);
         }
         return getPath();
}
```
so you return a partial path if it's longer than the maximum depth. Is it by literature or is it only a choice?
I'd expect that if I do not find a path that meets the requirements, the function returns an empty path... but maybe it's just my lack of expertise on this...

2) could you please send the PR on "develop" branch? I cannot merge it to master because we use master branch only for stable releases. I'll merge it as soon as it's available

Thanks!

Luigi

saeedtabrizi:
Hi @luigidellaquila  

1) In this case i have two choice that you attended . the first choice returns the **empty path** result or  the **partial **of routed path . I select the **second choice** because the **maxDepth **limitation helps to saving the resources in traversing the graph and in my philosophy that means the best routing until reach to depth limitation without lost the history of routing. 
In the fact of  many scenario in the real world apps , we want to know the routing path until to reach our limitation . for example **tracert **command in windows that shows to us the routing data and can be canceled in any iteration.

2) Of course :)

Many Thanks 
Saeed

luigidellaquila:
Hi @saeedtabrizi 

Good point, that use case makes a lot of sense. Maybe it's worth leaving it as a default and adding another parameter to change this behavior and return an empty path...? This could be an enhancement for another PR of course (or I can do it myself after the merge)

Thanks

Luigi

saeedtabrizi:
Hi @luigidellaquila 

That's good idea ,  are you agree we put new parameter (**emptyIfMaxDepth** = false) in A star function syntax ? 
If we set the **true** value for this parameter then returns the empty result if we reached max Depth limitation . by default this parameter must have false value absolutely .

Thanks 
Saeed

luigidellaquila:
Yes, I agree, I just would call it *emptyIfMaxDepthReached*

Thanks

Luigi

saeedtabrizi:
Just moved on #6019 PR

PR #6019, opened 2016-04-21
A star path finding function and test added to graph sql functions

saeedtabrizi:
A star path finding function and some test case added to graph sql functions ,
The latest syntax is :
```
astar(<sourceVertex>, <destinationVertex>, <weightEdgeFieldName>, [<options>]) 
 // options  : {direction:"OUT",edgeTypeNames:[] , vertexAxisNames:[] , parallel : false , emptyIfMaxDepth:false,heuristicFormula:'manhatan',tieBreaker:true,maxDepth:99999,dFactor:1.0,customHeuristicFormula:'custom_Function_Name_here'  }
```
**direction** : "out|in|both" , default is out .
**edgeTypeNames** : an array of string to filter edge type name in path finding , default empty array to get all edges .
**vertexAxisNames** : an array of string that contains the  axis fields that computes in heuristic cost . for example ['x','y'] ,default is empty array
**tieBreaker** : true|false , this option may be useful for breaking the ties for improve the path finding result , default is true.
**parallel** : true | false , if set the true this function working on parallel mode to improve the time of process , default is false (not implemented yet) .
**maxDepth** : long value to limit the depth of traversing graph in path finding . default is 99999
**dFactor** : a double , factor that mostly times to heuristic cost value to gain the improve result.
**customHeuristicFormula** : a function name to return a double value for computing the heuristic cost value .
**heuristicFormula** : a heuristic formula name that be , 'manhatan'|'maxaxis'|'diagonal'|'euclid'|'euclidnosqr','custom' , default is manhatan
**emptyIfMaxDepth** : true|false , if set true and reach to maxDepth limit route result sets to empty else the result contains routing to reach limitation .  default is false .

welcome to the idea and comments . :)

Thanks
Saeed

### DIFF
--- a/graphdb/src/main/java/com/orientechnologies/orient/graph/sql/functions/HeuristicFormula.java
+++ b/graphdb/src/main/java/com/orientechnologies/orient/graph/sql/functions/HeuristicFormula.java
@@ -1,0 +1,46 @@
+/*
+  *
+  *  *  Copyright 2014 Orient Technologies LTD (info(at)orientechnologies.com)
+  *  *
+  *  *  Licensed under the Apache License, Version 2.0 (the "License");
+  *  *  you may not use this file except in compliance with the License.
+  *  *  You may obtain a copy of the License at
+  *  *
+  *  *       http://www.apache.org/licenses/LICENSE-2.0
+  *  *
+  *  *  Unless required by applicable law or agreed to in writing, software
+  *  *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


luigidellaquila:
Merged, thanks!

Luigi

smolinari:
Great work! @luigidellaquila - how about the documentation of this?

Scott

luigidellaquila:
Hi @smolinari 

I'm opening a specific issue about this page of docs, I'll manage to do it in short

Thanks

Luigi

smolinari:
I was thinking maybe I could help. Sorry I wasn't clearer. Unless the information above isn't enough and you'd want to add more. 

Scott

PR #6028, opened 2016-04-22
refactor for improve handling of multiple server sessions

tglman:
Make able the client to keep all session open with all servers, and close them properly at the close.

It avoid as well no needed reopen in case of switching between servers


lvca:
I like it, merged.

Issue #3823, opened 2015-03-27
detachAll(*, true) for record with embedded fields works incorrectly


gypergrand:
Seems that 'detachAll' command improperly collecting identities for embedded field records in the map 'alreadyDetached' in OObjectEntitySerializer, which are always '#-1:-1', and therefore creating wrong nonProxiedInstance object.

Here is the simple JUnit test:

import static org.hamcrest.CoreMatchers.equalTo;
import static org.hamcrest.CoreMatchers.is;
import static org.hamcrest.CoreMatchers.not;
import static org.junit.Assert.assertThat;
import javax.persistence.Embedded;
import javax.persistence.Entity;
import org.junit.Test;
import com.orientechnologies.orient.object.db.OObjectDatabaseTx;

public class DetachAllTest {
    @Entity
    public static class EmbeddedClass {
        public EmbeddedClass() {
            super();
        }
    	private String name;

		public String getName() {
			return name;
		}

		public void setName(String name) {
			this.name = name;
		}
    }
    
    @Entity
    public static class ContainingClass {
    	@Embedded
    	private EmbeddedClass field1;
    	@Embedded
    	private EmbeddedClass field2;
		public ContainingClass() {
			super();
		}
		public EmbeddedClass getField1() {
			return field1;
		}
		public void setField1(EmbeddedClass field1) {
			this.field1 = field1;
		}
		public EmbeddedClass getField2() {
			return field2;
		}
		public void setField2(EmbeddedClass field2) {
			this.field2 = field2;
		}
    }
    
    @Test
    public void detachAllForRecordsWithEmbeddedFieldsBugTest() {
    	OObjectDatabaseTx db = new OObjectDatabaseTx("plocal:./LocalTestDB");
    	if (!db.exists()) {
    		db.create();
    	} else {
    		db.open("admin", "admin");
    	}
    	try {
			db.setAutomaticSchemaGeneration(true);
			db.getEntityManager().registerEntityClass(EmbeddedClass.class);
			db.getEntityManager().registerEntityClass(ContainingClass.class);
			db.reload();
			EmbeddedClass emb1 = new EmbeddedClass();
			emb1.setName("emb1");
			EmbeddedClass emb2 = new EmbeddedClass();
			emb2.setName("emb2");
			ContainingClass con = new ContainingClass();
			con.setField1(emb1);
			con.setField2(emb2);
			con = db.save(con);
			ContainingClass concopy = db.detachAll(con, true);
[...]

PR #6072, opened 2016-04-29
OObjectProxyMethodHandler not properly handling multiple @Embedded fields on detachAll()

nathanmbrown:
`OObjectProxyMethodHandler.detachAll()` was incorrectly caching detached embedded objects against their empty record id, meaning the first embedded object was subsequently assigned to all further embedded fields, causing corruption/exceptions.
Add unit test to test multiple embedded fields.

Can this please also be backported to the 2.1.* versions as that's what we are using at present, thanks.


nathanmbrown:
This fixes this issue : https://github.com/orientechnologies/orientdb/issues/3823

tglman:
This pool request look good, i tested locally and do not create issue, merging it.

PR #6107, opened 2016-05-09
System database now is reachable from OServer, not only security

lvca:
System database now is reachable from OServer, not only from security


PhantomYdn:
@lvca , could you please briefly describe what's the goal to have system db? What's implication?

Issue #5113, opened 2015-10-13
Remove OSimpleVersion object


lvca:
We don't need it anymore, the version could be stored in an integer saving objects. We could still leave the class with static methods to separate responsibility.

lvca:
Bye bye OSimpleVersion...

PR #6122, opened 2016-05-11
Set correct status when unmarshalling embedded/link collections to avoid marking records as dirty

mcculls:
Backported unmarshalling improvements made to ORecordSerializerBinaryV0 under issue #5113 to 2.1.x

Diff is simpler with ignore whitespace: https://github.com/orientechnologies/orientdb/pull/6122/files?w=1 :)


lvca:
Merged, thanks!

PR #6166, opened 2016-05-19
Refactor client retry

tglman:
Refactor for remove exception management based on handleExcaption in favor of specific catch.
Refactor for centralize the remove of url for not available servers
Minor fix on pool management


lvca:
Perfect!

Issue #6092, opened 2016-05-06
OClassTrigger doesn't take into account multiple superclasses


PhantomYdn:
OrientDB supports multi super-classes.
But code of OClassTrigger was not changed to support that. Snippet:

``` java
while (fieldName == null || fieldName.length() == 0) {
        if (superClz == null || superClz.getName().equals(CLASSNAME))
          break;
        fieldName = superClz.getCustom(attr);
        superClz = superClz.getSuperClass();
      }
```

Btw, it will be helpfull to have some utility class for caching results of culculations till next schema update. 
For example OClassTrigger is trying to find proper field on every call instead of caching previously obtained results obtained from schema in a cache.

tglman:
commented on the pull request.

PR #6207, opened 2016-05-25
OClassTrigger should become aware about multiple superclasses

PhantomYdn:
This is fix for issue #6092 
Also it introduce new feature to get custom value from OClass taking into account hierarchy
Please consider to merge to 2.2.1


tglman:
hi @PhantomYdn,
the pool request seems ok, two points though:
I don't like to add new public API at this point of the release process, so i think is better add this API just to OImmutableClass and open another issue to evaluate how to make public consistent API for this specific case.  
I would like a test case ;)

PhantomYdn:
I can add test case - no problem:)
@tglman , @lvca , but can we please agreed about adding  String getCustomHierarchical(...) methods to API? It's just adding: no changes for existing methods. As I wrote it will help a couple of related issues which rely to "custom" fields.

PhantomYdn:
Guys, any thoughts? Is it possible to add those methods to OClass?

PhantomYdn:
According to Code 2.2 multiple hierarchy is being now supported. Closing as outdated.

lvca:
Thanks.

PR #6248, opened 2016-06-02
Fix status command not working on small terminals

avindra:
The 'w' flag for 'ps' needs to be passed twice in order to enable
unlimited wide output.

Per man page for ps:

```
w               Wide output. Use this option twice for unlimited width.

```

Tested on SuSE Linux Enterprise Server 11.

### DIFF
--- a/server/script/orientdb.sh
+++ b/server/script/orientdb.sh
@@ -41,7 +41,7 @@ stop() {
 }
 
 status() {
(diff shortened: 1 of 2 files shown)


avindra:
Closing this since I really don't care for OrientDB (although you SHOULD merge this on behalf of all your customers) :smile:

Issue #6246, opened 2016-06-02
Refactoring of parameters inside server.[sh|bat] and consolidation with orientdb-server-config.xml


madama:
There  are some improvements that can be done:
- MAXDISKCACHE should be overridable from env
- invoking "sever.sh debug" could enable debug options without cut&paste
- "-Dprofiler.enabled=true" is a duplicate of what is in xml config file, and maybe it should be false by default

PR #6251, opened 2016-06-03
#6246 Refactoring of server and config files

madama:
server.bat still need some attentions

### DIFF
--- a/server/script/server.sh
+++ b/server/script/server.sh
@@ -81,9 +81,14 @@ fi
 
 # DEBUG OPTS, SIMPLY USE 'server.sh debug'
 DEBUG_OPTS=""
-if [ -n $1 ] && [ $1 = "debug" ]; then
-    DEBUG_OPTS="-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=1044"
-fi
+ARGS='';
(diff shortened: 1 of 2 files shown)


server/script/server.sh
```diff
@@ -80,24 +79,28 @@ if [ -f "$ORIENTDB_PID" ]; then
...
-
 if [ -z "$ORIENTDB_OPTS_MEMORY" ] ; then
     ORIENTDB_OPTS_MEMORY="-Xms512m -Xmx512m"
 fi
-# ORIENTDB MAXIMUM DISKCACHE IN MB, EXAMPLE, ENTER -Dstorage.diskCache.bufferSize=8192 FOR 8GB
-MAXDISKCACHE=""
+
+# ORIENTDB SETTINGS LIKE DISKCACHE, ETC
+if [ -z "$ORIENTDB_SETTINGS" ]; then
+    ORIENTDB_SETTINGS=""
```
avindra: If `ORIENTDB_SETTINGS` is blank, set it to blank?

Am I missing something here or is this just redundant?

madama: Yep, is not necessary, but we already have a place were to set future default settings

PR #6287, opened 2016-06-11
Correct the date format in HTTP Response Headers

mmacfadden:
The current date header in the http responses looks like this: Sat Jun 11 13:45:12 GMT 2016 This is because the default serialization of Java Dates via the Date.toString() function is "EEE MMM dd HH:mm:ss zzz yyyy"

However, RFC 7231 defines the correct date format to be something like Sun, 06 Nov 1994 08:49:37 GMT, which corresponds to the java date format of "EEE, dd MMM yyyy HH:mm:ss zzz". HTTP Clients are not required to correctly parse non-standard date formats, and we are currently having an issue with this.

This patch brings the date header into compliance with HTTP 1.1, per RFC 7231.

### DIFF
--- a/server/src/main/java/com/orientechnologies/orient/server/network/protocol/http/OHttpResponse.java
+++ b/server/src/main/java/com/orientechnologies/orient/server/network/protocol/http/OHttpResponse.java
@@ -31,6 +31,7 @@ import com.orientechnologies.orient.server.OClientConnection;
 import java.io.*;
 import java.net.Socket;
 import java.nio.charset.Charset;
+import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.zip.GZIPOutputStream;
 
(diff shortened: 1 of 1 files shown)


mmacfadden:
Wondering if this can be merged in.  It seems like it is pretty simple.

tglman:
hi @mmacfadden,

Just reviewed the code, SimpleDateFormat is not thread safe so it cannot be static, i'm sorry but i've to reject this.

is welcome a new one with a not shared SimpleDataFormat

Thanks anyway for all the effort.

mmacfadden:
@tglman Does OrientDB still need to support Java 1.7?  If not I can simply use DateTimeFormatter instead, which is thread safe.  If we have to support Java less than 1.8 then I can add a fix for the static simple data format

mmacfadden:
@tglman I updated the code to address your concerns, I am not sure if you can reopen this, or if I need to create a new merge request.

Issue #5656, opened 2016-01-30
SQL create/alter property attributes


matanshukry:
Currently the only way to edit any property attributes is by using the `ALTER PROPERTY` command. That is, one can not specify attributes with the `CREATE PROPERTY` command.

Also, the `ALTER PROPERTY` statement seems to support only a single attribute. Would be nice if it could support more than a single attribute at a time.

smolinari:
+1

Scott

mpmlj:
+1

mmacfadden:
@matanshukry Check out #5360.  It enables some of what you are asking for.  Supporting NOT NULL, MANDATORY, and READONLY.  I have an almost completed pull request for MAX, MIN, and DEFAULT as well.

smolinari:
Why to go @mmacfadden :+1: 

Scott

PR #6292, opened 2016-06-12
Allow setting attributes in CREAT PROPERTY 

mmacfadden:
This is another attempts at issue #5360.  It allows users to set attributes of a property when creating it.  This also addresses #5656.

The implemented syntax looks like this:

```CREATE PROPERTY Foo.bar INTEGER (MANDATORY true, READONLY false, DEFAULT 5)```

The currently supported properties are:
- MANDATORY
- NOTNULL
- READONLY
- MIN
- MAX
- DEFAULT

These were the ones specifically asked for.  Adding ```REGEX``` and ```COLLATE``` should be relatively simple, and can be done before this is merged, if desired.

The following attributes are purpousely excluded because they are already defined elsewhere in the Create syntax and would cause confusion:
- NAME
- TYPE
- LINKEDCLASS
- LINKEDTYPE

The ```CUSTOM``` attribute syntax is not yet supported during CREATE.  This can be added, but I would suggest doing that in a subsequent pull request if there is interest.

NOTE:  A large number of test were added since the overall grammar and parsing for the CREATE PROPERTY syntax was changed.

### CODE REVIEW

**Comment by smolinari** [comment]
Does an attribute always need a value to work or are there defaults? For instance, if I only add `MANDATORY` with no value, will `true` be the given default value?

Scott

**Comment by mmacfadden** [comment]
@smolinari Yes.  For options that a "boolean based", the presence of the keyword can be viewed as turning the attribute on.  This is a valid statement:

```CREATE PROPERTY Foo.bar INTEGER (MANDATORY, READONLY, NOTNULL)```

This property will be mandatory, readonly, and non null.  The preceding is synonymous with:

```CREATE PROPERTY Foo.bar INTEGER (MANDATORY true, READONLY true, NOTNULL true)```

The main reason for supporting both is syntactic consistency with the other attributes (e.g. <ATTRIBUTE_NAME> <ATTRIBUTE_VALUE>). It is also consistent with how the ALTER ATTRIBUTE command works.  The version with MANDATORY, READONLY, and NOTNULL is mainly syntactic sugar / shorthand.

**Comment by smolinari** [comment]
And `UNSAFE` will be at the end, outside of the parentheses? 

`CREATE PROPERTY Foo.bar INTEGER (MANDATORY, READONLY, NOTNULL) UNSAFE`

Also, you mention syntactic consistency. However, the current `ALTER PROPERTY` command has neither commas nor parentheses in use, so I'd like to suggest they also be added to `ALTER PROPERTY` too. 

`ALTER PROPERTY Account.age (MANDATORY TRUE, READONLY FALSE)`

or make the `CREATE PROPERTY` more like `ALTER PROPERTY`and have no commas or parentheses.

`CREATE PROPERTY Foo.bar INTEGER MANDATORY true READONLY false NOTNULL true UNSAFE true`

(I like the lower letters for the values, so the command isn't just a blob of caps. :smile: )

This would be more logical, as such a change doesn't require (for consistency sake) a change to `ALTER PROPERTY`.

It matters less to me that we have commas or parenthesis, although I prefer not having them more, however consistency is majorly important (which I believe you believe in too) and this solution, unfortunately, brings a good bit of inconsistency.

Oh, And once this gets down pat, the next step is to be able to do this..... :smile: 

https://github.com/orientechnologies/orientdb/issues/5751

I find it really great of you to dig into the code and help with this solution. I wish I had Java experience, I'd help out too. I hope I can help a little (and am not annoying) with my own suggestions and arguments. 

Scott

**Comment by mmacfadden** [comment]
The consistency I was meaning was the following:

```ALTER PROPERTY Foo.bar MANDATORY true;```

In this case the general form is <attribute_name> <attribute_value_expresion>.  Meaning you can't do ```ALTER PROPERTY Foo.bar MANDATORY``` when altering a property.  The reason is, when you are creating a property, READONLY, MANDATORY, and NOTNULL have a default value of false.  So one can assume that if you are mentioning it during create, its because you want to set it to true.  So you don't really EVER need to put true in the CREATE case.  I could have just left it so that ```CREATE PROPERTY Foo.bar (MANDATORY true) was invalid, but I felt since the ALTER syntax allows MANDATORY true, I would allow that.

In terms of ALTER PROPERTY, there are no Parens, or Comma because (as far as I know) ALTER does not allow you to specify multiple attributes of a properties at the same time.  Also, CREATE is a much more complicated statement to parse.  The syntax tree for create is relatively simple.  Whereas CREATE is more COMPLEX.  With CREATE we have CREATE Property MyClass.prop <Type>.  After that we might have:

1. LINKEDCLASS ATTRIBUTES UNSAFE
2. LINKEDCLASS UNSAFE
3. LINKEDCLASS ATTRIBUTES
4. ATTRIBUTES UNSAFE
5. ATTRIBUTES
6. UNSAFE

This gets even more complex if we can't somehow separate these three blocks.  The reason for the parens in my mind is because it clearly indicates the presence of attributes and we can now parse the section as such.  Other wise it is much harder to distinguish it from other parts of the query.

In terms of UNSAFE.  I think this belongs way out there at the end.  Keep in mind that UNSAFE is not about the property.  Meaning, UNSAFE is not an attribute of the property.  Once created you can tell if the property was UNSAFE or not.  It is just an instruction to the executor to not check the database to see if setting the attribute would violate any records that already exist.  To me, something like

```CREATE PROPERTY Foo.bar Integer MANDATORY READONLY UNSAFE``` or 
```CREATE PROPERTY Foo.bar Integer MANDATORY, READONLY, UNSAFE```makes me think that unsafe if a property just like the others.  Where as: ```CREATE PROPERTY Foo.bar Integer (MANDATORY READONLY) UNSAFE``` allows me to see that the two properties are MANDATORY and READONLY.

This whole thing becomes easier if MANDATORY, READONLY, NOTNULL, MIN, MAX, DEFAULT, REGEX, etc. are reserved words. If they are then parsing becomes much easier.  But that would potentially be a breaking change if some one was using that as a property name or something.  This could be looked at for a 3.0 release I suppose, but it seems like they don't want to do that for a 2.2.x which makes sense.

My actual first stab at this and my preferred syntax was something like:

```CREATE PROPERTY Foo.bar Integer MANDATORY READONLY DEFAULT 5 UNSAFE```

But this doesn't work if MANDATORY is not a keyword.  The proposal was to put a comma after 'Integer' as a delimiter to signify we have started talking about attributes.  I found that ugly, and thought the grouping them in PARENS was a better looking and easier to parse solution.

Up for continnued comments from you and the committers on this....

**Comment by smolinari** [comment]
> In terms of ALTER PROPERTY, there are no Parens, or Comma because (as far as I know) ALTER does not allow you to specify multiple attributes of a properties at the same time.

Good point. But, if it did (and it should), there should be consistency. I'm starting to warm up to the parenthesis and commas idea. You could theoretically do something like this too.
```` 
CREATE PROPERTY Foo.bar INTEGER (
                               MANDATORY true,
                               READONLY false,
                               NOTNULL true
                               DEFAULT = 2
                               ) UNSAFE
```
Or something to that effect. Right? :smile: 

So the general syntax rule could be. 

> With only one attribute and value, you don't need to add parentheses, except when `UNSAFE` is being used in the `CREATE PROPERTY` command. For more than one attribute and value, you MUST add parentheses around the attributes and commas between them.

Sounds good to me! :+1: 

Now to change `ALTER PROPERTY` too. :wink: 

Scott

mmacfadden:
Yeah.  ALTER PROPERTY could certainly be extended to allow you to manipulate multiple at the same time, in which case we would want the syntax to be roughly the same, or as similar as possible.  Your multi-line example is correct.  I did have one question.  Why the "=" near default.

There is also the matter of supporting custom attrributes.  Which in alter look like this:

```ALTER PROPERTY Foo.bar DEFAULT 5```
vs.
```ALTER PROPERTY Foo.bar CUSTOM my-prop = 5```

So CUSTOM props use "=" regular ones do not.  I don't know why that is.  Before my time.  But that is how it is.  I think we are getting closer.

**Comment by smolinari** [comment]
> Why the "=" near default.

Ooops...hehe...still had my first suggestion in my head. Corrected.

```
CREATE PROPERTY Foo.bar INTEGER (
                                   MANDATORY true,
                                   READONLY false,
                                   NOTNULL true,
                                   DEFAULT 2
                                 ) UNSAFE
```

And with a custom property.

```
CREATE PROPERTY Foo.bar INTEGER (
                                   MANDATORY true,
                                   READONLY false,
                                   NOTNULL true,
                                   DEFAULT 2,
                                   CUSTOM my-attribute = false
                                 ) UNSAFE
```

Scott

mmacfadden:
Now we're talk'n.

luigidellaquila:
Hi @mmacfadden 

Thank you very much!!! Please give me some time to review it and to discuss the syntax with the team, I'll be back to you very soon

Thanks

Luigi

mmacfadden:
@luigidellaquila Any thoughts on this?

I see that there were some conflicts with the develop branch, so I merged them so it is clean.  Should this be targeted at develop or 2.2.x.  I noticed that it seems like develop is moving towards 3.0.

Thanks.

luigidellaquila:
Hi @mmacfadden 

I did a quick check of the PR and it seems very good, I think I'll manage to finish my review in next few days.

In the meantime we switched develop branch to 3.0 development and we created 2.2.x branch, could you please re-submit your PR on both 2.2.x and develop (rebased)? Sorry for this small inconvenience...

Thanks

Luigi

mmacfadden:
@luigidellaquila Great.  With regards to resubmitting the PR.  Would that be two separate PRs?  I am not sure I know exactly what you want.  But whatever the case.  Happy to do it.

luigidellaquila:
Hi @mmacfadden 

Let me try to merge it to develop, then I'll cherry-pick to 2.2.x
I'll let you know in short

Thanks!

Luigi

mmacfadden:
Awesome.  Quick question.  Is the desired approach now to do PRs on 2.2.x and pull them into develop, or do target them for develop and then pull them into 2.2.x.  I suppose it might depend on if it is supposed to be a quick fix for 2.2.x or if it is part of the major work for 3.0?

luigidellaquila:
Hi @mmacfadden 

Exactly, it depends on the type of contribution, if it's a major change then the best candidate is develop, if it's a fix then the target should be 2.2.x (now the two branches are quite similar, but in the near future they will diverge a lot)

Thanks

Luigi

luigidellaquila:
Merged, thank you very much!

I'm opening a separate issue to update the docs

Thanks

Luigi

PR #6316, opened 2016-06-17
Enhancements and fixes for CREATE PROPERTY

mmacfadden:
Added support for COLLATE and REGEX in the CREATE PROPERTY syntax.  Also determined that we were not handling mixed case values or quoted values properly, so also fixed that while I was as it and added several more tests.

@luigidellaquila will probably want to review and merge.


luigidellaquila:
Thanks @mmacfadden 

I'll check it asap

Luigi

luigidellaquila:
Merged, thank you very much @mmacfadden !

Thanks

Luigi

PR #6410, opened 2016-07-12
Refactor get changes

tglman:
Hi,

This is a medium refactor for remove the double call to the atomic operation for get wal changes, the last commit is the one with more changes but that is just a code cleanup, all the required logic is on the first two commits, I haven't seen a big performance change but i think it still make sense merge it.

another thing, for make the refactor work i had to change this assert: 379257b#diff-4da6f88c14ce4d0fafc152774eeb7508L195 but it may hide a bigger problem, not due to the refactor, it seems that in some cases we lock an release in a not a balanced way, or is better say we load and lock the same page twice doing: lock,lock,release,release causing that assert to fail, this shouldn't be an issue because it happen only for new pages, that exist only in the current thread local, so no concurrency issues, but i would keep an aye on this.

all the rest should be strait forward, let me know if this make sense for you.


laa:
Thank you for patch :-)

PR #6480, opened 2016-07-29
Implemented multi-keys support for HTTP API "GET - Index"

azusa0127:
Basic multi-keys support for HTTP API "GET - Index"
Multi-key format: "key1,key2,key3"

Example:
(GET) http://localhost:2480/index/demo/Courses/CPSC,100

Limitation: there cannot be any comma in a single key, as commas are using as the delimiter of keys.


azusa0127:
ops, just realized that I should pull against develop branch.

PR #6481, opened 2016-07-29
Basic multi-keys support for HTTP API "GET - Index"

azusa0127:
Multi-key format: "key1,key2,key3"

Example:
(GET) http://localhost:2480/index/demo/Courses/CPSC,100

Limitation: there cannot be any comma in a single key, as commas are using as the delimiter of keys.


tglman:
hi @azusa0127,

There is no test case that show that this work, we can evaluate it with test cases.

Regards

PR #6600, opened 2016-08-24
Current Live query works only  final matched datas after crud operation

gibrit:
there is no problem with  logic create and delete
but  in update  there is a logic problem 
for example  if  we have a table  discount  for products .. 

   select name , discount from products where discount>10 limit 5 ; 

and think that result is like that 
    Apple   20
    Peace   15 
    Milk    12
current  update logic  works only   if discount matches  handle  update .. 
first problem  is  if Bread's promotion is changed as  25  ,  it is not update for  this query  it is a new create... 
second problem  if Apple's discount  is changed as  5 , it must a be a delete for this query 
this commit supports  this logic and adds projections to Live query ..

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/query/live/OLiveQueryHook.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/query/live/OLiveQueryHook.java
@@ -29,21 +29,19 @@ import com.orientechnologies.orient.core.db.ODatabaseDocumentInternal;
 import com.orientechnologies.orient.core.db.ODatabaseInternal;
 import com.orientechnologies.orient.core.db.ODatabaseListener;
 import com.orientechnologies.orient.core.db.document.ODatabaseDocument;
-import com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx;
 import com.orientechnologies.orient.core.db.record.ORecordOperation;
 import com.orientechnologies.orient.core.hook.ODocumentHookAbstract;
 import com.orientechnologies.orient.core.record.impl.ODocument;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Created by luigidellaquila on 16/03/15.
  */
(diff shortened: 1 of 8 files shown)


lvca:
@luigidellaquila WDYT about this?

luigidellaquila:
@lvca I had no chance to review the PR yet, but in principle it makes a lot of sense.

@gibrit thank you very much, I'll give you a feedback asap

Thanks

Luigi

gibrit:
is there any way to get affected classtypes   from parsed query ? 
what i want to do . 

1) creating pub/sub of hazelcast  to send  for Live events 
https://hazelcast.org/use-cases/messaging/ in  
OrientDBLiveSystem.publish(token, result); 
OrientDBLiveSystem.listen(token , listener); 

2)  create a single  uid  key for  select * from users where _id=1 query . ( all same queries returns one  same uid  )
3)    token will be created from session ( maybe based on logged user _id)  , databasename,   and  query uid  ( for security reasons)  and query execute will return this  token
4)  register it to the ClassType#registerLiveQuery(token , query)  using a map for registering query  with this uid ...check query uid exist if exists get use registered else use current query    query.addToken(token)  and put the query in map 

5)  i want to create a hook  get the class type of record and call all registered  ClassType#callLiveQueries() this will runs the all queries and   will publish results  with  all registered tokens   using the method  :  OrientDbLiveSystem.push(token, result); 

current live query creates a new  token every time when requested  and registers a new Listener i don't know how current live token system works    but if app crashes  token can be lost and the lost still be there and working ..  for non required live query working.. 

we will have single live query token and listener db type.  and we can use it with different  apps with same trigger 

it will give us possibility to avoid lots of exceptions from  server side  for any reason if system down and restarted at the first time the query when query executed all waiting listeners will be back ..

luigidellaquila:
Hi @gibrit 

Sorry, this fell through the cracks and I completely forgot to check it.
In the meantime we re-implemented live queries for better management of UPDATE statements (now you can check the doc status before and after the update)

The new API is 
```
ODatabase.live(String query, OLiveQueryResultListener listener, Map<String, ?> args);
```

Could you please try it and see if it fits your needs?

Thanks

Luigi

PR #6613, opened 2016-08-26
Allow non-core OrientDB modules to register streamable types with OStreamableHelper

mcculls:
At the moment `OStreamableHelper` relies on `Class.forName` to lookup `OStreamable` types. While that works for types that are on the same classpath as the core module, it doesn't work well in OSGi or other modular systems, where types might be isolated to their defining module.

The simplest solution I found was to let non-core modules register their `OStreamable` types, using weak references so those types can still be unloaded when their defining module is unloaded. (To keep things simple dangling keys aren't cleaned up, but this shouldn't matter because there will only be a few entries in the cache and the keys are simple strings.)

We're using this change to deploy OrientDB distributed on top of Karaf and would also like it backported to 2.2.x. (Note the 2.1.x branch doesn't have this issue because it used a different approach for the distributed protocol.)


mcculls:
Found an edge case which isn't covered by this change, so will rework it and re-push

PR #6614, opened 2016-08-26
Support use of custom ClassLoader in OStreamableHelper

mcculls:
At the moment `OStreamableHelper` relies on `Class.forName` to lookup `OStreamable` types. While that works for types that are on the same classpath as the core module, it doesn't work well in OSGi or other modular systems, where types might be isolated to their defining module.

This PR adds support for customizing the `OStreamableHelper` classloading process so deployers can control how and where streamable types are loaded from, in the same way that the extension loading process can already be configured in `OServer`.


lvca:
Thanks. Backporting it to 2.2.x...

PR #6667, opened 2016-09-07
Fixed out-of-memory issue on initializing direct byte buffer pool.

e-d:
With 32-bit Java and smaller memory footprints, this bug manifests. For example, if you run with a max heap size of 494MB and a MaxDirectMemorySize of 512MB, you will instantly get an out of memory error when trying to initialize a direct memory buffer. This little commit fixes the issue.


laa:
Thank you !

PR #6696, opened 2016-09-14
Correct the date format in HTTP Response Headers

mmacfadden:
The current date header in the http responses looks like this: Sat Jun 11 13:45:12 GMT 2016 This is because the default serialization of Java Dates via the Date.toString() function is "EEE MMM dd HH:mm:ss zzz yyyy"

However, RFC 7231 defines the correct date format to be something like Sun, 06 Nov 1994 08:49:37 GMT, which corresponds to the java date format of "EEE, dd MMM yyyy HH:mm:ss zzz". HTTP Clients are not required to correctly parse non-standard date formats, and we are currently having an issue with this.

This patch brings the date header into compliance with HTTP 1.1, per RFC 7231.

@tglman this is my second attempt.  If changes are needed, just comment and I will updated, rather than closing.  Thanks.


tglman:
thanks, checking

tglman:
Look good, merging.

We still need to be compatible with java8 on 2.2.x so is definitly fine like this, so i can port it.

Regards

PR #6749, opened 2016-09-27
oetl [CSV]: support for missing columns, fix for EXCEL format

fabriziofortino:
List of changes:

* new flag to skip empty columns
* fix for Excel predefined format to correctly handle empty headers
* unit tests


fabriziofortino:
@lvca @luigidellaquila Can you please have a peek at this? I can't handle csv exported from Excel properly. Thanks

Issue #6643, opened 2016-08-31
OObjectEntitySerializer fails identifying binary fields


diegomtassis:
## OrientDB Version, operating system, or hardware.
- v2.2.8
## Expected behavior and actual behavior

`OObjectEntitySerializer.getTypeByClass` should return `OType.BINARY` either for `byte[]` or `Byte[]` fields, but both are identified as `OType.EMBEDDEDLIST`.
## Steps to reproduce the problem

The following unit test demonstrates the bug.

```
public class OObjectEntitySerializerTest {

    // Both test fail because
    // (genericMultiValueType.isPrimitive() && Byte.class.isAssignableFrom(genericMultiValueType)
    // can never be true

    @Test
    public void testGetTypeByClass_PrimitiveBytes() throws NoSuchFieldException, SecurityException {

        // exercise
        OType oType = OObjectEntitySerializer.getTypeByClass(FooEntity.class, "primitiveBytes",
                FooEntity.class.getField("primitiveBytes"));

        // verify
        Assert.assertEquals(OType.BINARY, oType);
    }

    @Test
    public void testGetTypeByClass_WrappingBytes() throws NoSuchFieldException, SecurityException {

        // exercise
        OType oType = OObjectEntitySerializer.getTypeByClass(FooEntity.class, "wrappingBytes",
                FooEntity.class.getField("wrappingBytes"));

        // verify
        Assert.assertEquals(OType.BINARY, oType);
    }

    public static class FooEntity {
        public byte[] primitiveBytes;
        public Byte[] wrappingBytes;
    }
}
```

tglman:
Hi @diegomtassis,

Thanks for the clear test, i triead to fix this, but fix this will change the way how the interface behave on storing byte[], so probably we can fix this just in 3.0 to not break api compatibility.

PR #6769, opened 2016-10-03
Improved approach to store binary data in object API

diegomtassis:
Object API: Possibility to choose via configuration the approach to store binary data (among the ones described in the document API documentation).
- Added interface (strategy pattern) for saving/loading fields from a Document
- ObjectField handling strategies take into account:
  *\* The class schema
  *\* The existing data
  *\* null values
- Added handler for accessing a field handling strategy.
- Aded OGlobalConfiguration entry to select the strategy to be used
- OObjectProxyMethodHandler refactor to use ObjectFieldHandlingStrategy
- Added strategy interface and implementations for custom Otypes
- Added strategy to store binary data in a single ORecordBytes document
- Added strategy to store binary data split in several ORecordBytes documents
- Fixed bug identifying otype binary


diegomtassis:
I think as well this could be backported to 2.2.x.

It is related to this: https://github.com/orientechnologies/orientdb/issues/6643

diegomtassis:
I think as well this could be backported to 2.2.x.

It is related to this: #6643

lvca:
@diegomtassis thanks for this PR, it's well done. We could use the same strategy approach to binding binary fields with Objects and Graphs. @tglman WDYT?

tglman:
This looks definitely good, only one fix is need from my point of view, the strategy should not be static, but associated to the database, and the configuration should be made persistent using the context configuration at database create, so the behavior will not depends on the environment but on the way the database is created.

All the rest is fine, fixed this detail it can be merged.

Bye

diegomtassis:
So you mean reading the configuration value at database creation time, store its value in the context configuration and, from that moment on, everytime the `OObjectFieldHandler` is asked for a strategy, read the value from there and instantiate the strategy. Am I right?

I see the benefit, it is possible to have a different strategy per database.

Is it possible to store in the context configuration the strategy itself? Otherwise the strategy would be created everytime the `OObjectFieldHandler` is requested, which could/should be avoided since the strategy is stateless.

Do you have any other place in the code where you do something like this so I can take a look?

tglman:
hi @diegomtassis,

You can use `database.getConfiguration().setValue` and  `database.getConfiguration().getValue` they use as key a OGlobalConfiguration entry, everything is in that configuration is going to be persisted, so you can add a `setValue` on database create and a `getValue` on database open.

Hope this help ;)

diegomtassis:
Ok, thanks @tglman.

I've added as well a factory and a registry for Strategies, working kind of a dependency injection context. What I store in the database context configuration is the id of the strategy.

I've also renames the classes since they were not tied to be used in the object api. Now it's up to you whether to move them to the core or not.

tglman:
Hi @diegomtassis,

From a quick look seems good, i'll test it and double check with the rest of the team.

I'll let you know and merge it ASAP

Bye

tglman:
Hi,

I checked out this branch and run the tests, they are failing, can you double check this pull ?

Thank You for all your work anyway.

Bye

diegomtassis:
I see.

I've been off, I'm working on it now.

diegomtassis:
I have fixed the failing tests. It was my fault, there were some tests in the "tests" module which needed to be adapted.

However, now new failing tests have showed up.

I don't understand the exact reason behind, but I know the line causing the error (it took me some time to figure it out). It's the change I did in the previous commit in order to store the value read from the configuration in the database context.

`com.orientechnologies.orient.object.db.OObjectDatabaseTx`, line 870.

If I omit that step then all the tests pass, but then I'm not doing what you suggested of saving the configuration value per database instance.

What is odd is that the failing tests fail only when executed in batch. When they are executed isolated everything is fine.

Could you please take a quick look and tell me how to store the configuration value?
I think, though I'm not sure, that maybe the moment I'm accessing the `OContextConfiguration` it is not created yet.

diegomtassis:
Regarding the code quality check, most of the issues are due to code existing already before I started adding changes.

There are some issues as well breaking this rule: Method names should not contain underscores 

I can change the name of these methods, however using underscores in test methods names is something suggested by Roy Osherove in is book "The art of unit testing".

diegomtassis:
I found myself the solution to the problem.

Moving the storage of the setting in the database context configuration from the moment the database is instantiated to the moment it's opened solved the problem.

Everything should be fine now.

tglman:
Hi,

Checked the tests, and all seems fine now, we will merge it soon.

Bye

lvca:
@tglman can we merge this?

tglman:
merged it manually, closing

PR #6809, opened 2016-10-13
Porting of stresstester.sh to windows platform

santo-it:
I have followed the example of other .bat scripts included in the bin
directory, e.g. console.bat

I am not sure if some deploy scripts have to be changed so that this
file can now be included in next releases. Please let me know in case

Thanks,


lvca:
@santo-it we use master only for releases. Could you please send the PR to develop?

PR #6970, opened 2016-12-05
ORIENTDB_ROOT_PASSWORD log message should be info, not warn

mcculls:
We set `ORIENTDB_ROOT_PASSWORD` when embedding OrientDB so the user doesn't have to manually enter it on the console. But when `ORIENTDB_ROOT_PASSWORD` is set there's a log warning which makes it look like something needs fixing, even though everything is ok:
```
Found ORIENTDB_ROOT_PASSWORD variable, using this value as root's password
```
This PR changes this log message from `warn` to `info`, since it's really just informational.


lvca:
Cherry-picked in 2.2.x. Thanks.

Issue #6983, opened 2016-12-08
MATCH optimization is broken: MATCH does not work with multiple class: statements


lodrion:
## OrientDB Version, operating system, or hardware.
- v2.2.12

## Operating System
- [ x] Linux
- [ ] MacOSX
- [ ] Windows
- [ ] Other Unix
- [ ] Other, name? 

## Expected behavior and actual behavior

Adding class: qualifiers to the intermediate clauses in the MATCH query should not cause query to return no results.

## Steps to reproduce the problem

Start with very simple schema with two vertex classes and an edge
```

create class V1 extends V
create class V2 extends V
create class E1 extends E

create vertex V1 set name = 'foo'
create vertex V2 set name = 'bar1'
create vertex V2 set name = 'bar2'

create edge E1 from (SELECT FROM V1) to (SELECT FROM V2)

```

a simple query that roundtrips through E1 works

```

MATCH {
	class: V2,
	as: origin
}.in('E1') {
	as: middle
}.out('E1') {
	as: dest,
	where: (@rid = $matched.origin.@rid)
}
RETURN origin, middle, dest

```

and returns 2 results. 
Adding a class specified to the middle clause 

```

MATCH {
	class: V2,
	as: origin
}.in('E1') {
        class: V1,
	as: middle
}.out('E1') {
	as: dest,
	where: (@rid = $matched.origin.@rid)
}
RETURN origin, middle, dest

```

causes the query to return no results, even through "middle" results from the first query are of class V1.
Removing the where: clause from the third clause causes the query to return (unfiltered) results

```

MATCH {
	class: V2,
	as: origin
}.in('E1') {
        class: V1,
	as: middle
}.out('E1') {
	as: dest
}
RETURN origin, middle, dest

```

And the most interesting thing is that we can make this query to work in a very non-obvious way:
 
```

create vertex V1 set name='foo2'
create vertex V1 set name='foo3'
create vertex V1 set name='foo4'

MATCH {
	class: V2,
	as: origin
}.in('E1') {
        class: V1,
	as: middle
}.out('E1') {
	as: dest,
	where: ($matched.origin.@rid = @rid)
}
RETURN origin, middle, dest

```

And the query has started to work!

Running with the real data of the similar shape, the warning messages that we get when the query fails indicate that this is caused by the query optimizer starting evaluation from the middle clause.

obi1kenobi:
[...]

Issue #6931, opened 2016-11-18
MATCH bug in location existence check


obi1kenobi:
## OrientDB Version, operating system, or hardware.
- v2.0 SNAPSHOT[ ] - .18[ ] .17[ ] .16[ ] .15[ ] .14[ ] .13[ ] .12[ ] .11[ ] .10[ ] .9[ ] .8[ ] .7[ ] .6[ ] .5[ ] .4[ ] .3[ ] .2[ ] .1[ ] .0[ ]
- v2.1 SNAPSHOT[ ] - .16[ ] .15[ ] .14[ ] .13[ ] .12[ ] .11[ ] .10[ ] .9[ ] .8[ ] .7[ ] .6[ ] .5[ ] .4[ ] .3[ ] .2[ ] .1[ ] .0[ ]
- v2.2 SNAPSHOT[ ] - .13[x] .rc1[ ] .beta2[ ] .beta1[ ]

## Operating System
- [x] Linux
- [x] MacOSX
- [ ] Windows
- [ ] Other Unix
- [ ] Other, name? 

## Expected behavior and actual behavior
Expected: If a traversal location named `baz` exists in the output, then checking `$matched.baz IS null` should be false in the MATCH query.

Actual: See below for an instance where data is returned for a location named `baz`, and yet the query evaluates `$matched.baz IS null` as `true`.

### Here's what the formed structure looks like:
```
Foo ==> Bar ==> Baz
  \
   \=> Far
```

### Setup:
```
CREATE CLASS Foo EXTENDS V
CREATE CLASS Bar EXTENDS V
CREATE CLASS Baz EXTENDS V
CREATE CLASS Far EXTENDS V
CREATE CLASS Foo_Bar EXTENDS E
CREATE CLASS Bar_Baz EXTENDS E
CREATE CLASS Foo_Far EXTENDS E

CREATE VERTEX Foo SET name = 'foo'
CREATE VERTEX Bar SET name = 'bar'
CREATE VERTEX Baz SET name = 'baz'
CREATE VERTEX Far SET name = 'far'

CREATE EDGE Foo_Bar FROM (SELECT FROM Foo) TO (SELECT FROM Bar)
CREATE EDGE Bar_Baz FROM (SELECT FROM Bar) TO (SELECT FROM Baz)
CREATE EDGE Foo_Far FROM (SELECT FROM Foo) TO (SELECT FROM Far)
```

### Problematic queries:
- `baz` must always exist because it's not optional
```
[...]

PR #7015, opened 2016-12-16
Added a topological scheduler for MATCH execution.

obi1kenobi:
Also included test cases that check for correct behavior for:
- using undefined aliases in $matched evaluations (error is raised)
- queries with circular dependencies (error is raised)
- complex query execution that requires topological sorting (correct results are returned).

I created this against 2.2.x to speed up the development time and allow for release with 2.2.14. If you approve, I can work with you to adapt it to the `develop` branch.

This is a systematic solution to issues like #6931 and #6983.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OMatchStatement.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OMatchStatement.java
@@ -21,8 +21,6 @@ import com.orientechnologies.orient.core.sql.filter.OSQLTarget;
 import com.orientechnologies.orient.core.sql.query.OBasicResultSet;
 import com.orientechnologies.orient.core.sql.query.OSQLAsynchQuery;
 import com.orientechnologies.orient.core.sql.query.OSQLSynchQuery;
-import com.sun.javafx.collections.MappingChange;
-import com.sun.javafx.geom.Edge;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
@@ -449,36 +447,6 @@ public class OMatchStatement extends OStatement implements OCommandExecutor, OIt
     }
 
     return resultingSchedule;
-  }
-
-  /**
-   * checks if this traversal is consistent, ie. if it can actually be calculated
-   *
-   * @param traversal
-   */
-  private void checkConsistency(List<EdgeTraversal> traversal) {
-    Set<String> matchAliases = new HashSet<String>();
-    for (EdgeTraversal edge : traversal) {
-      PatternNode from = edge.out ? edge.edge.out : edge.edge.in;
-      PatternNode to = edge.out ? edge.edge.in : edge.edge.out;
-
-      String fromAlias = from.alias;
-      OWhereClause fromFilter = aliasFilters.get(fromAlias);
-      if (fromFilter != null && fromFilter.baseExpression != null) {
-        List<String> matchPatternAliases = fromFilter.baseExpression.getMatchPatternInvolvedAliases();
-        if (matchPatternAliases != null) {
-          for (String s : matchPatternAliases) {
-            matchAliases.add(s);
-          }
-        }
-      }
-
-      String toAlias = to.alias;
-      if (matchAliases.contains(toAlias)) {
-        throw new OCommandExecutionException("This query contains MATCH conditions that cannot be evaluated: " + toAlias
-            + " (probably a circular dependency on a $matched condition");
-      }
-    }
   }
 
   protected Object getResult(OSQLAsynchQuery<ODocument> request) {
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OMatchStatement.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OMatchStatement.java
@@ -334,8 +334,17 @@ public class OMatchStatement extends OStatement implements OCommandExecutor, OIt
           // However, if there are edges to the startNode, we must visit the startNode from an already-visited
           // neighbor, to preserve the validity of the traversal. Therefore, we negate the value of isOutbound
           // to ensure that the edge is always scheduled in the direction from the already-visited neighbor
-          // toward the startNode.
-          boolean traversalDirection = !isOutbound;
+          // toward the startNode. Notably, this is also the case when evaluating "optional" nodes -- we always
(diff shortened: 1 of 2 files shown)


obi1kenobi:
@lvca and @luigidellaquila please let me know what you think!

luigidellaquila:
Wow, thank you very much man!
I'm reviewing it now, I hope I'll give you a feedback in a few hours

Thanks

Luigi

luigidellaquila:
Hi @obi1kenobi 

I found a little problem in the algorithm: in a particular situation you reverse the traversal direction

```
 if (!visitedEdges.contains(edge)) {
          // If we are executing in this block, we are in the following situation:
          // - the startNode has not been visited yet;
          // - it has a neighboringNode that has already been visited;
          // - the edge between the startNode and the neighboringNode has not been scheduled yet.
          //
          // The isOutbound value shows us whether the edge is outbound from the point of view of the startNode.
          // However, if there are edges to the startNode, we must visit the startNode from an already-visited
          // neighbor, to preserve the validity of the traversal. Therefore, we negate the value of isOutbound
          // to ensure that the edge is always scheduled in the direction from the already-visited neighbor
          // toward the startNode.
          
          boolean traversalDirection = !isOutbound;
          visitedEdges.add(edge);
          resultingSchedule.add(new EdgeTraversal(edge, traversalDirection));
}

```

In some cases it's not possible, eg. when an edge involves a deep traversal (when it has a WHILE condition or a MAXDEPTH) or when the target node is optional, so you should first check this situation with a

```
edge.item.isBidirectional();
```

Looking at the code, I think that changing the condition is not a solution

```
 if (!visitedEdges.contains(edge) && edge.item.isBidirectional()) {
```

as it will break the whole algorithm. 

Possible solutions here (but no one satisfies me completely):

1. add a `if(!edge.item.isBidirectional()){ continue; }`, but I'm pretty sure that in some situations we will end up with a `throw new OCommandExecutionException("This query contains MATCH conditions that cannot be evaluated")` even if the query could be evaluated following another path

2. abort current evaluation and iterate on another equivalent topological sort, but it complicates the algorithm a bit...

Any thoughts?

Thanks

Luigi

obi1kenobi:
Hi @luigidellaquila, thanks for looking into it.

The direction reversal is not a problem with optional nodes -- in fact, it guarantees that the edge is always considered going *into* the optional node, and not out of it, just the way OrientDB requires it. Please see the test case I included to confirm this.

Would you mind producing a test case for the MAXDEPTH / "while" condition you pointed out will produce an incorrect schedule, so I can use it to debug the problem? The current code satisfies all test cases in the repo, so I don't think that case is covered by the existing code.

Do you think we can merge this in time for 2.2.14? I'd be very happy if that were the case.

luigidellaquila:
Hi @obi1kenobi 

Sure, I'm doing it now. 
2.2.14 should be released tomorrow, with a little luck we can include it ;-)

Thanks

Luigi

obi1kenobi:
Thanks! I'll be AFK in the next three hours to take care of some real-world errands, but I'll look into what you come up with as soon as I get back.

luigidellaquila:
Ok, I have a test case:

```
@Test
  public void testCyclicDeepTraversal() {
    db.command(new OCommandSQL("CREATE CLASS testCyclicDeepTraversalV EXTENDS V")).execute();
    db.command(new OCommandSQL("CREATE CLASS testCyclicDeepTraversalE EXTENDS E")).execute();

    db.command(new OCommandSQL("CREATE VERTEX testCyclicDeepTraversalV SET name = 'a'")).execute();
    db.command(new OCommandSQL("CREATE VERTEX testCyclicDeepTraversalV SET name = 'b'")).execute();
    db.command(new OCommandSQL("CREATE VERTEX testCyclicDeepTraversalV SET name = 'c'")).execute();
    db.command(new OCommandSQL("CREATE VERTEX testCyclicDeepTraversalV SET name = 'z'")).execute();

    // a -> b -> z
    // z -> c -> a
    db.command(new OCommandSQL("CREATE EDGE testCyclicDeepTraversalE from"
        + "(select from testCyclicDeepTraversalV where name = 'a') to (select from testCyclicDeepTraversalV where name = 'b')"))
        .execute();

    db.command(new OCommandSQL("CREATE EDGE testCyclicDeepTraversalE from"
        + "(select from testCyclicDeepTraversalV where name = 'b') to (select from testCyclicDeepTraversalV where name = 'z')"))
        .execute();

    db.command(new OCommandSQL("CREATE EDGE testCyclicDeepTraversalE from"
        + "(select from testCyclicDeepTraversalV where name = 'z') to (select from testCyclicDeepTraversalV where name = 'c')"))
        .execute();

    db.command(new OCommandSQL("CREATE EDGE testCyclicDeepTraversalE from"
        + "(select from testCyclicDeepTraversalV where name = 'c') to (select from testCyclicDeepTraversalV where name = 'a')"))
        .execute();

    // "bar" in the following query declares a dependency on the alias "baz", which doesn't exist.
    OSQLSynchQuery query = new OSQLSynchQuery(
        "MATCH " + "{ class: testCyclicDeepTraversalV, as: foo, where:(name ='a')}.out() {while: ($depth < 2), where: (name = 'z'), as: bar},"
            + "{ as: bar}.out() {while: ($depth < 2), as: foo}" + " RETURN $patterns");

    List<?> result = db.query(query);
    assertEquals(1, result.size());

  }
```

BTW, I'll add it to 2.2.x anyway, it was definitely missing

Thanks!

Luigi

obi1kenobi:
I assume the `// "bar" in the following query declares a dependency on the alias "baz", which doesn't exist.` comment is erroneously left over from a copy-paste?

luigidellaquila:
Ouch! Yes, sorry, just a copy-paste ;-)

obi1kenobi:
@luigidellaquila I added your test case and some extra logic to cover it -- thanks for the help!

I think this scheduler is definitely an improvement over the old approach, and it passes all tests. Unless there are any remaining serious edge cases you can come up with, please consider releasing this in 2.2.14.

obi1kenobi:
Fantastic, thank you! Let me know if you need help to merge these changes into the `develop` branch as well.

luigidellaquila:
I merged them to develop today, I only have to check that they are correctly used by the new executor as well, but it's an easy task.

Thanks!

Luigi

PR #7020, opened 2016-12-18
Fix method releasePage and reset

areumchoi:
I fixed empty method, releaseFilePage and reset.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/storage/cache/pages/OPassthroughPageCache.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/storage/cache/pages/OPassthroughPageCache.java
@@ -57,7 +57,7 @@ public class OPassthroughPageCache implements OPageCache {
 
   @Override
   public void releaseFilePages(long fileId, OWriteCache writeCache) {
-    // do nothing
+    releasePage(fileId, writeCache);
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/orientechnologies/orient/core/storage/cache/pages/OPassthroughPageCache.java
```diff
@@ -57,7 +57,7 @@ public void releasePage(OCacheEntry cacheEntry, OWriteCache writeCache) {
 
   @Override
   public void releaseFilePages(long fileId, OWriteCache writeCache) {
-    // do nothing
+    releasePage(fileId, writeCache);
```
taburet: Passthrough cache caches nothing, so there is nothing to release from it. I doubt this even compiles, since there is no `releasePage` method accepting `long` as a first argument. "do nothing"-like comment means that the method is intentionally left blank.

core/src/main/java/com/orientechnologies/orient/core/storage/cache/pages/OPassthroughPageCache.java
```diff
@@ -67,7 +67,7 @@ public OCacheEntry purgePage(long fileId, long pageIndex, OWriteCache writeCache
 
   @Override
   public void reset(OWriteCache writeCache) {
-    // do nothing
+    writeCache = null;
```
taburet: This just nulls out the method-local parameter value.

Issue #7103, opened 2017-01-20
A* SQL function returns paths of increasing length for consecutive, identical queries


martineutinger:
## OrientDB Version, operating system, or hardware.
- v2.2.14

## Operating System
- [ ] Linux
- [ ] MacOSX
- [x] Windows
- [ ] Other Unix
- [ ] Other, name? 

## Expected behavior and actual behavior
When sending the same query (e.g. 'SELECT astar(#23:19965, #23:19187, 'length')') more than once, only the first result is correct. The consecutive results are longer and longer paths until it seems to either plateau or return no path at all. 

So e.g. the shortest path contains 5 nodes.
The first query would return the 5 nodes. The second would return 10. Then 17. And then it returns 21 for every query past that point.

## Steps to reproduce the problem
- Start the server, or connect to a local database. 
- Send a query like 'SELECT astar(#23:19965, #23:19187, 'length')' multiple times WITHOUT restarting the server in between each query.

## Possible solutions
The problem seems to be, that in com.orientechnologies.orient.graph.sql.functions.OSQLFunctionAstar
'PriorityQueue<OrientVertex> open' and 'Set<OrientVertex> closedSet' are carried over from the last execution of the algorithm.
It seems like ...
... either .clear() should be called for 'open' and 'closedSet' before each execution
... or a new instance of OSQLFunctionAstar should be created, every time the function is executed.

## Misc
This is not the case with Dijkstra's algorithm, because a new instance of A* is created each time.

saeedtabrizi:
Hi @martineutinger  , @luigidellaquila  
yes i detect some bugs in the astar function too . i check it asap .

luigidellaquila:
Thank you @saeedtabrizi, I'll check it as well

Thanks!

Luigi

saeedtabrizi:
Hi @martineutinger , @luigidellaquila 
I fix some bugs (reported bug) and more test to `astar` function now . i will push it asap .

luigidellaquila:
Great, thank you very much @saeedtabrizi !

Looking forward to the pr!

Thanks

Luigi

saeedtabrizi:
@luigidellaquila the bugs fixed now .

martineutinger:
Thank you @saeedtabrizi.

So does that mean the single-astar-instance-per-server is by design?

saeedtabrizi:
[...]

Issue #7081, opened 2017-01-14
Better support for time expression


lvca:
It would be cool having something like:

```sql
select * from Log where date >= date() - 1h
```

In this case `h` stands for hour. We could have `d` for days. This query extract all the logs of last 24h:

```sql
select * from Log where date >= date() - 1d
```

Eric24:
Interesting. Maybe look at [momentjs](http://momentjs.com), with a syntax like: date().add(1, 'd') or date().subtract(5, 'h'), plus options like: date().startOf('Y') and date().endOf('M')

saeedtabrizi:
Thanks to @lvca ,  @Eric24 for recommendation . 
let me add INTL and support Calendars (Like the gregorian,[japanese](http://icu-project.org/~yoshito/jacoco_57.1/com.ibm.icu.util/JapaneseCalendar.java.html),[islamic](http://icu-project.org/~yoshito/jacoco_57.1/com.ibm.icu.util/IslamicCalendar.java.html) ,  [persian](http://icu-project.org/~yoshito/jacoco_57.1/com.ibm.icu.util/PersianCalendar.java.html)  and etc... ) may useful in date and time computation . its a big **IMPORTANT** to achieve more users in global market .  
i think we have following syntax exactly : 

**add** :     select * from Log where date >= calendar('persian',date).add('YEAR',5).toDate()
**add** :     select * from Log where date >= calendar('hijri',date).add('MONTH',8).toDate()
**add** :     select * from Log where date >= calendar('hijri',date).add('DAY',8).toDate()  
**add** :     select * from Log where date >= calendar('persian',date).add('HOUR',5).toDate()
**add** :     select * from Log where date >= calendar('hijri',date).add('MINUTE',8).toGMTDate()
**add** :     select * from Log where date >= calendar('hijri',date).add('SECOND',8).toISODate()
**add** :     select * from Log where date >= calendar('hijri',date).add('WEEK',8).toUTCDate()

**subtract** :     select * from Log where date >= calendar('persian',date).subtract('HOUR',6).toDate()
**subtract** :     select * from Log where date >= calendar('persian',date).add('DAY',5).subtract('DAY',3).toDate() // returns 1395-02-01
**subtract** :     select * from Log where date >= calendar('persian',date).subtract('DAY',3).toFormat("yyyy-mm-dd") // returns 139511-02
[...]

PR #7131, opened 2017-01-27
add toCalendar method  #7081 , and fix astar graph function #7103 some bugs

saeedtabrizi:
- fix bug report for astar graphdb in #7103 issue 
-fix null checking in astar function .
- add toCalendar method as we discussed on Better support for time expression #7081.
- add some test cases for toCalendar Method . (need to be documented) .

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -245,6 +245,13 @@
             <version>3.0.1</version>
             <scope>provided</scope>
         </dependency>
+
+        <!-- https://mvnrepository.com/artifact/com.ibm.icu/icu4j -->
+        <dependency>
+            <groupId>com.ibm.icu</groupId>
+            <artifactId>icu4j</artifactId>
+            <version>58.2</version>
+        </dependency>
     </dependencies>
 
 </project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -246,7 +246,7 @@
             <scope>provided</scope>
         </dependency>
 
-        <!-- https://mvnrepository.com/artifact/com.ibm.icu/icu4j -->
+        <!-- https://mvnrepository.com/artifact/com.ibm.icu/icu4j 58 -->
         <dependency>
             <groupId>com.ibm.icu</groupId>
             <artifactId>icu4j</artifactId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -246,11 +246,11 @@
             <scope>provided</scope>
         </dependency>
 
-        <!-- https://mvnrepository.com/artifact/com.ibm.icu/icu4j 58 -->
+        <!-- https://mvnrepository.com/artifact/com.ibm.icu/icu4j 59.1 -->
         <dependency>
             <groupId>com.ibm.icu</groupId>
             <artifactId>icu4j</artifactId>
-            <version>58.2</version>
+            <version>59.1</version>
         </dependency>
     </dependencies>
 
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/method/intl/OCalendar.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/method/intl/OCalendar.java
@@ -1,0 +1,607 @@
+/*
+ * Copyright 2010-2016 OrientDB LTD (http://orientdb.com)
+ * Copyright 2013 Geomatys.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.orientechnologies.orient.core.sql.method.intl;
+
+import com.ibm.icu.util.*;
+import com.orientechnologies.orient.core.command.OCommandContext;
+import com.orientechnologies.orient.core.db.record.OIdentifiable;
+import com.orientechnologies.orient.core.util.ODateHelper;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.time.Year;
+import java.util.Date;
+import java.util.TimeZone;
+
+/**
+ * @author Saeed Tabrizi (saeed a_t  nowcando.com)
+ */
+public class OCalendar {
+
+    static final String CAL_SATURDAY = "saturday";
+    static final String CAL_SUNDAY = "sunday";
+    static final String CAL_MONDAY = "monday";
+    static final String CAL_TUESDAY = "tuesday";
+    static final String CAL_WEDNESDAY = "wednesday";
+    static final String CAL_THURSDAY = "thursday";
+    static final String CAL_FRIDAY = "friday";
+
+    static final String YEAR = "year";
+    static final String QYEAR = "qyear";// shortcut 90 days
+    static final String HYEAR = "hyear";// shortcut 180 days
+    static final String MONTH = "month";
+    static final String DAY = "day";
+    static final String WEEK = "week"; // shortcut 7 days
+    static final String HOUR = "hour";
+    static final String MINUTE = "minute";
+    static final String SECOND = "second";
+
+    static final String CAL_GREGORIAN = "gregorian";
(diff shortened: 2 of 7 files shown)


luigidellaquila:
Hi @saeedtabrizi 

Thank you very much, I'm checking the PR about the calendar.
About the A* function I'm afraid your fix does not work in multi-thread... anyway I found the root cause and fixed it, the problem was how the function was registered in OFunctionFactory:
```
register(OSQLFunctionAstar.NAME, new OSQLFunctionAstar());
```
This way the server has a single instance of the function. I changed it with
```
register(OSQLFunctionAstar.NAME, OSQLFunctionAstar.class);
```
so now the server returns a new instance at every invocation. 

Could you please re-submit the PR removing the changes to A*

Thanks

Luigi

saeedtabrizi:
Hi @luigidellaquila . 
Thanks to solve the root of problem for multiple  instance in A* function , but this PR have the some null  value checking that would be in the A* function . i detect this problem when i review the A* code  and i write a test method to cover it . 
So this PR not only related to A* multiple function and have the changes for the A* null checking too .

luigidellaquila:
Got it, thank you @saeedtabrizi 
I'll merge it with the rest then

Thanks!

Luigi

saeedtabrizi:
@luigidellaquila , Is there any chance to have this merge in 2.2.17 release and change the milestone from 3 to 2.2.17 ?

lvca:
Hi @luigidellaquila what's the status of this?

saeedtabrizi:
Hi @lvca , @luigidellaquila 
Is there any chance to have this PR in ODB 3.0 :)

saeedtabrizi:
@lvca , @luigidellaquila I guess we have not any chance to have this pull request in ODB v3.0 .

luigidellaquila:
Hi @saeedtabrizi 

checking it now. Just a couple of doubts:

- the A* function is already fixed, could you please remove that part from the PR?
- I see you added a new library (ICU4J) that has a different license from Apache2. I had a quick look at that, but I'd like @lvca to double-check and make sure that it is OK
- as far as I can see the toCalendar() method is not registered in `ODefaultSQLMethodFactory` so it won't work from SQL. I think it's worth registering it and adding some SQL tests, WDYT?
- I guess the main purpose of this PR is to provide an API to modify dates from SQL, but to do this we also will have expose some additional methods as OSQLMetods, like add() subtract() and so on, did I get it right?

Thanks

Luigi

saeedtabrizi:
Hi @luigidellaquila ,
First of all i'm observing ODB 3.0 changes . that must be great and i appreciate  to orientdb team .

- For A* function i wrote some  test and fixing the null value  checking . but your fixture completely solve the problem and i agree to remove from PR .

- For the licencing and permission of ICU4J we have already discussed in https://github.com/orientechnologies/orientdb/issues/7081 . some major platform , library or applications like nodejs using the ICU .

- For the toCalendar() registering in the ODefaultSQLMethodFactory , i have not any idea and i'm agree for any changes that help us to using calendar functionalities in ODB .

As a summary , this PR goal is adding the calendar calculation and time methods to ODB 3.0 via ICU4J . i write a wrapper  class and methods too . 

Thanks 
Saeed Tabrizi

saeedtabrizi:
Hi @luigidellaquila 
Is there any chance to have merged this PR in ODB 3.0 M2 ?

Thanks

luigidellaquila:
Hi @saeedtabrizi 

Sure, my intention is to merge it, but to do it you should remove the changes to AStar, as it's already fixed in the main branch.
About the registration of `toCalendar()` in the function factory, there is no need to do it in the PR, I can do it separately

Thanks

Luigi

saeedtabrizi:
Hi @luigidellaquila 
Ok i remove the Astar changes  and push it again .
Thanks

saeedtabrizi:
Hi @luigidellaquila ,
As i review my code i have the following changes in `Astar` function that must be important to include for Astar safe running .

 ```
  open.clear();
    closedSet.clear();
    route.clear();
    currentDepth=0;
```

As i see the history of `OGraphFunctionFactory` file i have not any modification about it and you can merge it without any problem . 

So please merge  this PR and put it to 3.0 milestone . 
I just started to work about anothe graph sql functions that we discussed in https://github.com/orientechnologies/orientdb/issues/7132#issuecomment-320417279 . 

Thanks . 
Saeed Tabrizi

**Comment by luigidellaquila** [comment]
Hi @saeedtabrizi 

I did a deep review of this PR and I still have a couple of doubts.

I guess the main purpose of this PR is to have a way to manipulate dates in SQL, if I'm wrong please correct me, but I cannot find any other rationale
The problem is that as it is it, won't work in SQL for at least three reasons:

1. the function toCalendar() is not registered as a SQL function, so it just doesn't work from SQL. As I wrote before, I can easily add it, so it's not a problem

2. the function returns an OCalendar object, that is not a standard OrientDB type, so the binary protocol doesn't know how to serialize it. 

To make it clear, the following query will fail in remote

```
SELECT toCalendar( "20151004T00:00:00") as theCal
```
because the protocol doesn't know how to serialize `theCal`

3. I guess the main goal of OCalendar is to manipulate dates in SQL (again, correct me if I'm wrong), so it has some Java methods like `add()`, `subtract()`, `getMonth()` and so on. These methods are not exposed as SQL methods, so they won't be accessible in SQL queries, eg. the following
```
SELECT toCalendar( "20151004T00:00:00").getMonth()
```
will fail because `getMonth()` is not a registered SQL method.
To fix this, we should write a lot of separate OSQLMethod implementations, one for each method on OCalendar.

Please help me to understand if I'm missing something, my feeling is that we still have a lot of things to do to merge this PR...

Thanks

Luigi

saeedtabrizi:
Hi @luigidellaquila 
The goal  of using calendar base date & time calculation is achieving more relaxation in globalized application queries . so i recommended to have an extension **method** for Date data type exactly so its not a function . 
The true syntax for working with calendar must be  :
`Select Birthday , Birthday.toCalendar("persian").getMonth() as PBM from User `
Result :
    Birthday  | PBM
----------------|-------
  1982-3-22 | 1       // The thirth month of gregorian calendar equals to first month of persian calendar . 
------------------------
That Birthday is a DateTime data type and `toCalendar` method returns an instance of special calendar that implemented some methods . it just very similar to  `format` method for `DateTime` data type in orientdb . in the other word my template for DateTime method extending is the date `format` method implementation in the ODB .
I have defined multiple sample for added methods and functions for implementing calendar and time span calculation in https://github.com/orientechnologies/orientdb/issues/7081 issue. 

I hope to transfer my idea for calendar implementation clearly . 

Thanks 
Saeed  Tabrizi

lvca:
Very cool and useful contribution, +1. @saeedtabrizi please look at the output from Codacy for static warnings and errors: https://www.codacy.com/app/OrientDB/orientdb/pullRequest?prid=495581.

saeedtabrizi:
Many Thanks @lvca . I just lint the code  and try to cleanup errors and warning :)

lvca:
I was reading Luigi's comments and he's right. You should register a lot of new methods in order to be used in SQL. At that point, the question is: why using `getCalendar()` if you could apply such methods directly to the date types? Like `select birthday.getMonth()`. `getCalendar()` could be useful only if you want to change calendar at this point. Am I missing anything?

saeedtabrizi:
@lvca the short answer is yes . i use `toCalendar` method for instantiate a calendar that using in date calculation operations . 
So i have some mistakes to implementing the date and calendar calculation related methods in orientdb . 
In my case the `OCalendar` class have an wrapper role exactly and when calling the `toCalendar` method in  `DateTime` instances an `OCalendar` class instantiated and all of `OCalendar` method now can call by the user . 
It is so clear and simple .  but as i understand from you and dear @luigidellaquila comments, i should implement all methods of `OCalendar` class  as a orientdb sql method too . is it true  ? 
for example if i have support the following syntax :
` select * from Log where logDate >= logDate.toCalendar('persian').add('YEAR',5).toDate() ;`

Which `.toCalendar('persian')` returns an `OCalendar` instance and `.add('YEAR',5)` accumulates 5 years adding to `OCalendar` instance  and returns `OCalendar` and finally by calling `.toDate()` method of `OCalendar` it just returns a standard `Date` datatype , I should implement an `add()` method and a `toCalendar()` and a `toDate()` method in orientdb sql methods.
Please correct me if i miss understood .    

Thanks Saeed .

saeedtabrizi:
Hi @lvca  , @luigidellaquila 
Is there any alternative to support calendar and time utils in ODB ?

luigidellaquila:
Hi @saeedtabrizi 

I'm afraid for now there are no further alternatives, we are more or less at the same point of when this issue was reported

Thanks

Luigi

PR #7230, opened 2017-03-13
Fix 7229

peturrun:
The fundamental problem seems to be that OrientVertex.getFieldNames() loops over all relevant classes in the schema and calculates the corresponding field names. The mappings between field names and classes is then discarded and only the field names returned. In getEdges(), there is a loop over the returned field names, which calls getConnection() for each field name to get back the class name. Within getConnection(), there is another loop over all the relevant classes, causing the quadratic behavior.

This PR attempts to fix this by replacing getFieldNames with a new method that returns the class information in addition to the field names, completely avoiding the need to call getConnection when this information is available. The call to getAllSubClasses and the nested loop in getConnection are also replaced with a call to isSubClassOf.


luigidellaquila:
Hi @peturrun 

I see you did this PR on develop branch, that is for next 3.0.

If you need this fix on on 2.2.x could you please do the PR on 2.2.x branch?

Thanks

Luigi

peturrun:
Hi Luigi

Here is the same PR for the 2.2.x branch:

https://github.com/orientechnologies/orientdb/pull/7231

tglman:
Same of the 2.2.x branch, merging

PR #7341, opened 2017-04-15
Avoid extra piping with grep

BooleanCat:
Matching on "[p]attern"  will not include the grep itself from the process list.

https://unix.stackexchange.com/questions/74185/how-can-i-prevent-grep-from-showing-up-in-ps-results


robfrank:
Could you please perform the PR against develop branch?

PR #7360, opened 2017-04-24
Case sensitive schema refactoring

luigidellaquila:
**Goal**: 
- fix an old problem regarding case sensitivity of properties in schema. Until now, properties where case sensitive while schemaless. When you declared schema for a property, it became case insensitive. 

**Side effects**:
- Index names are now CASE SENSITIVE!

**Challenges**: 
- when creating an index from Property API, its name is assigned as `ClassName.propertyName`. Index file names are the same as the index name. Defining two properties with colliding case insensitive names (eg. `Person.firstname` and `Person.firstName`) would lead to two index file names that collide on case insensitive file systems (Windows systems and the vast majority of OSX instances), so we had to refactor index file name generation to avoid such a collision

**Code Changes**:
- a big change to low level index API to explicitly pass file name when creating an index 
- a change to ODurableComponent and to index metadata to store the file name
- removal of some (well, more than fourty...) `toLowerCase()` in the Schema implementation to make properties case sensitive
- increase OStorageConfiguration.CURRENT_VERSION number

**Open/Critical Points**:
- please check that I didn't miss anything in the refactoring ( @laa )
- I had to do a fix on ODirectMemoryOnlyDiskCache to make tests pass (line 367). Please check it, because it's quite obscure...  ( @lvca @laa )
- Please double check autosharding index behavior ( @lvca )
- Is there anything specific to do for distributed? ( @lvca )
- Geospatial indexes will likely need a little fix ( @maggiolo00 )

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/index/OIndexAbstract.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/index/OIndexAbstract.java
@@ -76,6 +76,7 @@ public abstract class OIndexAbstract<T> implements OIndexInternal<T> {
   protected final OAbstractPaginatedStorage storage;
   private final   String                    databaseName;
   private final   String                    name;
+  private String fileName;
   private final OReadersWriterSpinLock rwLock         = new OReadersWriterSpinLock();
   private final AtomicLong             rebuildVersion = new AtomicLong();
   private final      int                version;
@@ -88,7 +89,7 @@ public abstract class OIndexAbstract<T> implements OIndexInternal<T> {
   private volatile boolean             rebuilding       = false;
   private          Map<String, String> engineProperties = new HashMap<String, String>();
 
-  public OIndexAbstract(String name, final String type, final String algorithm, final String valueContainerAlgorithm,
+  public OIndexAbstract(String name, String fileName,  final String type, final String algorithm, final String valueContainerAlgorithm,
       final ODocument metadata, final int version, final OStorage storage) {
     acquireExclusiveLock();
     try {
@@ -96,6 +97,7 @@ public abstract class OIndexAbstract<T> implements OIndexInternal<T> {
 
       this.version = version;
       this.name = name;
+      this.fileName = fileName;//TODO!!!
       this.type = type;
       this.algorithm = algorithm;
       this.metadata = metadata;
@@ -114,6 +116,7 @@ public abstract class OIndexAbstract<T> implements OIndexInternal<T> {
   public static OIndexMetadata loadMetadataInternal(final ODocument config, final String type, final String algorithm,
       final String valueContainerAlgorithm) {
     final String indexName = config.field(OIndexInternal.CONFIG_NAME);
+    final String fileName = config.field(OIndexInternal.CONFIG_FILE_NAME);
 
     final ODocument indexDefinitionDoc = config.field(OIndexInternal.INDEX_DEFINITION);
     OIndexDefinition loadedIndexDefinition = null;
@@ -168,7 +171,7 @@ public abstract class OIndexAbstract<T> implements OIndexInternal<T> {
 
     final Set<String> clusters = new HashSet<String>((Collection<String>) config.field(CONFIG_CLUSTERS, OType.EMBEDDEDSET));
 
-    return new OIndexMetadata(indexName, loadedIndexDefinition, clusters, type, algorithm, valueContainerAlgorithm);
+    return new OIndexMetadata(indexName, fileName, loadedIndexDefinition, clusters, type, algorithm, valueContainerAlgorithm);
   }
 
   public void flush() {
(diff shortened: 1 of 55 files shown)


core/src/main/java/com/orientechnologies/orient/core/index/OIndexAbstract.java
```diff
@@ -88,14 +89,15 @@
...
-  public OIndexAbstract(String name, final String type, final String algorithm, final String valueContainerAlgorithm,
+  public OIndexAbstract(String name, String fileName,  final String type, final String algorithm, final String valueContainerAlgorithm,
       final ODocument metadata, final int version, final OStorage storage) {
     acquireExclusiveLock();
     try {
       databaseName = storage.getName();
 
       this.version = version;
       this.name = name;
+      this.fileName = fileName;//TODO!!!
```
laa: Why TODO is added here?

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/paginated/base/ODurableComponent.java
```diff
@@ -65,19 +64,22 @@
...
+  public ODurableComponent(OAbstractPaginatedStorage storage, String name, String fileName, String extension, String lockName) {
     super(true);
 
     assert name != null;
+    assert fileName != null;
     this.extension = extension;
     this.storage = storage;
-    this.fullName = name + extension;
+    this.fileName = fileName;
+    this.fullName = fileName + extension;
```
laa: That is incorrect you mix full component name and full file name. For example, in start operation, you should use getFullName but when you add a file you should use gitFullFileName.

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/paginated/base/ODurableComponent.java
```diff
@@ -98,7 +98,6 @@ public String getName() {
 
   public void setName(String name) {
     this.name = name;
-//    this.fullName = name + extension;
```
laa: Again please do not comment out the code )).

luigidellaquila: Already removed in last commit (see the full PR)

tests/src/test/java/com/orientechnologies/orient/test/database/auto/IndexTest.java
```diff
@@ -1658,9 +1658,18 @@ public void testValuesContainerIsRemovedIfIndexIsRemoved() {
...
+    Assert.assertTrue(existsFile(writeCache.files().keySet(), "ValuesContainerIsRemovedIfIndexIsRemovedIndex", ".irs"));
     database.command(new OCommandSQL("drop index ValuesContainerIsRemovedIfIndexIsRemovedIndex")).execute();
-    Assert.assertTrue(!writeCache.exists("ValuesContainerIsRemovedIfIndexIsRemovedIndex.irs"));
+    Assert.assertFalse(existsFile(writeCache.files().keySet(), "ValuesContainerIsRemovedIfIndexIsRemovedIndex", ".irs"));
+  }
+
+  private boolean existsFile(Set<String> strings, String prefix, String suffix) {
+    for (String s : strings) {
+      if (s.startsWith(prefix) && s.endsWith(suffix)) {
+        return true;
```
luigidellaquila: Good point, this should definitely be fixed

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/paginated/base/ODurableComponent.java
luigidellaquila: I just replicated the existing logic, just using fileName instead of name. I really don't like it either, probably I can do better

luigidellaquila:
Hi @laa 

I agree with you, I don't like this general approach either. I tried to do it at the low level but I failed. The problem is that it already leaked at the high level.

Let me rework it and see what I can do

Thanks

Luigi

core/src/main/java/com/orientechnologies/orient/core/index/OIndexEngine.java
```diff
@@ -46,8 +46,15 @@ void create(OBinarySerializer valueSerializer, boolean isAutomatic, OType[] keyT
...
-  void load(String indexName, OBinarySerializer valueSerializer, boolean isAutomatic, OBinarySerializer keySerializer,
+//  void load(String indexName, OBinarySerializer valueSerializer, boolean isAutomatic, OBinarySerializer keySerializer,
+//      OType[] keyTypes, boolean nullPointerSupport, int keySize, Map<String, String> engineProperties);
+
+//  default
+  void load(String indexName, String fileName, OBinarySerializer valueSerializer, boolean isAutomatic, OBinarySerializer keySerializer,
       OType[] keyTypes, boolean nullPointerSupport, int keySize, Map<String, String> engineProperties);
+//  {
+//    load(indexName, valueSerializer, isAutomatic, keySerializer, keyTypes, nullPointerSupport, keySize, engineProperties);
+//  }
```
luigidellaquila: it was already removed in last commit

core/src/main/java/com/orientechnologies/orient/core/index/OIndexAbstract.java
luigidellaquila: it was already removed in last commit

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/paginated/base/ODurableComponent.java
```diff
@@ -96,7 +98,16 @@ public String getName() {
 
   public void setName(String name) {
     this.name = name;
-    this.fullName = name + extension;
+//    this.fullName = name + extension;
```
luigidellaquila: this was removed too in last commit

Issue #2573, opened 2014-07-14
Extends OrientDB SQL with new functions


lvca:
## FOR functions

Like Java for cycle. Syntax:

```
for( <initial-expression>; <expression to check at every cycle>; <expression to execute at every cycle> )
```
## Math functions

A good idea could be bind all the math functions in one shot by wrapping the Java Math class. Syntax: 

```
math.<math-function(<args>*)
```

Example:

  select math.sin( x ) from points

(`split()` already supported in 2.2)

fabriziofortino:
@lvca I have implemented a new OSQLFunction that uses reflection to easily bind any static method (with overload support). It is extremely easy to bind all the static methods of a given class. I have added the support for the java.lang.Math methods. The only problem is the prefix: I cannot used dots in the functions' names (eg: math.random). OSQLFilterItemAbstract (line 59) splits the string removing the part before the dot. I have currently removed the prefix but some names clash with the existing functions (eg: max, min). Any idea on how to sort it out?

You can see the changes here:
https://github.com/fabriziofortino/orientdb/commit/71f37b34717a3f5544c1cfd1fca9432644c340f9

lvca:
We could introduce the concept of package in function (we should have some issues about it). What about: `Math::max()` ?

fabriziofortino:
The introduction of packages sounds good to me. I guess we should modify the parser to support it. I have no experience with javacc though. Any luck to add this feature soon? 

With the current implementation, adding the prefix you proposed I get the following exception

```
com.orientechnologies.orient.core.sql.parser.ParseException: Encountered " ":" ": "" at line 1, column 12.
Was expecting one of:
    <EOF> 
    <TIMEOUT> ...
    <AS> ...
    <FETCHPLAN> ...
    <LOCK> ...
    <LET> ...
    <NOCACHE> ...
    <PARALLEL> ...
    <UNWIND> ...
    ";" ...
    "," ...
    <AS> ...
    "," ...
    
	at com.orientechnologies.orient.core.sql.parser.OrientSql.generateParseException(OrientSql.java:11559)
	at com.orientechnologies.orient.core.sql.parser.OrientSql.jj_consume_token(OrientSql.java:11432)
	at com.orientechnologies.orient.core.sql.parser.OrientSql.parse(OrientSql.java:88)
	at com.orientechnologies.orient.core.sql.OCommandExecutorSQLAbstract.preParse(OCommandExecutorSQLAbstract.java:234)
[...]

PR #7406, opened 2017-05-09
OSQLStaticReflectiveFunction to automatically bind static methods

fabriziofortino:
* Added mechanism to automatically register static methods of a given class to sql functions
* Registered `java.lang.Math` methods with `math_` prefix

#2573

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/functions/misc/OSQLStaticReflectiveFunction.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/functions/misc/OSQLStaticReflectiveFunction.java
@@ -86,19 +86,19 @@ public class OSQLStaticReflectiveFunction extends OSQLFunctionAbstract {
       OCommandContext iContext) {
 
     final Supplier<String> paramsPrettyPrint = () ->
-        Arrays.stream(iParams).map(p -> p + " [ " + p.getClass().getName() + " ]").collect(Collectors.joining());
+        Arrays.stream(iParams).map(p -> p + " [ " + p.getClass().getName() + " ]").collect(Collectors.joining(", ", "(", ")"));
 
     Method method = pickMethod(iParams);
 
     if (method == null) {
-      throw new OQueryParsingException("Unable to find a function with name " + name + " and parameters " + paramsPrettyPrint.get());
+      throw new OQueryParsingException("Unable to find a function for " + name + paramsPrettyPrint.get());
     }
 
     try {
       return method.invoke(null, iParams);
-    } catch (ReflectiveOperationException e) {
+    } catch (ReflectiveOperationException | IllegalArgumentException e) {
       e.printStackTrace();
-      throw new OQueryParsingException("Error executing function " + name + " with parameters " + paramsPrettyPrint.get());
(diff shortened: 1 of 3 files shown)


luigidellaquila:
Hi @fabriziofortino 

Thank you, I had a quick look at this and I like it very much.
I'll do a full review in next hours and keep you posted

Thanks!

Luigi

fabriziofortino:
Good to hear that @luigidellaquila 

With OSQLStaticReflectiveFunction we could easily register a number of other useful functions (I have some idea in mind). 

Looking forward to your comments.

core/src/main/java/com/orientechnologies/orient/core/sql/functions/misc/OSQLStaticReflectiveFunction.java
```diff
@@ -0,0 +1,170 @@
...
+    });
+  }
+
+  @Override
+  public Object execute(Object iThis, OIdentifiable iCurrentRecord, Object iCurrentResult, Object[] iParams,
+      OCommandContext iContext) {
+
+    Method method = pickMethod(iParams);
+
+    try {
```
luigidellaquila: Shouldn't we add a null check here...?

luigidellaquila:
Hi Fabrizio,

I just reviewed the PR and I'd say it's OK. I only added a comment about a null check we could add, just to make it perfect. WDYT?

Thanks

Luigi

fabriziofortino:
@luigidellaquila PR amended to address your feedback

luigidellaquila:
Great, thank you very much, merging...

luigidellaquila:
Merged, thank you very much @fabriziofortino !

Luigi

lvca:
One of the best PR ever received! Good job @fabriziofortino !

luigidellaquila:
I was wondering if we can use this mechanism as a simplified way for end users to create custom functions.
Any Java developer can write a class with a static method and create a Jar file. A small additional configuration (perhaps in the orientdb-server-config.xml) to define the function prefix and it's done!
WDYT?

Thanks

Luigi

fabriziofortino:
I was actually thinking to use the same mechanism to bind the string similarity and distance algorithms provided by https://github.com/tdebatty/java-string-similarity/ (It's a very small library ... around 45KB). I think these functions can be very useful in a database. Oracle has something similar too (https://docs.oracle.com/cd/E18283_01/appdev.112/e16760/u_match.htm).

With that said I could try to implement the generic mechanism @luigidellaquila proposed and use java-string-similarity to prove it works. What about creating a separate github repo that contains all the **EXTRA** functions?

@luigidellaquila @lvca WDYT?

luigidellaquila:
Hi @fabriziofortino 

For me the idea of having a separate project for extra functions is very good, it will make community contributions much easier.
@lvca wdyt?

Thanks

Luigi

lvca:
I like it. The best would be creating multiple directories by type, so when we compile we generate multiple jars, and the user can pick whatever he wants. WDYT?

fabriziofortino:
I will create a separate PR to allow external custom functions as @luigidellaquila suggested.

After this we can work on the extra function repo.
Do you want to create it under the `orientechnologies` github organization?

lvca:
Just created: https://github.com/orientechnologies/extra-functions

PR #7409, opened 2017-05-11
Case sensitive file names

laa:
A new version of file registry in write cache was implemented.
The main difference is that it contains now two different file names.
The first name is case sensitive and is used to refer files inside of storage second name is a name of the file used by a file system.  Name of the file is used by a file system may be and is different than the name of the file used by storage.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
@@ -67,15 +67,34 @@ import java.util.zip.CRC32;
  * @since 7/23/13
  */
 public class OWOWCache extends OAbstractWriteCache implements OWriteCache, OCachePointer.WritersListener {
-  // we add 8 bytes before and after cache pages to prevent word tearing in mt case.
-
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
```diff
@@ -1126,33 +1226,149 @@ private void initNameIdMapping() throws IOException, InterruptedException {
...
   }
 
-  private void readNameIdMap() throws IOException, InterruptedException {
-    nameIdMap = new ConcurrentHashMap<String, Integer>();
+  private void convertNameIdMapFromV1ToV2() throws IOException, InterruptedException {
+    final File nameIdMapHolderFileV2T = new File(storagePath, NAME_ID_MAP_V2_T);
+
+    if (nameIdMapHolderFileV2T.exists()) {
+      if (!nameIdMapHolderFileV2T.delete()) {
+        throw new OStorageException("Can not deleter temporary version of registry of storage files "
```
taburet: typo, "deleter"

laa: Fixe

PR #7415, opened 2017-05-12
(refactor) some opportunities to use multi-catch

rbonifacio:
We found some opportunities to use multi-catch in the source code. We changed the source code and assigned to the license statement--- the main problem with the previous pull request


rbonifacio:
Dear all, not sure whether or not this contribution will be accepted. Should I close it?

laa:
Hi @rbonifacio, in general, it is a good one. But we can not apply it at the current moment because it will kill our ability to merge more complex pull requests which are waiting for review now. We will merge it a bit later.

rbonifacio:
Thanks @Iaa. I will keep it opened.

PR #7431, opened 2017-05-18
custom sql functions server plugin

fabriziofortino:
* server plugin to easily register custom sql functions (configured in an external json file)
* custom ad hoc `OSQLFunctionFactory` to handle custom functions registration
* basic `custom-sql-functions.json` with java.lang.Math registration (not hard coded anymore)
* aligned tests

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/functions/OCustomSQLFunctionFactory.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/functions/OCustomSQLFunctionFactory.java
@@ -1,0 +1,77 @@
+package com.orientechnologies.orient.core.sql.functions;
+
+import com.orientechnologies.common.exception.OException;
+import com.orientechnologies.common.log.OLogManager;
+import com.orientechnologies.orient.core.exception.OCommandExecutionException;
+import com.orientechnologies.orient.core.sql.functions.misc.OSQLStaticReflectiveFunction;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Factory for custom SQL functions.
+ *
+ * @author Fabrizio Fortino
+ */
+public class OCustomSQLFunctionFactory implements OSQLFunctionFactory {
+  private static final Map<String, Object> FUNCTIONS = new HashMap<>();
+
+  public static void register(final String prefix, final Class<?> clazz) {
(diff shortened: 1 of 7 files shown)


fabriziofortino:
@lvca @luigidellaquila here is the initial work to easily add custom sql functions (based on our last conversation). Please review.

lvca:
Looks very good to me, thanks! Waiting for @luigidellaquila on this.

luigidellaquila:
Thank you @fabriziofortino 

The PR seems to be technically perfect. 
My only concern is about the registration of math functions with the same mechanism, this will lead to a situation where a stand-alone server will have math functions by default, while in a plocal instance you will have to manually register them. This could be a bit confusing for the end user. 

@lvca @fabriziofortino wdyt? Is it worth manually register default (eg. math) functions and leave the user the ability to register only their custom functions with this mechanism?

Thanks

Luigi

fabriziofortino:
I understand your point @luigidellaquila 

I had the same doubt while I was implementing it. I have decided to put `math` registration in the configuration file because I don't think those kind of functions are going to be used by the majority of OrientDB users.

From a workload perspective registering `java.lang.Math` means adding 73 new functions that need to be loaded in memory. I know the footprint for that functions is irrelevant but from a user point of view it's probably better to not register them if there is no need.

My original idea was to disable the custom functions by default in OrientDB server (there is an `enabled` flag in the config file for that).

To enable them users need to change the flag from `false` to `true` on the server or to manually register them (https://github.com/orientechnologies/orientdb/pull/7431/files#diff-4a8735659709840d3760855b10bd065dR22) for plocal instances.

Let me know what do you think. I am OK registering the math functions by default (it's a 2 minutes change).

luigidellaquila:
@fabriziofortino it makes sense, IMHO the only thing that really matters is that by default remote and embedded behave the same way. Let's wait for @lvca 's opinion

Thanks

Luigi

lvca:
I think having math functions is a huge bonus for everybody, once they are there. So I'd like to see all of them registered by default and in v3.0 docs, and leave the component open to 3rd party implementations. WDYT?

luigidellaquila:
Ok, it makes sense to me. Again, my main concern is behavior consistency, but I also agree with Luca that math functions are a must-have (shame on us, we had to do it before ;-) )

I'm merging the pull-request, then I'll do a little hack to manually register Math.

Thanks

Luigi

lvca:
Thanks @fabriziofortino, this was a very nice contribution!

PR #7486, opened 2017-06-13
file size tracking backported from 3.0, checksum verification re-enabled

taburet:


core/src/main/java/com/orientechnologies/orient/core/storage/fs/OFileClassic.java
```diff
@@ -572,38 +546,49 @@ private void setVersion(int version) throws IOException {
...
-   * 
+   *
    * @see com.orientechnologies.orient.core.storage.fs.OFileAAA#open()
    */
   public void open() {
     acquireWriteLock();
     try {
       if (!osFile.exists())
-        throw new OIOException("File " + osFile.getPath() + " was not found");
+        throw new FileNotFoundException("File: " + osFile);
```
laa: Cool !!

core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
```diff
@@ -667,60 +659,166 @@ public Future store(final long fileId, final long pageIndex, final OCachePointer
...
-              verifyChecksums);
+        OCachePointer pagePointers[];
+        try {
+          //load requested page and preload requested amount of pages
+          pagePointers = loadFileContent(intId, startPageIndex, pageCount, verifyChecksums);
 
-          if (pagePointers.length == 0)
-            return pagePointers;
+          if (pagePointers != null) {
+            if (pagePointers.length == 0)
```
taburet: Seems like that is impossible, but that is how it's implemented in 3.0. Should I change it?

core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
```diff
@@ -1280,97 +1378,96 @@ private void removeCachedPages(int fileId) {
...
+              final ByteBuffer buffer = bufferPool.acquireDirect(false);
+              assert buffer.position() == 0;
 
-            if (verifyChecksums && (checksumMode == OChecksumMode.StoreAndVerify || checksumMode == OChecksumMode.StoreAndThrow))
-              verifyChecksum(buffer, fileId, startPageIndex, null);
+              fileClassic.read(firstPageStartPosition, buffer, false);
 
-            buffer.position(0);
+              if (verifyChecksums && (checksumMode == OChecksumMode.StoreAndVerify || checksumMode == OChecksumMode.StoreAndThrow))
+                verifyChecksum(buffer, fileId, startPageIndex, null);
```
taburet: Actually, it's released in `verifyChecksum`, if `buffersToRelease` parameter value is `null` then buffer pointed by `buffer` parameter is released. It's an optimization, so we don't allocate additional array if we read only one page, which is the most common case.

core/src/main/java/com/orientechnologies/orient/core/storage/fs/OFileClassic.java
```diff
@@ -136,16 +133,15 @@ public void read(long offset, byte[] iData, int iLength, int iArrayOffset) throw
...
     int attempts = 0;
 
     while (true) {
       try {
         acquireReadLock();
         try {
           offset = checkRegions(offset, buffer.limit());
-
-          readByteBuffer(buffer, channel, offset);
+          readByteBuffer(buffer, channel, offset, throwOnEof);
```
taburet: This whole retry behaviour seems strange to me :) What is the purpose of it?

core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
laa: Yep, I know that is implemented in this way )) but IMHO we should not keep dead code it makes understanding of code worse than it could be.

core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
laa: aha cool

core/src/main/java/com/orientechnologies/orient/core/storage/fs/OFileClassic.java
laa: It was done in case OS automatically closed file descriptor. We just left it for the safety reasons.

core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
taburet: fixed, assertion added

core/src/main/java/com/orientechnologies/orient/core/storage/fs/OFileClassic.java
taburet: fixed, EOF is considered unrecoverable by `reopenFile`

PR #7520, opened 2017-06-30
Improve complexity of defragmentation of pages from O(n^2) to O(n)

taburet:


core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/paginated/OClusterPage.java
```diff
@@ -481,39 +476,70 @@ private boolean isPositionInsideInterval(final int recordPosition) {
...
+
+      if (entryKind == ENTRY_KIND_DATA && shift > 0) {
+        final int positionIndex = getIntValue(position + OIntegerSerializer.INT_SIZE);
+        setIntValue(PAGE_INDEXES_OFFSET + INDEX_ITEM_SIZE * positionIndex, position + shift);
+
+        lastDataPosition = position;
+        mergedDataSize += size;
+      }
+
+      if (mergedDataSize > 0 && (entryKind == ENTRY_KIND_HOLE || i == 0)) {
```
laa: Please add a comment for this branch because it is not clear at first glance why we need to accumulate size of the data.

taburet: fixed

PR #7526, opened 2017-07-03
Add "overwrite" record conflict strategy

luigidellaquila:
This conflict strategy just overwrites the record with latest value provided regardless of the version number. The version number is updated at the max value between the newly provided and the existing

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/conflict/OOverwriteConflictStrategy.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/conflict/OOverwriteConflictStrategy.java
@@ -1,0 +1,48 @@
+/*
+ *
+ *  *  Copyright 2014 Orient Technologies LTD (info(at)orientechnologies.com)
+ *  *
+ *  *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  *  you may not use this file except in compliance with the License.
+ *  *  You may obtain a copy of the License at
+ *  *
+ *  *       http://www.apache.org/licenses/LICENSE-2.0
+ *  *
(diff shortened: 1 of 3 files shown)


luigidellaquila:
@lvca my only concern here is on how it works in a distributed environment.
Are the changes on a single record supposed to arrive in the same order on all the nodes?
Is the conflict resolved separately on each node?

Thanks

Luigi

lvca:
Changes on the same record, on the same server, will arrive with the original order to all the other servers. And each node executes the conflict strategy.

In case 1 edge is added on 2 servers at the same time, since on distributed there is not TreeRidBag, one of this edge would be lost. It's much better to use the existent MERGE strategy, where at least, vertices would be merged.

So I think this could be very dangerous for the end user.

luigidellaquila:
mmm... good point...

luigidellaquila:
Ok, we have two alternatives here IMHO:

- drop the PR

- allow it only on non-graph clusters, but it will imply some additional checks on the schema manipulation AND on the ALTER CLUSTER logics

WDYT?

Luigi

dalegarv:
Hi Luigi,

  If making it a built in strategy increases the risk of a customer having bad data, then I have no problem with the PR being dropped.  In our use case we’re only going to use the strategy on a specific generic class used for health checks, and it’s fine if we need to create a custom strategy to do that.  Thank you for the time you spent considering this and creating the PR.

S. Dale Garvey
Principal Architect - Applications
Portal Integration Development
770.360.2071 (External)
445-2071 (Internal)

luigidellaquila:
Ok, I'm closing this PR then

Thanks

Luigi

PR #7542, opened 2017-07-11
Add new interface ODatabaseSession that describe the correct name for…

tglman:
… the session


laa:
So now that is a session, cool!

PR #7563, opened 2017-07-17
Fixed database backup directory handling.

ataylor284:
When orientdb in distributed mode transfers a database from another node, it backs up the old copy. If the backup/databases exists, it is deleted. In the original code, the directory is not recreated after clearing, causing an error on when the old database is backed up with renameTo. In addition, the it was erroneously testing the condition on backupdirectory without the database directory prepended.

This fix correctly removes the old backup if it exists, and recreates the backup/databases if needed in all cases. Additionally for better diagnositics, replaced File.renameTo with Files.move to generate exception on failure instead of testing boolean return value.


lvca:
Managed in pretty much the same way: https://github.com/orientechnologies/orientdb/commit/84594bac4992b671acc546e096662ed93e2fe3bd.

Issue #7539, opened 2017-07-10
Error on restore of cluster from file during replace cluster


tglman:
## Steps to reproduce  
Run distributed test TestShardingManualSync

taburet:
This specific issue if fixed. But in general cluster replication is broken, the only file being replicated is *.pcl; *.cpm and related index files are ignored. See OSyncClusterTask#execute for details.

tglman:
hi,

this is all fixed now. closing.

PR #7570, opened 2017-07-19
Error on restore of cluster from file during replace cluster #7539

taburet:


taburet:
Test case added.

laa:
@taburet thank you very much, good as usual

Issue #5931, opened 2016-04-04
Use an atomic rename for the backup file, instead of just opening and writing


obi1kenobi:
I'm running OrientDB 2.1.9 configured with the backup plugin. I noticed that when backing up, the database simply opens its intended backup file and keeps writing data to it until finished. This is bad because it would leave the backup file corrupted in the event of a database crash mid-backup, but I wouldn't be able to notice since the file exists in its expected location.

A suggested solution is to do what other databases do -- write the backup to a separate file, say `<intended_backup_filename>.tmp` and when done writing the file, perform an _atomic rename_ operation to change the name of the file from the `.tmp` one to the final filename.

This is a non-breaking change, since externally the behavior will be the same -- the backup file will have the same name and be in the same location. However, we will be guaranteed that any `.zip` files in the backup directory are valid backups and not corrupted.

obi1kenobi:
Is there a planned milestone for this?

laa:
That is enhancement so I put it in 3.0 version

taburet:
Implemented, closing.

obi1kenobi:
The code in the linked diff doesn't actually perform an atomic rename operation -- the `ATOMIC_MOVE` option isn't set in the `Files.move()` method: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#move(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...)

taburet:
@obi1kenobi thank you for the observation, reopening. Usually, if the file system supports atomic renames, regular renames are also atomic. Anyway, for the sake of strictness I will pass `ATOMIC_MOVE` and then fallback to regular move if the move can't be done atomically. WDYT?

taburet:
`ATOMIC_MOVE` is now used to move files, closing.

obi1kenobi:
Looks great, thank you! It might also be good to log an error to inform the user that atomic moves are not supported by their system -- I definitely would like to know if that were the case for my database.

taburet:
@obi1kenobi ok, I will add a warning to the log.

taburet:
Warning added: https://github.com/orientechnologies/orientdb/commit/2c5c1e3504dce38721741792dff8fc39d61a2ae1

obi1kenobi:
Looks great, thank you!

PR #7590, opened 2017-07-27
Use an atomic rename for the backup file, instead of just opening and writing #5931

taburet:


core/src/main/java/com/orientechnologies/orient/core/db/tool/ODatabaseExport.java
```diff
@@ -69,11 +73,22 @@ public ODatabaseExport(final ODatabaseDocumentInternal iDatabase, final String i
...
     if (f.exists())
+      //noinspection ResultOfMethodCallIgnored
       f.delete();
 
-    final GZIPOutputStream gzipOS = new GZIPOutputStream(new FileOutputStream(fileName), compressionBuffer) {
+    this.tempFileName = fileName + ".tmp";
+    final File tempFile = new File(tempFileName);
+    if (tempFile.getParentFile() != null)
+      //noinspection ResultOfMethodCallIgnored
+      tempFile.getParentFile().mkdirs();
```
laa: That is 3.0 version which supports all features of Java 8 API. Could you use new file API to create and remove directories and files? This will allow us to produce meaningful exceptions if the creation of directory is failed for some reason.

core/src/main/java/com/orientechnologies/orient/core/db/tool/ODatabaseExport.java
```diff
@@ -69,11 +73,22 @@ public ODatabaseExport(final ODatabaseDocumentInternal iDatabase, final String i
...
+    if (tempFile.exists())
+      //noinspection ResultOfMethodCallIgnored
+      tempFile.delete();
```
laa: The same as above

core/src/main/java/com/orientechnologies/orient/core/db/tool/ODatabaseExport.java
```diff
@@ -69,11 +73,22 @@ public ODatabaseExport(final ODatabaseDocumentInternal iDatabase, final String i
     }
     final File f = new File(fileName);
     if (f.getParentFile() != null)
+      //noinspection ResultOfMethodCallIgnored
       f.getParentFile().mkdirs();
     if (f.exists())
+      //noinspection ResultOfMethodCallIgnored
       f.delete();
```
laa: I suppose would be good to put warning here,  I do not think that tmp file may contain important information but would be better for user if he would be aware if something is removed.

server/src/main/java/com/orientechnologies/orient/server/handler/OAutomaticBackup.java
```diff
@@ -346,7 +348,9 @@ protected void incrementalBackupDatabase(final String dbURL, String iPath, final
   protected void fullBackupDatabase(final String dbURL, final String iPath, final ODatabaseDocumentInternal db) throws IOException {
     OLogManager.instance().info(this, "AutomaticBackup: executing full backup of database '%s' to %s", dbURL, iPath);
 
-    final FileOutputStream fileOutputStream = new FileOutputStream(iPath);
+    final String tempPath = iPath + ".tmp";
+
+    final FileOutputStream fileOutputStream = new FileOutputStream(tempPath);
```
laa: Why do we not check here the presence of temp file on the file system?

core/src/main/java/com/orientechnologies/orient/core/db/tool/ODatabaseExport.java
taburet: fixed

core/src/main/java/com/orientechnologies/orient/core/db/tool/ODatabaseExport.java
taburet: fixed

core/src/main/java/com/orientechnologies/orient/core/db/tool/ODatabaseExport.java
taburet: fixed

server/src/main/java/com/orientechnologies/orient/server/handler/OAutomaticBackup.java
taburet: fixed, hope it will not conflict with the old behaviour

PR #7714, opened 2017-09-04
Use soft references in SQL SELECT result set

luigidellaquila:


core/src/main/java/com/orientechnologies/orient/core/sql/SoftQueryResultList.java
```diff
@@ -0,0 +1,361 @@
...
+  @Override
+  public boolean isEmpty() {
+    return buffer.isEmpty();
+  }
+
+  @Override
+  public boolean contains(Object o) {
+    for (SoftReference<T> ref : buffer) {
+      T item = getItem(ref);
+      if (item != null) {
```
lvca: At this point item is always == null

core/src/main/java/com/orientechnologies/orient/core/sql/SoftQueryResultList.java
```diff
@@ -0,0 +1,361 @@
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public boolean containsAll(Collection<?> c) {
+    for (Object o : c) {
+      if (!contains(c)) {
```
lvca: Bug, it should be `contains(o)`

laa:
Hi @luigidellaquila , honestly speaking I suppose we can improve your solution a lot by taking into account some assumptions. 

1.  Most of the queries may be satisfied by fetching not more than 100 items.
2. Soft referneces provide overhead on GC because it has to resolve how strong is patch to each object during collection.
3. We should detect "OOM" situation as early as possible.

So I propose you:

1. Make a mix of collections of soft referneces. We can collect first 100 items using strong referecnes and the rest using soft references it will decrease overhad on GC in dunno 80-85% of cases if not in 90%, also from users perspective it completely normal if it loads more than 100 items server becomes a litle slower.
Also even if size of each record will consume 1MB total (which I suppose quite ridicilous) total memory consumiton will be 100MB which is OK for me. We also can reduce this size to 10 items if you find it more reasonable.
2. To detect OOM situation as early as possible I propose to pass reference queue to each reference. And before each operation to check that queue is empty. It will save time if OOM happend only few last referneces will be removed and we and user need to wait a lot of time to find it. Also reference queue check may be used when you add itmes to the collection. Because during this operaion you obviously do not check other references on "nullity" but that is plase where 99% of OOM situations happen. So in such case users will not wait till 1 000 000 records will load and started to be processed before query will be aborted. Reference queue check is very light operation.

WDYT ?

core/src/main/java/com/orientechnologies/orient/core/sql/SoftQueryResultList.java
luigidellaquila: Good point, fixing

luigidellaquila:
Hi @laa 

Considering that it's a patch for a system that will become legacy in a few months, I'd avoid the complexity of managing two different lists with strong and soft references. 
Soft references are an overhead for the GC of course, but the overhead is significant when you have thousands of them, while if you have a few tens it should be acceptable IMHO (but please correct me if I'm wrong, I lack some low level details about this)

About 2, I completely agree, it should be easy to do, new PR coming soon

Thanks

Luigi

laa:
@luigidellaquila I suppose overhead of soft references is a subjective topic, probably you are right. 
But we need to be sure at 100%. Because we are professionals )) I suppose we need to make a benchmark to see a real difference. So I propose to add small benchmark which will measure system performance and gc overhead for few types of queries. WDYT ?

laa:
@luigidellaquila I suppose we can close this pull request.

PR #7715, opened 2017-09-05
Rid prealloc

tglman:
Add Functionality to pre-alloc record id in a single independent atomic operation

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
@@ -1420,6 +1420,93 @@ public abstract class OAbstractPaginatedStorage extends OStorageAbstract
     }
   }
 
+  public void preallocateRids(final OTransaction clientTx) {
+
+    try {
+      checkOpenness();
+      checkLowDiskSpaceRequestsAndBackgroundDataFlushExceptions();
+
+      final ODatabaseDocumentInternal databaseRecord = (ODatabaseDocumentInternal) clientTx.getDatabase();
+
+      databaseRecord.getMetadata().makeThreadLocalSchemaSnapshot();
+
+      @SuppressWarnings("unchecked")
+      final Iterable<ORecordOperation> entries = (Iterable<ORecordOperation>) clientTx.getAllRecordEntries();
+      final TreeMap<Integer, OCluster> clustersToLock = new TreeMap<>();
+      final Map<ORecordOperation, Integer> clusterOverrides = new IdentityHashMap<>();
+
+      final Set<ORecordOperation> newRecords = new TreeSet<>(COMMIT_RECORD_OPERATION_COMPARATOR);
+
+      for (ORecordOperation txEntry : entries) {
+
+        if (txEntry.type == ORecordOperation.CREATED) {
+          newRecords.add(txEntry);
+
+          final ORecord record = txEntry.getRecord();
+          final ORID rid = record.getIdentity();
+
+          int clusterId = rid.getClusterId();
+
+          clustersToLock.put(clusterId, getClusterById(clusterId));
+        }
+      }
+
+      final List<ORecordOperation> result = new ArrayList<ORecordOperation>();
+      stateLock.acquireReadLock();
+      try {
+        try {
+          try {
+
+            checkOpenness();
+
+            makeStorageDirty();
+
+            lockClusters(clustersToLock);
+
+            Map<ORecordOperation, OPhysicalPosition> positions = new IdentityHashMap<ORecordOperation, OPhysicalPosition>();
+            for (ORecordOperation txEntry : newRecords) {
+              ORecord rec = txEntry.getRecord();
+
+              if (rec.isDirty()) {
+                ORecordId rid = (ORecordId) rec.getIdentity().copy();
+                ORecordId oldRID = rid.copy();
+
+                final Integer clusterOverride = clusterOverrides.get(txEntry);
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1391,7 +1391,79 @@ public int getClusterIdByName(final String clusterName) {
...
+      throw logAndPrepareForRethrow(ee);
+    } catch (Throwable t) {
+      throw logAndPrepareForRethrow(t);
+    }
+  }
   public List<ORecordOperation> commit(final OTransaction clientTx, Runnable callback) {
+    return  commit(clientTx,callback,false);
+  }
+
+  public List<ORecordOperation> commitPreAllocated(final OTransaction clientTx, Runnable callback) {
```
taburet: javadoc, return value never used.

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1391,7 +1391,79 @@ public int getClusterIdByName(final String clusterName) {
...
+    return commit(clientTx,callback,true);
+  }
+
+  private List<ORecordOperation> commit(final OTransaction clientTx, Runnable callback,boolean allocated) {
```
taburet: `callback` never used.

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1466,7 +1538,9 @@ public int getClusterIdByName(final String clusterName) {
             for (ORecordOperation txEntry : newRecords) {
               ORecord rec = txEntry.getRecord();
 
-              if (rec.isDirty()) {
+              if (allocated && rec.getIdentity().isPersistent()) {
+                positions.put(txEntry, new OPhysicalPosition(rec.getIdentity().getClusterPosition()));
+              } else if (rec.isDirty() && !rec.getIdentity().isPersistent()) {
```
taburet: If the `rid` is persisted, we get some strange behavior here. The record will not receive its entry in `positions`, its physical position will be `null` in `commitEntry` call, that will trigger record update logic instead of create. Not sure is it intended behavior or not. If not, it's better to add an assertion here to postulate the expected state and behavior.

laa:
@tglman and a small note from me, could you merge latest development version.

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1391,7 +1391,79 @@ public int getClusterIdByName(final String clusterName) {
...
+      try {
+
+        checkOpenness();
+
+        makeStorageDirty();
+        atomicOperationsManager.startAtomicOperation((String) null, true);
+        try {
+          lockClusters(clustersToLock);
+
+          Map<ORecordOperation, OPhysicalPosition> positions = new IdentityHashMap<ORecordOperation, OPhysicalPosition>();
```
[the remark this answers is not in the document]
tglman: right, left over of copy and past, I'll get rid of it

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
tglman: Callback is there since long time ago because it was used for indexes, it's not needed anymore anyway it need a bit bigger refactor to remove it, so I will keep as it is in this PR and will be removed in a future commit.

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
tglman: This logic is the complex part, it has three behaviours, when the record is allocated ahead [1], when is not allocated[2], when the id is generated by another node[3]. 

1) when the record is allocated ahead we should put the entry in positions and that will trigger an update as you noticed, this update will change the record from allocated to persistent.

2) when the record is not allocated and i've the rid not persistent, this happen in the normal tx flow.

3) The rid is persistent but not allocated this will force a creation of a rid. this happen when a node apply a transaction from another node, the record created by the other node don't exist but have valid rid.

So from my concern here all the case are valid, is not too clear but is the best that we can get on the current storage architecture.

laa: @tglman  if you think that all cases are valid, could you add the same message which you put there as comment in code, so other developers will not scratch their head thinking what is going on ))

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
taburet: Sure, I'm aware about that callbacks on public interfacing methods, but still there is no point in having unused parameters in private methods.

tglman: I yep, i can get rid of it :)

taburet: Thanks ;)

PR #7759, opened 2017-09-20
Comments regarding tests problems

taburet:
To run individual test with the leak detector enabled from IDE, provide `-Dmemory.directMemory.trackMode=true -Djava.util.logging.manager=com.orientechnologies.common.log.OLogManager$DebugLogManager` options to the JVM.

### DIFF
--- a/server/src/main/java/com/orientechnologies/orient/server/OServer.java
+++ b/server/src/main/java/com/orientechnologies/orient/server/OServer.java
@@ -453,6 +453,11 @@ public class OServer {
     try {
       boolean res = deinit();
 
+      // Seems like the code below should be executed before deinit() call, otherwise databases are closed after the storages and
+      // legitimate leaks reported.
+      //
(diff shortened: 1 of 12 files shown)


server/src/main/java/com/orientechnologies/orient/server/OServer.java
```diff
@@ -533,6 +538,14 @@ protected boolean deinit() {
           OLogManager.instance().info(this, "Shutting down databases:");
           Orient.instance().shutdown();
         } catch (Throwable e) {
+          // We are eating all exceptions here. As a result, all state verification logic is unable to fail the tests. At least
+          // AssertionError-s should be passed through, or even better, all exceptions if we are running in a testing env
+          // (check for -ea?).
+          //
+          // Another thing to think about, in-process server tests will fail with pass-through exceptions, but tests running a
```
laa: If I remember correctly now all tests use in-process servers, so we can skip this problem for now at least.

PR #7783, opened 2017-10-02
add logic for preallocate from valid rids

tglman:
Add some logic to allow to pre-allocate rids from a already valid rid, this is needed for distributed when a transaction is arriving from another node.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
@@ -1497,15 +1497,23 @@ public abstract class OAbstractPaginatedStorage extends OStorageAbstract
 
           for (ORecordOperation txEntry : newRecords) {
             ORecord rec = txEntry.getRecord();
-
-            if (rec.isDirty()) {
-              ORecordId rid = (ORecordId) rec.getIdentity().copy();
-              ORecordId oldRID = rid.copy();
+            if (!rec.getIdentity().isPersistent()) {
+              if (rec.isDirty()) {
+                ORecordId rid = (ORecordId) rec.getIdentity().copy();
+                ORecordId oldRID = rid.copy();
+                final OCluster cluster = getClusterById(rid.getClusterId());
+                OPhysicalPosition ppos = cluster.allocatePosition(ORecordInternal.getRecordType(rec));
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1497,15 +1497,23 @@ public void preallocateRids(final OTransaction clientTx) {
...
-            if (rec.isDirty()) {
-              ORecordId rid = (ORecordId) rec.getIdentity().copy();
-              ORecordId oldRID = rid.copy();
+            if (!rec.getIdentity().isPersistent()) {
+              if (rec.isDirty()) {
+                ORecordId rid = (ORecordId) rec.getIdentity().copy();
+                ORecordId oldRID = rid.copy();
+                final OCluster cluster = getClusterById(rid.getClusterId());
+                OPhysicalPosition ppos = cluster.allocatePosition(ORecordInternal.getRecordType(rec));
+                rid.setClusterId(cluster.getId());
```
laa: Why you set cluster id if that already exists according to line 1504?

core/src/test/java/com/orientechnologies/orient/core/tx/TransactionRidAllocationTest.java
```diff
@@ -69,6 +70,43 @@ public void testAllocationCommit() {
   }
 
   @Test
+  public void testMultipleDbAllocationAndCommit() {
```
laa: The test is good but do not cover all cases, I mean there is no check that exception is thrown if rids do not match

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1497,15 +1497,23 @@ public void preallocateRids(final OTransaction clientTx) {
...
+                rid.setClusterPosition(ppos.clusterPosition);
+                clientTx.updateIdentityAfterCommit(oldRID, rid);
+              }
+            } else {
+              ORecordId rid = (ORecordId) rec.getIdentity();
```
laa: Could you add a comment which explains why we need the second branch, otherwise someone can remove it because will treat it as not needed?

PR #7799, opened 2017-10-09
Unification transaction api

tglman:
Introduced new OTransactionInternal interface to be used by the commit operations with all methods need by it, made both OptimisticTransaction and MicroTransaction inplement the interface, the commit methods are still two different implementation but can easily be merged due to the same behaviour, the merge will come in a not too far future.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/tx/OTransactionInternal.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/tx/OTransactionInternal.java
@@ -1,30 +1,27 @@
-/*
- *
- *  *  Copyright 2010-2016 OrientDB LTD (http://orientdb.com)
- *  *
- *  *  Licensed under the Apache License, Version 2.0 (the "License");
- *  *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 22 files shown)


core/src/main/java/com/orientechnologies/orient/core/tx/OTransactionInternal.java
```diff
@@ -1,30 +1,27 @@
...
-  }
+import com.orientechnologies.orient.core.db.ODatabaseDocumentInternal;
+import com.orientechnologies.orient.core.db.record.ORecordOperation;
+import com.orientechnologies.orient.core.id.ORID;
+import com.orientechnologies.orient.core.storage.OBasicTransaction;
+
+import java.util.Collection;
+import java.util.Map;
+
+public interface OTransactionInternal extends OBasicTransaction {
```
laa: 1. Copyright is absent.
2. JavaDoc explaining the meaning of interface is absent. 
3. JavaDocs for methods are absent. Please try to provide meaningful JavaDocs, not formal ones.

PR #7865, opened 2017-11-02
Add query text in thread name

luigidellaquila:


core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -3291,7 +3291,14 @@ public Object command(final OCommandRequestText iCommand) {
           executor.setProgressListener(iCommand.getProgressListener());
           executor.parse(iCommand);
 
-          return executeCommand(iCommand, executor);
+          String currentThreadName = Thread.currentThread().getName();
+          try {
+            Thread.currentThread().setName(currentThreadName + " <command>" + iCommand + "</command>");
+            return executeCommand(iCommand, executor);
+          } finally {
+            Thread.currentThread().setName(String.valueOf(currentThreadName));
```
laa: Why String.valueOf, the parameter is already String, is not it?

luigidellaquila: is null value accepted as a thread name...? In case we can just pass the value as it is

luigidellaquila:
Yes, good point, in a query we can invoke another query... Let me fix it

core/src/main/java/com/orientechnologies/orient/core/sql/query/OSQLSynchQuery.java
```diff
@@ -78,8 +78,13 @@ public void end() {
   public List<T> run(final Object... iArgs) {
     String currentThreadName = Thread.currentThread().getName();
     try {
-      Thread.currentThread().setName(currentThreadName + " <query>" + this.getText() + "</query>");
-      
+      if (currentThreadName == null || !(currentThreadName.contains("<query>") || currentThreadName.contains("<command>"))) {
+        try {
+          Thread.currentThread().setName(currentThreadName + " <query>" + this.getText() + "</query>");
+        } catch (SecurityException x) {
+        }
```
laa: in such case you increase Java warnings, could you add the comment inside catch?

luigidellaquila: :D :D no problem, doing it now

PR #7883, opened 2017-11-14
Use soft references in SQL query engine to avoid OOM

luigidellaquila:


core/src/main/java/com/orientechnologies/orient/core/config/OGlobalConfiguration.java
```diff
@@ -575,6 +575,13 @@ public void change(final Object iCurrentValue, final Object iNewValue) {
       Integer.class, 500),
 
   // QUERY
+  QUERY_USE_SOFT_REFENCES_IN_RESULT_SET("query.useSofReferencesInResultSet",
```
luigidellaquila: There is a typo here!!! `useSofReferences` should be `useSoftReferences`

tglman:
Hi,

I'm not yet 100% convinced by this solution, for have this solution working i saw that we have to use `Xms == Xmx` that may work when we control the setup(server dist) but it won't work when we don't control the setup(embedding), as well for the case of embedding OOM may come from a different source so soft reference will solve only half of the problem, and probably introducing some not deterministic behaviour(really hard to reproduce in case of error), like a result set disappearing while  iterating it, due to some other part of an application being overloaded(by design or mistake).

Regards

laa:
@tglman you are right that is in case of xms=xmx because our commercial server users suffer from this problem and we have several issues in a critical state where they ping us to solve it. This solution will solve all those issues.

laa:
Seems like we resolved all questions with @tglman .

tglman:
Hi,
From a quick private chat: if Xms != Xml the soft reference check is turned off, and if is on and is triggered a free of a soft reference the query get killed without not deterministic behaviour, this is good enough for me now.

core/src/main/java/com/orientechnologies/orient/core/engine/OMemoryAndLocalPaginatedEnginesInitializer.java
```diff
@@ -73,6 +84,75 @@ private void configureDefaults() {
...
+        OGlobalConfiguration.QUERY_USE_SOFT_REFENCES_IN_RESULT_SET.setValue(false);
+      }
+    }
+  }
+
+  private long extractMemoryLimitInBytes(String limit) {
+    final Pattern pattern = Pattern.compile("((-Xms)|(-Xmx))(\\d+)([gGmMkK]?)");
+    final Matcher matcher = pattern.matcher(limit);
+    if (!matcher.find()) {
+      throw new IllegalArgumentException("Invalid value of memory limit was provided '" + limit + "'");
```
luigidellaquila: why...? There is really no need to throw an exception if you don't explicitly define the memory limit...

core/src/main/java/com/orientechnologies/orient/core/engine/OMemoryAndLocalPaginatedEnginesInitializer.java
```diff
@@ -73,6 +84,75 @@ private void configureDefaults() {
...
+    dimension = dimension.toLowerCase(Locale.ENGLISH);
+
+    if (dimension.equals("g")) {
+      bytes = bytes * 1024 * 1024 * 1024;
+    } else if (dimension.equals("m")) {
+      bytes = bytes * 1024 * 1024;
+    } else if (dimension.equals("k")) {
+      bytes = bytes * 1024;
+    } else {
+      throw new IllegalArgumentException("Invalid dimension of memory limit + '" + dimension + "'");
```
luigidellaquila: same here, no need to throw an exception

core/src/main/java/com/orientechnologies/orient/core/engine/OMemoryAndLocalPaginatedEnginesInitializer.java
luigidellaquila: ah, ok, it's checked above...

core/src/main/java/com/orientechnologies/orient/core/sql/OSoftQueryResultTimSort.java
```diff
@@ -0,0 +1,911 @@
...
+          newSize = Math.min(newSize, aLength >>> 1);
+
+        //noinspection unchecked
+        tmp = new SoftReference[newSize];
+        tmpLength = newSize;
+      }
+      return tmp;
+    }
+  }
+}
```
luigidellaquila: Trusting you here...

PR #7907, opened 2017-11-30
fixing exception thrown from OSoftQueryResultList

wwannemacher:
In OSoftQueryResultList, the .toArray method is intentionally not implemented (https://github.com/orientechnologies/orientdb/blob/2.2.x/core/src/main/java/com/orientechnologies/orient/core/sql/OSoftQueryResultList.java#L176). However in OCommandExecutorSQLResultsetAbstract:718, java.util.ArrayList.addAll attempts to call .toArray, which results in an exception being thrown. This problem has made the current 2.2.x snapshot unusable for us because some basic select queries refuse to execute.


mrprescott:
Needless to say, we'd very much like to see this as part of 2.2.31. :)

luigidellaquila:
Hi @wwannemacher 

Thank you very much, I'll check it asap.
I'd say that if everything is ok it will be in 2.2.31

Thanks

Luigi

laa:
This solution is not enough, I will provide full solution today.

laa:
Fixed in https://github.com/orientechnologies/orientdb/commit/8b77b37a4fabbc0acddd840cf4d7459fb3d44150 .

PR #7917, opened 2017-12-06
Make shutdown timeout configurable.

ataylor284:
This PR simply adds a configuration property to control the distributed db thread timeouts on shutdown.


tglman:
Hi @ataylor284,

this looks good, I'll double check it and in case merge it.

Regards

PR #7922, opened 2017-12-11
introduced interface on top of the storage configuration to separate remote and embedded implementation

tglman:


tglman:
Well I pushed already the changes ( by mistake), anyway also if big, it was a trivial refactor, not needed a big review.

Regards

Issue #7950, opened 2017-12-19
Creation of abstract class extending multiple parent classes works only for one node properly


sa-ChristianNiessner:
## Expected behavior  
When creating an abstract class extending multiple parent classes it creates an abstract class on the primary node, but on other cluster members non-abstract classes with clusters are created.

## Actual behavior  
As described - It works for the node where the command is issued, but not for other members of the cluster.

## Steps to reproduce  
The initial creation was done via the API-Call
`OSchemaShared.createAbstractClass("myAbstractClass", "parentClass1", "parentClass2");`

This resulted in the following SQL statement transporting to other nodes:
```CREATE CLASS `myAbstractClass` EXTENDS `parentClass1`, `parentClass2` ABSTRACT```

When executing this sql command the database won't create an abstract class extending both given class but creates a non-abstract class only extending one parent class

sa-ChristianNiessner:
Closing issue as both PR's have been merged and this issue should be fixed now ;)

PR #7951, opened 2017-12-19
Bugfix in OCommandExecutorSQLCreateClass.java: handle ` escapings properly

sa-ChristianNiessner:
This is the Bugfix for Issue #7950 - Creation of abstract class extending multiple parent classes works only for one node properly

Please also backport to 2.2.x


luigidellaquila:
Merged, thanks!

Luigi

Issue #7968, opened 2017-12-27
How can I get graph-like results of traverse and shortestPath?


blueAnger:
### OS: <CentOS 7>  

I tried two kind of APIs, as mentioned in [issue #7957](https://github.com/orientechnologies/orientdb/issues/7957).

I also tried to use 'select $path', as mentioned in [issue #7962](https://github.com/orientechnologies/orientdb/issues/7962). 

None of those helps.

The result of traverse is a iterator or iterable. And the elements within are all vertices. The edge information are in_* and out_* properties. So I have to iterate and pick up all these information? If so,  then the orientdb traverse is not helpful because technically I am doing graph traverse by myself. If not, how can I get the graph-like result(for example, edge-set array)?

The result of the shortestPath() are a list of vertices ids. So I guess I should iterate and load them by using ODatabaseSession.load(id) ??? Same question, how do I get the edge-set array? Maybe I could go through properties of every document and filter the edges I want as well as the direction, but what's the point of the 'edgeClassName' and 'direction' parameters of shortestPath()？

In my opinion, the results are not friendly. I will appreciate that if you can show me  an example.

Thanks.

blueAnger:
# Update
I tried the following SQL in OrientDB 2.2.31:

```
select $path from (traverse inE(), outE(), outV(), inV() from (select from Person where name = 'Person2') maxdepth 10 strategy breadth_first)
```

it got something like:

```
{
"result": [
        {
            "@type": "d",
            "@rid": "#-2:0",
            "@version": 0,
            "$path": "(#22:0)"
        },
        {
            "@type": "d",
            "@rid": "#-2:1",
            "@version": 0,
            "$path": "(#22:0).outE[0](#28:0)"
        },
        {
            "@type": "d",
            "@rid": "#-2:2",
            "@version": 0,
            "$path": "(#22:0).outE[0](#28:0).inV(#20:0)"
        },
        {
            "@type": "d",
            "@rid": "#-2:3",
            "@version": 0,
            "$path": "(#22:0).outE[0](#28:0).inV(#20:0).outE[0](#25:0)"
        }
]
}
```

So, does it mean I should parse every single path string and load all vertices/edges through ODatabaseSession.load(rid)?

luigidellaquila:
Hi @blueAnger 

Please wait, we just realized that the $path option does not work properly in the RC1 and we are having a look at it. We also would like to do some changes to make it a bit easier to manipulate

Thanks

Luigi

[...]

PR #7983, opened 2018-01-04
feat(orientdb-core): add 'edge' option in shortestPath() function to return edge information

blueAnger:
since shortestPath() function didn't return edge information, as I mentioned in #7968 , I made a little change in the OSQLFunctionShortestPath.java to return edge information. 

Example: 

```
select shortestPath(#22:0, #24:0, null, null, {'maxDepth': 10, 'edge': true})
```

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/functions/graph/OSQLFunctionShortestPath.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/functions/graph/OSQLFunctionShortestPath.java
@@ -1,16 +1,16 @@
 package com.orientechnologies.orient.core.sql.functions.graph;
 
+import com.orientechnologies.common.collection.OMultiCollectionIterator;
 import com.orientechnologies.common.collection.OMultiValue;
+import com.orientechnologies.common.util.ORawPair;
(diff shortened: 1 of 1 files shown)


luigidellaquila:
Hi @blueAnger 

Thank you very much, the PR seems OK. 

Just a little change, if you don't mind: I see you added a dependency with `commons-lang3` just to use the `Pair` class; in the codebase we already have two implementations of Pair-like classes, that are `OPair` and `ORawPair`, the latter should be OK for this case, could you please review this PR using that implementation and remove the new dependency?

Thanks

Luigi

blueAnger:
Hi, @luigidellaquila 
Thanks for your notice. Actually I am not quite willing to add a new dependency too, since the ORawPair can do the same thing. I will change it tomorrow morning, for now my network transferring is super slow. -_-|||
Thanks.

blueAnger:
Sorry, I am not quite familiar with github forking workflow....I pull your changes from my local repository, make my own change, and push it to my github forked repository. But somehow my name is in front of your history commits. Is that right? If not, please tell me the right way.
Thanks.

luigidellaquila:
Merged, thank you very much!

Luigi

PR #7992, opened 2018-01-10
feat(orientdb-core): make outV(), inV() to support class filter

blueAnger:
I notice that outE(), inE() suport class filter, but outV(), inV() don't. 
However, there are some business scenarios in which we need vertex filter when traverse.

For example, two vertex classes: Person and Company, one edge class: invest. 
Person can invest company, company can invest company, and now I don't want get Person vertices. 
I know we can get there using other tricks, but I think it's not bad to make them support vertex class filter.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/record/OEdge.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/record/OEdge.java
@@ -19,12 +19,6 @@
  */
 package com.orientechnologies.orient.core.record;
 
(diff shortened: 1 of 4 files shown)


lvca:
Edges have only one out and one in. There are no labels.

luigidellaquila:
You can achieve the same with `out()[@class='Person']`

blueAnger:
OK, that's good to know.

PR #8032, opened 2018-01-29
Simple sbtrees size optimization

tglman:
Hi,

This is simple and i think we can merge it directly.

Regards


tglman:
wait it looks not what i expected, let me redo it ;)

PR #8033, opened 2018-01-29
simple optimization for avoid load of the sbtree root page

tglman:
Hi,

This I think we can merge it directly.

Regards.


laa:
yep you are right

PR #8083, opened 2018-02-08
First internal API for cluster paginated browse

tglman:
Hi,

This is the API i mentioned in the call, with the first implementation and test case, it's only internal and it will ever be internal so we are free to play around it as we want.

@luigidellaquila  have a quick look to the test case and see if is what you expect
@laa obviusly also an opinion on the style of the API is welcome and a review of the code as well, I expect to merge this, in the long run i expect to push a layer down the implementation so we can limit the "cpm" page loads to 1 per page not as today N+1 per page (counting that a cpm page has ~5k entries we could have 5k less cache hit for this specific case)

Bye


luigidellaquila:
Sure, will do

luigidellaquila:
It's OK for me

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1337,6 +1337,93 @@ public boolean isDeleted(ORID rid) {
...
+
+  private OBrowsePage nextPage(int clusterId, long lastPosition) {
+    try {
+      checkOpenness();
+      stateLock.acquireReadLock();
+      try {
+        checkOpenness();
+
+        final OCluster cluster = doGetAndCheckCluster(clusterId);
+        OPhysicalPosition[] nexts = cluster.higherPositions(new OPhysicalPosition(lastPosition));
```
laa: You use higherPositions call, it means that passed in position, will be omitted and also you use lastPosition's default value equal to 0. So during the first call of next method, record with position 0 will not be read. Could you change default value to -1 or explain why that is not the case

tglman: thanks i'll change that

tglman:
hi @laa,

Fixed also with the relative cleanup from codacy, actually the test case reported the problem but i assumed was an issue on the test case :), now should be fine.

Regards

PR #8100, opened 2018-02-13
Poc index value updater

tglman:
Hi @laa,

This is a pull request to replace the get + put  with a put for update, and as well to avoid to rewrite to reserialize the embedded data of the SBTreeBonsaiLocal if is the only thing that changed, i've to do an additional review of this as well so check this but do not merge it yet, from my benchmarks this can give a 10% speedup.


tglman:
Hi, 

I did some additional changes that should speedup a bit in case of tree bonsai, closing this and reopening another

PR #8101, opened 2018-02-13
Tree index code cleanup

tglman:
Hi @laa,

This is a start to replace boiler plate code for loadPage+releasePage with closable and try blocks, let me know if you like this approach and if can carry on doing more changes like this one, and even if this can be merged.

Anyway I haven't covered all the cases in this PR so if you like it may come more PR or we can do a new PR with all the changes.

Regards


tglman:
Hi @laa,

Do you agree on this ?

Regards

PR #8110, opened 2018-02-19
small optimization on file creation for avoid double flush

tglman:


core/src/main/java/com/orientechnologies/orient/core/storage/cache/local/OWOWCache.java
```diff
@@ -1693,9 +1689,10 @@ private void openFile(final OFileClassic fileClassic) {
 
   }
 
-  private void createFile(final OFileClassic fileClassic) throws IOException {
+  private void createFile(String fileName, int intId, final OFileClassic fileClassic) throws IOException {
     if (!fileClassic.exists()) {
       fileClassic.create();
+      writeNameIdEntry(new NameFileIdEntry(fileName, intId, fileClassic.getName()), false);
```
laa: Not bad but could you put this line below, so if we register file in write cache it will be already in the consistent state if the crash happens.

tglman:
Hi,
This seems that cannot be done, because the writeNameIdEntry needs to be done after the file is correctly flushed on the disc, so closing this.

Regards

PR #8114, opened 2018-02-20
Without class name

markodjurovic:
Avoided serialization of class name for all , except for embedded objects.
Serialization API slightly changed and now is up to user to provide info whether to serialize / deserialize class name or not.
Serialization / Deserialization of class name, for embedded documents, is turned on in serializeValue() and deserializeValue() methods.
New functionality is implemented in ORecordSerializerBinaryV1 class


markodjurovic:
Oboslete

PR #8125, opened 2018-02-27
Index put only and bonsai id caching

tglman:


tglman:
Hi Andrey,
This is almost the same PR of this: https://github.com/orientechnologies/orientdb/pull/8100 but it has one more step that cache the fileid of the ".irs" inside the index structure to avoid to check on the fs if the file it actually exist.

core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OIndexRIDContainer.java
```diff
@@ -55,8 +56,13 @@
...
    * Should be called inside of lock to ensure uniqueness of entity on disk !!!
    */
-  public OIndexRIDContainer(String name, boolean durableNonTxMode) {
-    fileId = resolveFileIdByName(name + INDEX_FILE_EXTENSION);
+  public OIndexRIDContainer(String name, boolean durableNonTxMode, AtomicLong bonsayFileId) {
+    if (bonsayFileId.get() == 0) {
+      fileId = resolveFileIdByName(name + INDEX_FILE_EXTENSION);
+      bonsayFileId.set(fileId);
+    } else {
+      fileId = bonsayFileId.get();
```
laa: I suppose to avoid double get of volatile field which will be cached in  most of the cases better to do 
```
fileId = bonsayFileId.get();
if (fileId == 0) {
      fileId = resolveFileIdByName(name + INDEX_FILE_EXTENSION);
   bonsayFileId.set(fileId);
} 
```

that is minor so up to you

tglman:
Hi @laa,

Changed that bit, let me know if we can merge this.

Regards

PR #8128, opened 2018-02-28
add microtx listener and refactor live queries to use it

luigidellaquila:


tglman:
We went to another direction, closing this, not needed anymore

PR #8138, opened 2018-03-05
fixed the exclusive lock in embedded with a feature flag to enable it.

tglman:


core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -2011,6 +2023,13 @@ public void preallocateRids(final OTransactionInternal clientTx) {
...
         }
       } finally {
+        if (pessimisticLock) {
+          for (ORecordOperation recordOperation : recordOperations) {
+            if (recordOperation.type == ORecordOperation.UPDATED || recordOperation.type == ORecordOperation.DELETED) {
+              releaseWriteLock(recordOperation.getRID());
+            }
+          }
+        }
         stateLock.releaseReadLock();
```
lvca: I think this like should be in the `finally{}` block in case the release above fails.

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -4001,12 +4020,19 @@ private ORawBuffer readRecordIfNotLatest(final OCluster cluster, final ORecordId
...
+        acquireReadLock(rid);
+      }
+
       ORawBuffer buff;
       checkOpenness();
 
       buff = doReadRecordIfNotLatest(cluster, rid, recordVersion);
       return buff;
     } finally {
+      if (pessimisticLock) {
```
lvca: You should check if the lock was acquired. If it wasn't it  throws a `OLockException` exception

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
tglman: you mean wrap the record lock release in a try{} finally {} ? in terms of business the release here should never fail, it may do due to a bug

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -1912,6 +1917,13 @@ public void preallocateRids(final OTransactionInternal clientTx) {
       final List<ORecordOperation> result = new ArrayList<>();
       stateLock.acquireReadLock();
       try {
+        if (pessimisticLock) {
+          for (ORecordOperation recordOperation : recordOperations) {
+            if (recordOperation.type == ORecordOperation.UPDATED || recordOperation.type == ORecordOperation.DELETED) {
+              acquireWriteLock(recordOperation.getRID());
```
laa: @tglman I did not find that you sort locked records by rids before locking, but you use exclusive locks not read locks. Are they already sorted or you just missed that step? Could you clarify?

PR #8139, opened 2018-03-05
moved the logic of live query from the listeners to direct calls

tglman:
Hi,

This move all the logic of the live query from the hooks/listeners to direct calls i didn't change the name of the classes to not make the commit too confusing, but probably is better rename them before push, this should fix as well the microTx commit


luigidellaquila:
Ok, for me. Merging

Thanks

Luigi

PR #8146, opened 2018-03-08
Optimize rid-bags' serialization

markodjurovic:
Optimize size of rid-bags serialization result, by using varints instead of fixed length data fields

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeRidBag.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeRidBag.java
@@ -668,8 +668,8 @@ public class OSBTreeRidBag implements ORidBagDelegate {
     // Keep this section for binary compatibility with versions older then 1.7.5
     OVarIntSerializer.write(bytes, size);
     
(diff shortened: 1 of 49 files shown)


lvca:
@markodjurovic are we keeping the previous serialization to open old databases?

core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeRidBag.java
```diff
@@ -619,21 +621,17 @@ public Object returnOriginalState(List<OMultiValueChangeEvent<OIdentifiable, OId
...
-    return result;
+  public int getSerializedSize() {    
+    return 0;    
   }
 
-  @Override
-  public int getSerializedSize(byte[] stream, int offset) {
-    return getSerializedSize();
-  }
+//  @Override
```
lvca: You can get rid of old methods, without commenting them. We have github for the history ;-)

markodjurovic: You are right :)

markodjurovic:
Invalid, not backward compatibile

PR #8147, opened 2018-03-08
Partial typed record deserialization optimization

markodjurovic:
Instead of copying result sub-arrays, now working with original array where record is marked with offset


lvca:
+1

tglman:
Good for me,
merging

PR #8152, opened 2018-03-10
Rid-bags serialization optimization

markodjurovic:
New rid-bags serializers using varints instead of fixed size length fields.
Introduced new rid-bags serializers' abstraction layer.


lvca:
Great! Is it compatible with the previous versions? Are we keeping both serializers?

markodjurovic:
Both serializers are present, so it is backward compatible. Old one is still default serializer.

lvca:
Cool, thanks. @tglman when can we enable as default all those changes by Marko?

laa:
I have a question, do we have any number form usage of both serializers?

tglman:
Hi all,

we actually had a private discussion we are going to simplify this changes, so probably this PR will be closed, the performance difference in term of cpu I don't think is going to be big, in term of size we will have an average of 50% less memory for the same content.

Regards

lvca:
@tglman how much time to redo this? Can we put this in v3.0 and change the impl later in some hotfix?

laa:
@tglman as I understand you do not have results of any measurements at this moment, is it possible to send numbers before and after implementation in terms of space and performance?

markodjurovic:
After conversation with @tglman we came to the conclusion that it can be done in another simpler way.

PR #8174, opened 2018-03-27
Optimize record header, flattened embedded maps

markodjurovic:
All changes are related only to V1 serializer. 

In header instead of writing value pointer(int), value_length(varint) is written. Before header new varint field header_length is written. With this approach, also there is no need to write 0 for field_name_length as signal for end of the header.

Embedded maps changes. So instead of having <key_type><key_value><value_data_pos><value_type>....<value_data>, 
value_data_pos is avoided value_data is written after value_type, and format looks like:
<key_type><key_value><value_type><value_data>

Debug deserialization responsibility is transferred to serializers.

Removed unnecessary recursive passing through all record children, instead when detecting records offsets , only current level records are processed.


markodjurovic:
will be created new PR

PR #8180, opened 2018-03-28
made sure that the HA statements execute the legacy commands

tglman:
Hi,

this make sure that the HA statements of the new executor execute the old logic, on not too long term this need to be replaced this implementation with a new implementation.

Regards


luigidellaquila:
Probably it's better to just copy/paste the old logic inside the new executor and extend `OSimpleExecStatement` (see OCreateFunctionStatement to have an example). 
It's definitely easy to do an it would not depend on the old executor

luigidellaquila:
OK, it's better than what we have now anyway, it works fine even though it relies on legacy API.
I'm merging it, I'll refactor it after 3.0 final

Thanks

Luigi

PR #8231, opened 2018-04-20
Optimize serialization header new flattenmap triple buffer

markodjurovic:
All changes are related only to V1 serializer.

In header instead of writing value pointer(int), value_length(varint) is written. Before header new varint field header_length is written. With this approach, also there is no need to write 0 for field_name_length as signal for end of the header.

Embedded maps changes. So instead of having <key_type><key_value><value_data_pos><value_type>....<value_data>,
value_data_pos is avoided value_data is written after value_type, and format looks like:
<key_type><key_value><value_type><value_data>

Debug deserialization responsibility is transferred to serializers.

Removed unnecessary recursive passing through all record children, instead when detecting records offsets , only current level records are processed.


core/src/main/java/com/orientechnologies/orient/core/serialization/serializer/record/binary/ORecordSerializerBinaryV1.java
```diff
@@ -824,124 +1112,120 @@ private static void writeSBTreeRidbag(BytesContainer bytes, ORidBag ridbag, UUID
...
       int size = OVarIntSerializer.readAsInteger(bytes);
       ridbag.getDelegate().setSize(size);
-      for (int i = 0; i < size; i++){
+      for (int i = 0; i < size; i++) {
         OIdentifiable record = readLinkOptimizedEmbedded(bytes);
-        if (record == null){
+        if (record == null) {
           //should never happen
           String errorMessage = "Deserialized null object during ridbag deserialization";
           OSerializationException exc = new OSerializationException("");
```
tglman: Can you add here the correct message

markodjurovic: as readLinkOptimizedEmbedded(byte[]) never returns null, this check is redundant, so it is removed

PR #8247, opened 2018-05-01
update maven-surefire-report-plugin with surefire.version

based2:


tglman:
Hi @based2,

This definitely make sense

PR #8258, opened 2018-05-10
Optimization: after sending last page, the server frees the query. Th…

lvca:
…e client does not need to explicitly send the close operation if the last page is fetched.


luigidellaquila:
Doing it in a slightly different way, so that the behavior is the same in embedded and remote.

Closing

PR #8278, opened 2018-05-21
Refactor noTx loop breaking logic for embedded databases

tglman:
Hi All,

This PR contains a couple of commit that refactor the logic of loop braking for the case of embedded, this new implementation will get all the record from the dirty manager, create a tx out of them and commit the tx, in this way we can use the tx way for generate rids and completely remove the need of creation of an empty record for breaking loops.

This implementation is an hard requirement for future evolutions in the distributed layer that require just one way to handle the data (the transaction).

Same logic could be implemented also for remote noTx, but currently is not done, will be done in future, with as well a major clean up of dirty management logic.


tglman:
Already merged

PR #8274, opened 2018-05-19
edit OrientDB license info so that GitHub recognizes it

eirinikos:
Hello! 👋

(CCing @dankohn who has requested this work so that the license will appear correctly in
https://landscape.cncf.io/selected=orient-db)

GitHub uses a library called Licensee to identify a project's license
type. It shows this information in the status bar and via the API if it
can unambiguously identify the license.

This commit modifies a few of OrientDB's docs so that Licensee is able
to recognize the repository's license type. It renames the Apache license
file to LICENSE and replaces its contents with the full text of the
Apache license. It also moves the info concerning 3rd-party software
to a new "Licensing" section in the README.

Collectively, these changes allow Licensee to successfully identify the
license type of OrientDB as Apache 2.0.

For comparison purposes, here is the output that I get when I run Licensee locally on OrientDB's remote repo:
```
$ licensee detect https://github.com/orientechnologies/orientdb/
License:        Other
Matched files:  license.txt, ODB-ASL-LICENSE.txt
license.txt:
  Content hash:  72bd30cd5a7d4d3b164171521cc25132decf304b
  License:       Other
  Closest licenses:
    Apache-2.0 similarity:  95.15%
    ECL-2.0 similarity:     92.46%
    EPL-2.0 similarity:     55.37%
ODB-ASL-LICENSE.txt:
  Content hash:  d70e5d500315df2a25cfa7b3e57e93afe151ba97
  Attribution:   Copyright 2010-2016 OrientDB LTD (http://orientdb.com)
  License:       Other
  Closest licenses:
    Unlicense similarity:  36.78%
    Zlib similarity:       33.33%
    MIT similarity:        31.17%
```

And here is the output that I get when I run Licensee on the `update-license` branch of my local OrientDB clone:
```
$ licensee detect ../orientdb/
License:        Apache License 2.0
Matched files:  LICENSE
LICENSE:
  Content hash:  4f3dee90d3ff02d566b067c25982c2bfd7e360c1
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       Apache License 2.0
```

PR #8288, opened 2018-05-23
edit OrientDB license info so that GitHub recognizes it

eirinikos:
Hello! 👋

(CCing @dankohn who has requested this work so that the license will appear correctly in
https://landscape.cncf.io/selected=orient-db)

This PR is a replacement of #8274. Here's a summary of the changes:
- `license.txt` contents have been replaced with only the full text of the Apache license
- info pertaining to 3rd party software licensing is now stored in the `README`
- `ODB-ASL-LICENSE.txt` has been removed
- `pom.xml` has been updated to reference `license.txt`

@luigidellaquila please let me know what you think. For the sake of clarity, it seems best to include only one file with the word `license` in its filename (this is why I removed `ODB-ASL-LICENSE.txt`).

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,26 @@ OrientDB Community Edition is free for any use (Apache 2 license). If you are in
 [Get started with OrientDB](http://orientdb.com/getting-started/).
 
 --------
+
+## Licensing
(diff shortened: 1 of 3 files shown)


luigidellaquila:
Merged, thanks!

Luigi

PR #8293, opened 2018-05-25
For geoserver

markodjurovic:
Extended geospatial functionality in order to use it with geotools jdbc
*In metadata added info about embedded type field class type
*Added support for generic Geometry
*Added support for ST_Overlaps. ST_Overlaps is slightly different than already supported overlaps (operator &&)
*Handling geometry collections


core/src/main/java/com/orientechnologies/orient/core/record/impl/ODocument.java
```diff
@@ -800,7 +800,11 @@ else if (fieldValue instanceof OIdentifiable) {
               "The field '" + p.getFullName() + "' has been declared as " + p.getType() + " with linked class '" + embeddedClass
                   + "' but the record has no class");
 
-        if (!(doc.getImmutableSchemaClass().isSubClassOf(embeddedClass)))
+        OImmutableClass docClass = doc.getImmutableSchemaClass();
+        OClass shapeClass = doc.getSchemaClass("OShape");
+        OClass geometryClass = doc.getSchemaClass("OGeometry");
+        if (!(docClass.isSubClassOf(embeddedClass)) && !(docClass.isSubClassOf(shapeClass) && embeddedClass
```
tglman: Hi,

What is the case for this check, I see here we use to check that the class was a subclass of the embedded class declared i the schema, so why do we need to check this also for the OShape/OGeometry ?

markodjurovic: Field of type OGeometry is generic geometry type and can contains any of OShape instances . So we don't want that storing some shape object in generic geometry field raise OValidationException

PR #8296, opened 2018-05-28
Serializer new optimization

markodjurovic:
*For v0 serializer optimized calculating embedded field length, by trying to read position of next field and then to subtract that position from position of current field.
*In v1 serializer removed unnecessary calls to deserializeValue function. No need for them because in in v1 header contains info about length of the field.
*Covered more types in deserializeValue when is used just to achieve length of the field


tglman:
Hi,

this is fine for me, if all test pass is go for me.

Regards

Issue #8285, opened 2018-05-23
Issues with starting embedded server


hrstoyanov:
### OS: Windows 7 64-bit

 
I am trying to start an Embedded server (over a database that I have previously created with ```console.bat```) with his piece of code:

```java
        //....
        log.info(()->"ODB-JCA resource adapter starting ...");
        log.info(()->"ODB-JCA server configuration: " + getEmbeddedServerConfiguration());
        log.info(()->"ODB-JCA server home: " + getOrientdbHome());
        if (embeddedServerConfiguration != null) 
        try {
            embeddedServer = OServerMain.create(true);
            embeddedServer.startup(new File(embeddedServerConfiguration));
            embeddedServer.activate();

            log.info(()->"ODB-JCA database directory: " + embeddedServer.getDatabaseDirectory());
            log.info(()->"ODB-JCA server id: " + embeddedServer.getServerId());
            log.info(()->"ODB-JCA server activated: " + embeddedServer.isActive());

            Map<String, String> storageNames = embeddedServer.getAvailableStorageNames();
            if (storageNames != null) {
                storageNames.forEach((String k, String v) -> log.info(()->"ODB-JCA storage: " + k + ", " + v));
            }

            Map<String, Class<? extends ONetworkProtocol>> networkProtocols = embeddedServer.getNetworkProtocols();
            if (networkProtocols != null) {
                networkProtocols.forEach((String k, Class<? extends ONetworkProtocol> v) -> {
                    log.info(()->"ODB-JCA network protocol name: " + k + ", " + v.getSimpleName());
                    OServerNetworkListener l = embeddedServer.getListenerByProtocol(v);
                    log.info(()->"ODB-JCA protocol listener inbound addr: " + l.getInboundAddr() + ", active:" + l.isActive() + ", alive:" + l.isAlive());
                });
            }

            log.info(()->"ODB-JCA resource adapter successfully started");

        } catch (Exception e) {
            log.log(Level.SEVERE,"ODB-JCA resource adapter failed to start", e);
            throw new ResourceAdapterInternalException(e);
        }
```
## Expected behavior 
I expected the server to start with no issues. 

## Actual behavior 
Instead, this call ```embeddedServer.activate();``` bombs with the below exception - __why is the server trying to create db system files when the database already exists__? Also, looking at the database folder, I do see a lot of files, but not __OSystem__ ?
```
[...]

Issue #8284, opened 2018-05-22
LOAD SCRIPT  screws up the path in Windows


hrstoyanov:
## Expected behavior  
>LOAD SCRIPT C:\temp\xxxxxx\recres\recres-webapp\src\main\odb\scripts\schema.sql 

should work as expected

## Actual behavior  
I see this instead:
```
ecres-webapp\src\main\odb\scripts\schema.sql...
ecres-webapp\src\main\odb\scripts\schema.sql'runcs\
```

which tells me that the LOAD SCRIPT command is trying to interpret the path and ```\r```, ```\p``` makes it bomb. This is scary! The console app **should never try to parse the path**, but accept is a provided by the user.

## Steps to reproduce  
One possible workaround is to force UNIX style paths. Then it works .. but cammon? 
What does not work is if one tries to wrap the path in "" or ''.

markodjurovic:
@hrstoyanov Thank you for reporting this issue. Fix will be in v3.0.2

PR #8299, opened 2018-05-31
Fix for Issues #8285 and #8284

markodjurovic:
* For issue #8285 support for legacy is removed. Cause of the issue is that that DB is detected as legacy DB.
* For issue #8284 is fixed encoding of jump characters. Now windows paths can be used as paths pointing to script. Nevertheless windows path should be between quotes or single quotes

Please pay extra attention to changes in OStringParser, because this tokenizer is used on few places

### DIFF
--- a/core/src/main/java/com/orientechnologies/common/console/OConsoleApplication.java
+++ b/core/src/main/java/com/orientechnologies/common/console/OConsoleApplication.java
@@ -303,7 +303,26 @@ public class OConsoleApplication {
       // COMMENT: JUMP IT
       return RESULT.OK;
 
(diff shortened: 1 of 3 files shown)


laa:
@markodjurovic Can we instead of fixing of StringParser apply come local fix? You see string parse is like "treatment from all diseases" so it is risky to change it and does not have side effects.

hrstoyanov:
Thanks, when is 3.0.2 going to be released in the  Maven repo?

markodjurovic:
@laa 
I can implement case specific fix. I will revert changes in StringParser for now, and later we can discuss about StringParser fix (I think it is valid fix, but let double check it).

markodjurovic:
OK I will close this PR and create new one

PR #8308, opened 2018-06-05
Sharding with Graph Database 2.2.x

HassounDev:
Hello, i have this configuration :

{
  "autoDeploy": true,
  "readQuorum": 1,
  "writeQuorum": "majority",
  "executionMode": "undefined",
  "readYourWrites": true,
  "newNodeStrategy": "dynamic",
  "servers": {
    "*": "master"
  },
  "clusters": {
    "internal": {
    },
    "client_usa": {
      "servers" : [ "usa"]
    },
    "client_europe": {
      "servers" : [ "europe" ]
    },
    "client_china": {
      "servers" : [ "china"]
    },
    "*": {
      "servers" : [ "<NEW_NODE>" ]
    }
  }
}
3 Nodes; usa, china and europe. I want to insert data in each node, for example insert into cluster:client_usa(id) values('my_id'), that's work for me, but it's not my objective, my objective is to have a graph data-set in each cluster (Vertexes and Edges), to create vertexes and edges in a cluster you should create classes before; and a cluster can't handle more than one class...

So my question is how to create multiple vertexes and edges into each cluster


luigidellaquila:
In general, develop and 2.2.x branches diverged too much, so they cannot be merged.
I'd just close this PR and manage the problem as a normal issue

Thanks

Luigi

Issue #5650, opened 2016-01-28
Suggestion - add database server version info on the "info" command 


smolinari:
Hi,

When using the console, it would be nice to be able to verify the database server's version number with the "info" command. Or is there another way to see the version?

See this discussion on GG for more info.

https://groups.google.com/forum/#!topic/orient-database/aXxfgI4Kp4Q

Scott

maggiolo00:
+1

tglman:
This is a good canditate for a new contributor

PR #8317, opened 2018-06-10
Adding server version in info response #5650

tictakk:
Browsing through some of the issues I picked #5650 so I could familiarize myself with the code base. I don't think this is exactly what we want. I'm thinking this probably should be stored in the OStorageConfiguration object but I figured I'd go ahead and start this request and get some feedback. Thanks.

Result after change.
+----+-----------------+-------------------+
|#   |NAME             |VALUE              |
+----+-----------------+-------------------+
|0   |Server Version     |3.0.3             |
|1   |Version                 |20                 |
|2   |Conflict-Strategy  |version          |
|3   |Date-Format        |yyyy-MM-dd  |
|4   |Datetime-Format  |yyyy-MM-dd HH:mm:ss|
|5   |Timezone              |America/New_York   |
|6   |Locale-Country     |US               |
|7   |Locale-Language  |en                |
|8   |Charset                 |UTF-8          |
|9   |Schema-RID         |#0:1             |
|10  |Index-Manager-RID |#0:2         |
+----+-----------------+-------------------+

### DIFF
--- a/tools/src/main/java/com/orientechnologies/orient/console/OConsoleDatabaseApp.java
+++ b/tools/src/main/java/com/orientechnologies/orient/console/OConsoleDatabaseApp.java
@@ -1397,6 +1397,7 @@ public class OConsoleDatabaseApp extends OrientConsole implements OCommandOutput
       if (dbCfg.getName() != null)
         resultSet.add(new ODocument().field("NAME", "Name").field("VALUE", dbCfg.getName()));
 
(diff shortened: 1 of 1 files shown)


tglman:
Hi, 

Thank you for this, i'll check and merge.

Regards

tglman:
Hi @tictakk,

This is not exactly correct because is getting the version from the console, that may be different from the server you are connecting to.

Closing

PR #8341, opened 2018-06-18
Redesign of integration tests using Docker 

robfrank:
This PR redesign the integraton tests of the distribution module and it use docker via  the testcontainers library:

- adds testcontainers dependency
- removes tesstNG and rely on junit
- removes custom startup code
- generate a docker image of the current build tagged as orientdb/orientdb:latest
- reviews demodb generation: split osql 

collateral
- adds the mvn wrapper to  enable building on systems without maven installed

Tu create distribution package and run qa tests:

``` mvn clean install -Pqa```


tglman:
Hi,
After this merge i'm getting this error: https://gist.github.com/tglman/883803d6d9182e7d080455a07f3c4fb4 in the base `mvn clean install` of the project, are we including a dependence to have a docker environment running for build orient ? If this is the case I would be against it.

Regards

robfrank:
@tglman yes, it needs docker to run integration tests. Each test spawns an ephemeral container built over the latest snapshot of orientdb and tests "something". 
Most of the tests are in place to test the demodb consistency, other are related to some issues closed in the past.
IMHO it is an easy and flexible way to execute integration tests without bothering with System Process in Java.
Write a test for a 2/3 nodes installation is a matter of a "compose" file. 
For sure I can reconfigure the pom to avoid the build of the  docker image at installation time and move it on the qa profile.

PR #8432, opened 2018-07-25
Diff update redesign tracking

markodjurovic:
Idea is to have documents' delta(old vs new value), so e.g. only deltas can be send in document update process.
Currently real deltas can be calculated only for lists and documents, and documents embedded in lists and lists embedded in documents...
If change is detected in other complex type (map, set, ridbag, ...) value (or in some embedded value) the whole root value will be marked as delta.
Usage of deltas as updates can be turned on and off. Currently by default they are turned off. Switch is flag in OTransactionPhase1Task class


core/src/main/java/com/orientechnologies/orient/core/metadata/schema/OType.java
```diff
@@ -108,13 +94,15 @@
 
   LINKBAG("LinkBag", 22, ORidBag.class, new Class<?>[] { ORidBag.class }),
 
-  ANY("Any", 23, null, new Class<?>[] {});
+  ANY("Any", 23, null, new Class<?>[] {}),
+
+  DELTA_RECORD("DeltaRecord", 24, ODocumentDelta.class, new Class<?>[] { ODocumentDelta.class });
```
tglman: The OType are the types allowed as value in OrientDB, define a new OType implies that you can use it as a field type or as schema property type, and as soon as you define a new OType this need to be supported everywhere (from binary protocol to query engine), I understand the use case for this new type in the diff serialization, but probably is less impacting define specific serialization types just inside the diff specification, like for example: FULL_CONTENT, DIFF_CONTENT, REMOVED  and leave the OTypes unchanged.

core/src/main/java/com/orientechnologies/orient/core/record/impl/ODocument.java
```diff
@@ -3169,4 +3366,341 @@ protected void unTrack(OIdentifiable id) {
...
+          //now find a diff
+          List currentList = (List) currentValue;
+          List previousList = (List) previousValue;
+          int i = 0;
+          for (i = 0; i < currentList.size(); i++) {
+            Object currentElement = currentList.get(i);
+            if (i > previousList.size() - 1) {
+              //these elements doesn't have pair in original list
+              ODocumentDelta deltaElement = new ODocumentDelta();
+              deltaElement.field("t", UpdateDeltaValueType.LIST_ELEMENT_ADD.getOrd());
```
tglman: Use a map to set values in the ODocumentDelta is definitely fine, anyway if we have this fields coded in our code would be more efficient have the real java fields inside the ODocumentDelta, anyway this is not blocking i'm happy to merge this part as it is now, and optimize it in a later future.

core/src/main/java/com/orientechnologies/orient/core/delta/ODocumentDeltaSerializerV1.java
```diff
@@ -0,0 +1,583 @@
...
+  private void serialize(ODocumentDelta deltaDoc, BytesContainer bytes) {
+    for (Map.Entry<String, Object> entry : deltaDoc.fields.entrySet()) {
+      String fieldName = entry.getKey();
+      Object fieldValue = entry.getValue();
+
+      //serialize field name
+      serializeValue(bytes, fieldName, OType.STRING);
+
+      //serialize field type
+      OType type = OType.getTypeByClass(fieldValue != null ? fieldValue.getClass() : null);
```
tglman: this is semantically correct, anyway have the type directly in ODocumentDelta sourced directly from the ODocument may be more efficient.

core/src/main/java/com/orientechnologies/orient/core/metadata/schema/OType.java
markodjurovic: I agree with this. 
Just one explanation. Related to deltas there are two types of type. One is operation type  (core/src/main/java/com/orientechnologies/orient/core/record/impl/UpdateDeltaValueType.java). This one is equivalent to your example FULL_CONTENT, DIFF_CONTENT, ...
The second one is value type and it is equivalent to OType, but it has one more data type to handle, and that is Delta Record type.
So as OType has very specific purpose I've introduced new interface (core/src/main/java/com/orientechnologies/orient/core/metadata/schema/OTypeInterface.java) and new enum (core/src/main/java/com/orientechnologies/orient/core/delta/ODeltaDocumentFieldType.java) just for deltas

core/src/main/java/com/orientechnologies/orient/core/delta/ODocumentDeltaSerializerV1.java
markodjurovic: I've done this optimization and right now delta document field carries type of field fetched from ODocument (when it is possible)

markodjurovic:
Added support for embedded ridbags delta

markodjurovic:
It will be reopened when time come

markodjurovic:
Time has come :)

PR #8481, opened 2018-08-23
Sequnces

markodjurovic:
Fixed issues and added new features to sequences:

- Upper and lower limit
- Cycling sequences
- Ascending and Descending sequences
- Warnings on non cycling sequences near limit value

Also introduced new SQL keywords for creating and altering sequences:
Both in create and alter:
MAXVALUE, MINVALUE, CYCLE, ASC, DESC

and just for alter:
NOLIMIT

Also added null check in ViewThread just to prevent null pointer exception, although exception didn't effect stability of system


core/src/main/java/com/orientechnologies/orient/core/metadata/sequence/OSequence.java
```diff
@@ -159,8 +223,27 @@ public synchronized boolean updateParams(CreateParams params) {
...
+      any = true;
+    }
+
+    if (params.turnLimitOff != null && params.turnLimitOff == true) {
+      this.setLimitValue(null);
+    }
+
     save();
-    reset();
+//    reset();
```
tglman: I do not know the behaviour of reset, but if you can keep it out just remove the line of code

markodjurovic: Done, thanks

tglman:
Hi, 

Pulled this and looks good for me @luigidellaquila could you double check the sql part ?

Regards

luigidellaquila:
Doing it now

luigidellaquila:
Why `setLimitValue()`? Don't you need min and max value? If it's like this, probably it makes sense to have only one keyword also in SQL

markodjurovic:
@luigidellaquila 
I agree with you to unify these two keywords into one. Right now instead of MINVALUE and MAXVALUE , to define limit value LIMIT should be used

markodjurovic:
CYCLE now expects parameter TRUE or FALSE. So there is no NOCYCLE any more

PR #8483, opened 2018-08-24
Fixed OCommandSQLParsingException which throws when user try update CUSTOM attribute from Java API

WeaxMe:
Wrapped custom name by `` for execute alter command in distributed database mode

Fixed OCommandSQLParsingException which throws when user try update CUSTOM
attribute from Java API

## Steps for reproduce
1. Run OrientDB in distributed mode
2. Try to update custom attribute for class or property. Name of custom property must contains dot for reproduce bug.
`property.setCustom("custom.name", "value")`

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/metadata/schema/OClassImpl.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/metadata/schema/OClassImpl.java
@@ -225,10 +225,10 @@ public class OClassImpl extends ODocumentWrapperNoClass implements OClass {
       final OStorage storage = database.getStorage();
 
       if (storage instanceof OStorageProxy) {
-        final String cmd = String.format("alter class `%s` custom %s=%s", getName(), name, value);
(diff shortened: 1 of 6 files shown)


tglman:
Hi @WeaxMe,

Could you provide also a test case that prove the fix in the PR ?

Regards

WeaxMe:
@tglman Yes will add tests

PhantomYdn:
@tglman , we are in Orienteer (https://github.com/OrienteerBAP/Orienteer) are working on support of cluster for it. Is there a way to run all single node test cases from OrientDB but in distributed mode? It will be great to make sure that there is no other hidden problems at just one shot:)

WeaxMe:
I added tests in 427fd3d5fbc1952228b08bfb8fdf410afdb9a6a2

Btw. When will be release 2.2.37 ?

WeaxMe:
What about PR?

tglman:
looks good for me, merging

Issue #8488, opened 2018-08-27
EMBEDDEDMAP of Class


freeart:
## Expected behavior  
It might work. I tried to create EMBEDDEDMAP to string. And it works fine.   

## Actual behavior  
<com.orientechnologies.orient.core.exception.OValidationException: The field 'DTest.document' has been declared as EMBEDDEDMAP but an incompatible type is used. Value: {name=testvalue} DB name="db1" Error Code="4">

## Steps to reproduce
I have a class ODocument which contains field `name` type STRING 
The second class DTest has EMBEDDEDMAP to ODocument
CREATE PROPERTY `DTest`.`document` EMBEDDEDMAP  `ODocument`
ALTER PROPERTY `DTest`.`document` linkedClass "ODocument" 
To write any data into document I do:
UPDATE DTest SET document = {"testkey":{"name":"testvalue"}}
WHERE id = "123";

markodjurovic:
Hi @freeart 
Thank you for reporting this one.
Solution will require that JSON contains class name of document. So update command should look like this:
`UPDATE DTest SET document = {"testkey":{"@class":"ODocument", "name":"testvalue"}} WHERE id = "123"`
(same for insert)
The fix will be in next release of 3.0 version, unless you want to build it from source and in that case you can do right now.

PR #8494, opened 2018-08-28
Issue #8488 proposal

markodjurovic:
Related to #8488 . Cause of the issue is because value of the embedded map entry is recognized as map instead as document instance. Idea is if in json "@class" field exists that json will be treated as document of specified class. So, with this fix, command:
`UPDATE DTest SET document = {"testkey":{"@class":"ODocument", "name":"testvalue"}} WHERE id = "123";`

will be executed as expected


luigidellaquila:
Excellent! We should also port it to 2.2.x and 3.0.x

Luigi

Issue #8498, opened 2018-08-29
pathElements duplicates


creisle:
## Expected behavior  
Return only no duplicate vertices 

## Actual behavior  
Returns many duplicates

## Steps to reproduce  

This affects 3.0.6 and works correctly in 2.2.35

```
> create vertex v set name = 'bob'
#9:0
> create vertex v set name = 'alice'
#9:1
> create edge from #9:0 to #9:1
> match {class: V, where: (name = 'bob')}.both(){while: ($depth < 10)} return $pathElements
```

in 2.2.35 the match query returns (as expected)
![image](https://user-images.githubusercontent.com/4342641/44821512-f0f24780-abaa-11e8-8713-7e3bee0a3579.png)
but in 3.0.6 it returns
![image](https://user-images.githubusercontent.com/4342641/44821524-0bc4bc00-abab-11e8-9b63-b525a63db836.png)

luigidellaquila:
Hi @creisle 

Yes, this is an expected behavior in v 3.0, please see https://orientdb.com/docs/3.0.x/release/3.0/SQL-Changes.html#match-and-distinct

You just have to add a DISTINCT to the RETURN block to make it behave like in v 2.2

Thanks

Luigi

PR #8499, opened 2018-08-30
Solution proposal for #8498

markodjurovic:
As described in  #8498 all unrolled elements will be contained in result set, which is contradictory to documentation:

> $pathElements (since 2.2.1), the same as $paths, but for each node present in the pattern, a single row is created in the result set (no duplicates)

 My idea is to track all elements previously fetched, and to ignore all reoccurred elements


markodjurovic:
It looks like that described behavior is valid, so no need for fix

PR #8508, opened 2018-09-03
Remote pessimistic locking

tglman:
Hi All,

this add a new api for locking records and implement it in remote, the new api is just ` db.lock()` without considering possibilities of read/write lock, we should check the use of lock api in the query engine.
I had to reimplement as well the lock manager to not be thread-bound.


tglman:
already merged manually closing

Issue #8611, opened 2018-10-18
Import database with -merge=true but classes still removed


jymcheong:
## Expected behavior  
import database with -merge=true should not delete existing classes 

## Actual behavior  
![](https://www.dropbox.com/s/d4px6b9x8q3uffp/Screenshot%202018-10-18%2021.49.57.png?raw=1)

## Steps to reproduce  

1. Create a "test" database
2. Create a "MyV" vertex
3. Create a dummy function
4. Export only functions as per https://orientdb.com/docs/2.1.x/Console-Command-Export.html
5. Import functions.gz

markodjurovic:
Hi @jymcheong 
thank you for reporting this one. Fix will be in release 3.0.10, or you can build it from 3.0.x branch right now.

PR #8614, opened 2018-10-19
Import options fix

markodjurovic:
Fix issue described in https://github.com/orientechnologies/orientdb/issues/8611. 
Also in `OWALFile` switched to create file if not exists, because tests in `OETLProcessorTest` fail because wal file exists


markodjurovic:
merge to invalid branch

PR #8807, opened 2019-03-09
Update OTransactionOptimistic.java

robinZhao:
different transaction type between client and remote server when no changing data in the transaction.

once this pooled resource reused to execute ddl statement ,server response error.

eg: client NoTx but remote server is OTransactionOptimistic

I don't know if this will affect the remote server side, tested this modify at client side only;

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/tx/OTransactionOptimistic.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/tx/OTransactionOptimistic.java
@@ -532,9 +532,8 @@ public class OTransactionOptimistic extends OTransactionRealAbstract {
 
     status = TXSTATUS.COMMITTING;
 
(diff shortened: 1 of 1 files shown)


tglman:
Hi @robinZhao,

This make sense we will double check it, can you sign clahub in the meanwhile ?

Bye

robinZhao:
> Hi @robinZhao,
> 
> This make sense we will double check it, can you sign clahub in the meanwhile ?
> 
> Bye

i signed..

tglman:
Hi @robinZhao,

I just went through this code again, i think this is trying to fix an issue resolved recently with this commit: https://github.com/orientechnologies/orientdb/commit/a292545f1324ce71a3a3c26ef219c9f06ce6bdac  so i'm closing this, in case is not the case can you reopen this re-explaining the new case ?

Regards

Issue #8891, opened 2019-06-01
[Bug Report] Read timeout (always) on select `outE()` - Reproducible


matanshukry:
## Expected behavior
Return edges

## Actual behavior  
Hangs and after a while read timeout

## Steps to reproduce  
```
create class person extends V
create property person.name String

create class friend extends E
create property friend.status String


create vertex person set name="Bob1"
create vertex person set name="Bob2"
create vertex person set name="Bob3"
create vertex person set name="Bob4"

create edge friend from (select from person where name="Bob1") to (select from person where name="Bob2") set status="Good"
create edge friend from (select from person where name="Bob2") to (select from person where name="Bob3") set status="Very Good"
create edge friend from (select from person where name="Bob3") to (select from person where name="Bob4") set status="Bad"

orientdb {db=my_db}> select outE() from person
```

## Extra query info

```
orientdb {db=my_db}> select from person

+----+-----+------+----+----------+---------+
|#   |@RID |@CLASS|name|out_friend|in_friend|
+----+-----+------+----+----------+---------+
|0   |#24:0|person|Bob1|[#25:0]   |         |
|1   |#24:1|person|Bob2|[#25:1]   |[#25:0]  |
|2   |#24:2|person|Bob3|[#25:2]   |[#25:1]  |
|3   |#24:3|person|Bob4|          |[#25:2]  |
+----+-----+------+----+----------+---------+

4 item(s) found. Query executed in 0.014 sec(s).


orientdb {db=my_db}> select from friend

+----+-----+------+-----+-----+---------+
|#   |@RID |@CLASS|out  |in   |status   |
+----+-----+------+-----+-----+---------+
|0   |#25:0|friend|#24:0|#24:1|Good     |
|1   |#25:1|friend|#24:1|#24:2|Very Good|
|2   |#25:2|friend|#24:2|#24:3|Bad      |
+----+-----+------+-----+-----+---------+
```

## Other technical info

- The log had no error either
- when using `list connections` in another console, it just shows "Listening" for command.

p.s.
I was trying to do some other queries and it seems they are also causing read timeout.
e.g.
```
select unionall(tags, [2,3]) from place
```
where place is a vertex that has 3 records in it, tags is a property containing an embeddedlist, each record contains different values.

## Actual Bug
Did some debugging and found the stack trace (which is silented). I tried to see if there were any recent changes around the code but couldn't find any.

```
[...]

PR #8892, opened 2019-06-02
Fixed checking of embeddedlist serialization output

matanshukry:
Solves #8891


tglman:
Hi @matanshukry,

This is a good job, and it actually fix the issue, though the problem is not in the serialization, the serialization support some subset of values and should support only that, we are reviewing the query engine to make sure we never fill a result with a value that cannot be serialized, thank you for your amazing job, with the details from the issue we could pin point the exact failure!

PR #8908, opened 2019-06-14
Wip visibility api

tglman:
Hi, 

List of changes of this PR:
- Refactor of Security API to remove deprecated APIs 
- Refactor of the security Implementation to remove the use of Thread Locals and deprecated internal APIs usages
- First definition of the internal API for filtering read/write access to properties
- Test cases and implementation of filtering API in ODocument


core/src/main/java/com/orientechnologies/orient/core/metadata/security/OPropertyAccess.java
```diff
@@ -0,0 +1,23 @@
...
+
+  public OPropertyAccess(ODocument document, OSecurityInternal security) {
+    filtered = security.getFilteredProperties(document);
+  }
+
+  public OPropertyAccess(Set<String> filtered) {
+    this.filtered = filtered;
+  }
+
+  public boolean isReadable(ODocument document, String property) {
```
luigidellaquila: IMHO this is extremely confusing. The relationship between the objects is implicit and it makes it terribly error prone (eg. use the wrong OPropertyAccess for an ODocument).
I'd prefer to have something linke

```
doc.getPropertyAccess().isReadable("property")
```

And to avoid the constructor that uses the security (ie. the security should build the OPropertyAccess)

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentAbstract.java
```diff
@@ -1429,17 +1429,17 @@ public void release() {
    * Creates a new ODocument.
    */
   public ODocument newInstance() {
-    return new ODocument();
+    return new ODocument(this);
```
luigidellaquila: What is the reason to do this? 
This thing scares me a lot... there is always a chance that the user does a `new ODocument().save()`

tglman: this is just removing an access to the thread local, inside the `new ODocument()` we access to the tread local to get the current database, i just moved the lookup of the database from the thread local to the API itself.

core/src/main/java/com/orientechnologies/orient/core/metadata/security/OPropertyAccess.java
tglman: So you mean that we should have na OPropertyAccess that works for all the documents ?

luigidellaquila: No, I mean the opposite.
If you want to have a pre-populated bean that contains access rules, it should be stored in the document, and not built on-the-fly: 

```
doc.getPropertyAccess().isReadable("property")

instead of 

x = new OPropertyAccess(doc, security);
x.isReadable(doc2, "property"); //see doc vs doc2

```
If you want to create it every time you use it, then you should ask for it to the security:

```
security.getPropertyAccess(doc).isReadable("property")
```

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentAbstract.java
luigidellaquila: OK, but what if you close the DB and then save the document with another DB? 
We are changing the behaviour and breaking backward compatibility

luigidellaquila: Oh ok, I got it, never mind, it's good ;-)

core/src/main/java/com/orientechnologies/orient/core/metadata/security/OPropertyAccess.java
tglman: oh yep, i definitely think we can remove the parameter from `isReadable` we can work around in case we need the doc in isReadable to keep inside the one you have in the constructor

PR #8948, opened 2019-07-08
Add unit tests for com.orientechnologies.orient.core.sql.parser.JavaCharStream

JohnLBergqvist:
I've analysed your codebase and noticed that `com.orientechnologies.orient.core.sql.parser.JavaCharStream` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important in a subsequent PR.

I would like to sign a CLA as required in your contribution guidelines, but when I hit the submit button after filling in the form, I get an Application Error from heroku.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -282,6 +282,27 @@
             <artifactId>javax.activation-api</artifactId>
             <version>1.2.0</version>
         </dependency>
+
+        <dependency>
+          <groupId>org.powermock</groupId>
+          <artifactId>powermock-api-mockito</artifactId>
+          <version>1.6.6</version>
+          <scope>test</scope>
+        </dependency>
+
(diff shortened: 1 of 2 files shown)


tglman:
Hi @JohnLBergqvist, 

I appreciate your work, anyway the JavaCharStream is code that is generated by "JavaCC", so add test cases on it may just add some maintenance effort, probably would be more useful have more high level test on the correct parsing of specific queries.

Also on the implementation you add 3 test dependencies, we are trying to avoid any new dependencies.

Regards

PR #8951, opened 2019-07-09
Refactor for transaction graph link management to guarantee atomic operations also in delete and move

tglman:


core/src/main/java/com/orientechnologies/orient/core/record/impl/OVertexDocument.java
```diff
@@ -216,25 +216,36 @@ public OEdge addEdge(OVertex to, OClass type) {
...
+  protected static ORID moveTo(final String iClassName, final String iClusterName, OVertex toMove, ODatabaseSession db) {
+    if (toMove == null) {
+      throw new ORecordNotFoundException(null, "The vertex is null");
+    }
+    if (checkDeletedInTx(toMove.getIdentity())) {
+      throw new ORecordNotFoundException(toMove.getIdentity(), "The vertex " + toMove.getIdentity() + " has been deleted");
+    }
+    boolean moveTx = !db.getTransaction().isActive();
+    if (moveTx) {
+      db.begin();
```
luigidellaquila: perhaps a try/finally is more safe here...

PR #8962, opened 2019-07-19
Wip index rebuild without docs

tglman:
Hi, 

This change the rebuild of the index to use OResultBinary instead of ODocument and raw browsing of records instead of high level iterator.

Would be cool as well a bit more high level evaluation and consider if it does make sense to do this  rebuild on this way.

Regards

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/index/OCompositeIndexDefinition.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/index/OCompositeIndexDefinition.java
@@ -26,6 +26,7 @@ import com.orientechnologies.orient.core.db.record.ORecordElement;
 import com.orientechnologies.orient.core.metadata.schema.OType;
 import com.orientechnologies.orient.core.record.impl.ODocument;
 import com.orientechnologies.orient.core.sql.OCommandExecutorSQLCreateIndex;
+import com.orientechnologies.orient.core.sql.executor.OResult;
 
 import java.lang.reflect.InvocationTargetException;
 import java.util.*;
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/orientechnologies/orient/core/record/impl/ODocument.java
```diff
@@ -128,17 +128,17 @@
 
   protected Map<String, ODocumentEntry> fields;
 
-  protected           boolean                             trackingChanges        = true;
```
luigidellaquila: This could probably be a separate refactoring...

core/src/main/java/com/orientechnologies/orient/core/storage/cluster/v2/OClusterPositionMapV2.java
```diff
@@ -346,7 +346,7 @@ public void remove(final long clusterPosition, final OAtomicOperation atomicOper
           if (bucket.exists(i + index)) {
             final OClusterPositionMapBucket.PositionEntry val = bucket.get(i + index);
             assert val != null;
-            result[entriesCount] = new OClusterPositionEntry(startIndex + i, val.getPageIndex(), val.getRecordPosition());
+            result[entriesCount] = new OClusterPositionEntry(index + i, val.getPageIndex(), val.getRecordPosition());
```
luigidellaquila: Is it a fix for a different bug...?

tglman: yes, probably i will do a differeent PR for this

tglman: i moved this to da a separate commit

core/src/main/java/com/orientechnologies/orient/core/record/impl/ODocument.java
tglman: this was part of a separate refactoring, i pushed it separately now is not part of this PR

PR #8968, opened 2019-07-29
Ridbag management refactor

tglman:
Hi,

This delay the delete of ridbags tree to avoid concurrency issues between read and delete, this do not handle yet correctly the shutdown and the crash case, so we cannot merge it yet.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeCollectionManagerShared.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeCollectionManagerShared.java
@@ -20,15 +20,18 @@
 
 package com.orientechnologies.orient.core.storage.ridbag.sbtree;
 
+import com.orientechnologies.common.exception.OException;
 import com.orientechnologies.common.serialization.types.OIntegerSerializer;
 import com.orientechnologies.orient.core.OOrientShutdownListener;
 import com.orientechnologies.orient.core.OOrientStartupListener;
+import com.orientechnologies.orient.core.Orient;
 import com.orientechnologies.orient.core.config.OGlobalConfiguration;
 import com.orientechnologies.orient.core.db.ODatabaseDocumentInternal;
 import com.orientechnologies.orient.core.db.ODatabaseRecordThreadLocal;
 import com.orientechnologies.orient.core.db.record.OIdentifiable;
 import com.orientechnologies.orient.core.db.record.ridbag.ORidBag;
 import com.orientechnologies.orient.core.exception.OAccessToSBtreeCollectionManagerIsProhibitedException;
+import com.orientechnologies.orient.core.exception.ODatabaseException;
 import com.orientechnologies.orient.core.serialization.serializer.binary.impl.OLinkSerializer;
 import com.orientechnologies.orient.core.storage.impl.local.OAbstractPaginatedStorage;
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeCollectionManagerShared.java
```diff
@@ -186,4 +183,28 @@ public void clearPendingCollections() {
...
+    synchronized (lock) {
+      SBTreeBonsaiContainer container = treeCache.getQuietly(cacheKey);
+      assert container != null;
+      if (container.usagesCounter != 0 || container.lastAccessTime > System.currentTimeMillis() - delay) {
+        return false;
+      }
+
+      treeCache.remove(cacheKey);
+    }
+    OSBTreeBonsai<OIdentifiable, Integer> treeBonsai = this.loadSBTree(collectionPointer);
```
laa: I can not completely grasp it. At first, you remove tree rid bag from the cache. Then you load it into cache remove an instance, but keep a pointer into the cache again. Did I miss something?

tglman: loadSBTree do not read from the cache, go direct to the disc, i remove it from the cache to delete it also from there

laa: Load sbtree https://github.com/orientechnologies/orientdb/pull/8968/commits/539125530e614cce80a2964f6fb6120b671bfbf6#L101 delegates to https://github.com/orientechnologies/orientdb/blob/539125530e614cce80a2964f6fb6120b671bfbf6/core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeCollectionManagerAbstract.java#L165 which populates cache back. May be I missed somethign ?

tglman: yes you are right, i got it wrong, loadTree is acually the one that do not use the cache, loadSBTree does use it, so i'm going to make sure to use the one that do not access to the cache

core/src/main/java/com/orientechnologies/orient/core/storage/ridbag/sbtree/OSBTreeCollectionManagerShared.java
```diff
@@ -197,13 +197,11 @@ public boolean tryDelete(OBonsaiCollectionPointer collectionPointer, long delay)
 
       treeCache.remove(cacheKey);
     }
-    OSBTreeBonsai<OIdentifiable, Integer> treeBonsai = this.loadSBTree(collectionPointer);
+    OSBTreeBonsai<OIdentifiable, Integer> treeBonsai = this.loadTree(collectionPointer);
```
laa: I still have doubts about change. Let suppose we removed a tree from the cache. Then let suppose that thread is paused there. After that, we load rid bag again, and after that, we delete ridag in the original thread. So we have to delete rid bag inside the cache. Could you confirm or explain the reason why my doubts are not true?

tglman: yes this can happen, if though quite unlucky, the only solution here is protecting this delete operation with a lock, anyway this lock was not there before.

PR #8996, opened 2019-08-20
Wip refactor tracking

tglman:
Hi all,

This is a big refactor, it does restructure the way we track changes in the document and nested structures:

The behaviour before this refactor was: the document had a ODocumentEntry for each field that was used for track the full value changes (like value replaces) and as well in case the value was a "Tracked Collection" i will have also a listener to the collection and a list of changes of the collection, the tracked collections itself had only an event management system where the ODocument would attach.

After this refactor: the ODocumentEntry manage just the value changes and not anymore the tracked collections, each tracked collection has inside a tracker that keep all the changes, so this allow as to have nested collection at any level that an keep their own tracking, after this change as well now any tracked collection has a dirty flag, that is set on change, and in case of a collection change is flag and all the parent flags are set to dirty, so to make easy to track when a document is dirty witch collection field/collection/collection-element  is actually changed.

Please have a look and ask any question it comes to your mind.

This implementation it also invalidate the current delta implementation, that will be removed with another iteration after this is merged.

Regards


tglman:
Hi, 

Just to add a couple of details: this remove as well all the state management for the collections, that was use to avoid lazy load/tracking on serialization deserialization, relative specific API are implemented for this cases, on the other side the current API to access to dirty fields and changes timeline where not changed yet, so this PR do not impact any indexing or database structure, just the ODocument, tracked collections, and serialization.

tglman:
Merged by hand, here seems that github is not getting my rebase, closing this.

Issue #9037, opened 2019-10-08
SELECT @rid property list Result list the All elements are null.


ClownfishYang:
## Expected behavior  
Use Mybatis get the @rid property list of the class.

## Actual behavior  
Result list the All elements are null.

## Steps to reproduce  
Mybatis execute `SELECT @rid FROM V` SQL Statement all will this issues.

## Troubleshoot
I found that the issues in `getResult` method in `org.apache.ibatis.type.BaseTypeHandler` class of Mybatis through debug application.The method be able to judged whether the result value was null according to `wasNull` method in `java.sql.ResultSet` interface class, and returned null value when `wasNull` method was true.
But the issues is that the `com.orientechnologies.orient.jdbc.OrientJdbcResultSet` in the class have not properly maintain the return value of the `wasNull` methdo, which is the property `lastReadWasNull` value.
issues code:
>     public Object getObject(String columnLabel) throws SQLException {
        if (!"@rid".equals(columnLabel) && !"rid".equals(columnLabel)) {
            if (!"@class".equals(columnLabel) && !"class".equals(columnLabel)) {
                try {
                    Object value = this.result.getProperty(columnLabel);
                    if (value == null) {
                        this.lastReadWasNull = true;
                        return null;
                    } else if (value instanceof ORecordLazyMultiValue) {
                        ORecordLazyMultiValue lazyRecord = (ORecordLazyMultiValue)value;
                        lazyRecord.convertLinks2Records();
                        this.lastReadWasNull = lazyRecord == null;
                        return lazyRecord;
                    } else {
                        this.lastReadWasNull = value == null;
                        return value;
                    }
                } catch (Exception var4) {
                    throw new SQLException("An error occurred during the retrieval of the Java Object at column '" + columnLabel + "'", var4);
                }
            } else {
                String r = (String)this.result.toElement().getSchemaType().map((t) -> {
                    return t.getName();
                }).orElse((Object)null);
                this.lastReadWasNull = r == null;
                return r;
            }
        } else {
            return this.result.getIdentity().toString();
        }
    }	
	public String getString(String columnLabel) throws SQLException {
        if (!"@rid".equals(columnLabel) && !"rid".equals(columnLabel)) {
            if (!"@class".equals(columnLabel) && !"class".equals(columnLabel)) {
                try {
                    String r = (String)Optional.ofNullable(this.result.getProperty(columnLabel)).map((v) -> {
                        return "" + v;
                    }).orElse((Object)null);
[...]

PR #9038, opened 2019-10-08
revise jdbc result set lastReadWasNull property

ClownfishYang:
Details about the pull request I have detailed on the issues #9037 .


luigidellaquila:
Hi @ClownfishYang 

Thank you very much, it definitely makes sense.
I'm merging it

Thanks

Luigi

PR #9078, opened 2019-11-27
ISSUE-8842 Resolves client connection strategy problem even if one se…

ozgursucu:
Resolves client connection strategy problem even if one sets the property other than STICKY, It does not effect the strategy value because It is controlled with deprecated "connectionStrategy" key value whereas it is set as "client.connection.strategy". Another bug is about getting next available url when the strategy is ROUND_ROBIN_CONNECT. When one of the servers is somehow down and dropped from the server urls and when the current index was the last one as the index management is not handled before getting from array list, The application always gets ArrayIndexOutOfBound Exception.


tglman:
Hi @ozgursucu,

This PR looks good, i will double check it an merge soon.

Regards

ozgursucu:
Hi @tglman 
I could not find the fix in 3.0.27?

ozgursucu:
What is the effective version of this pull request? I could not find the connectionStrategy solution in 3.0.42 version?

Issue #9099, opened 2019-12-31
Can't TRAVERSE through LINKMAP


mmacfadden:
## Expected behavior  
Ability to use TRAVERSE in the Document API through a LINKMAP.

## Actual behavior  
Only the initial entry point node is returned. 

## Steps to reproduce  
1. Create the DataValue class
```SQL
CREATE CLASS DataValue;
CREATE PROPERTY DataValue.id String;
CREATE PROPERTY DataValue.children LINKMAP DataValue;
```

2. Insert Data and then `TRAVERSE`
```SQL
LET r1 = INSERT INTO DataValue SET id="r1", children={};
LET r2 = INSERT INTO DataValue SET id="r2", children={};

LET r1c1 = INSERT INTO DataValue SET id="r1c1", children={};
LET r1c2 = INSERT INTO DataValue SET id="r1c2", children={};
LET r1c2c1 = INSERT INTO DataValue SET id="r1c2c1", children={};
LET r1c2c2 = INSERT INTO DataValue SET id="r1c2c2", children={};

LET r2c1 = INSERT INTO DataValue SET id="r2c1", children={};
LET r2c2 = INSERT INTO DataValue SET id="r2c2", children={};

UPDATE DataValue SET children.c1 = $r1c1[0] WHERE id = "r1";
UPDATE DataValue SET children.c2 = $r1c2[0] WHERE id = "r1";

UPDATE DataValue SET children.c1 = $r1c2c1[0] WHERE id = "r1c2";
UPDATE DataValue SET children.c2 = $r1c2c2[0] WHERE id = "r1c2";

UPDATE DataValue SET children.c1 = $r2c1[0] WHERE id = "r2";
UPDATE DataValue SET children.c2 = $r2c2[0] WHERE id = "r2";

TRAVERSE children FROM (SELECT FROM DataValue WHERE id = "r1");
```

Only the r1 element is returned. 
```JSON
{
    "result": [
        {
            "@type": "d",
            "@rid": "#40:0",
            "@version": 3,
            "@class": "DataValue",
            "children": {
                "c1": "#34:1",
                "c2": "#35:1"
            },
            "id": "r1"
        }
    ],
    "notification": "Query executed  in 0.021 sec. Returned 1 record(s)"
}
```

Is it not possible to TRAVERSE through links in a LINKMAP? I have not gone back and verified, but I thought this was possible in a previous 3.0.X version.

mmacfadden:
[...]

PR #9100, opened 2019-12-31
Added traversal through map elements.

mmacfadden:
Previously a LINKLIST would be TRAVERSED but not a LINK Map. See issue #9099. This definitely needs to be reviewed for correctness to make sure there are not unintended consequences, but this seems to resolve the issue. I can retarget this at the develop branch if needed.


luigidellaquila:
Hi @mmacfadden 

Thank you very much, it makes a lot of sense.
I do not see any reason why it should have unintended side effects, so I'm merging it

Thanks

Luigi

PR #9122, opened 2020-02-03
Added possibility for invalidate pool cache and fixed connection leak

WeaxMe:
Added possibility for invalidate pool cache


WeaxMe:
I fixed connection leak in 051327c

### How reproduce
1. Schedule event 
2. Database connection uses only when task runs. In other time database connection is open, but not used. Database connection uses only when event runs, other time it doesn't need.

This changes requires implement [ODatabaseThreadLocalFactory](https://github.com/orientechnologies/orientdb/blob/47a44a54f858982273b74636cb02bbbf3f67c0b3/core/src/main/java/com/orientechnologies/orient/core/db/ODatabaseThreadLocalFactory.java) in each application which uses OScheduler

WeaxMe:
Please see my changes

PhantomYdn:
@laa, @wolf4ood , @tglman , could you please help with merging this PR?

tglman:
Hi,

This looks good for me, but here: https://github.com/orientechnologies/orientdb/pull/9122/files#diff-6638c025443f4d6b83bc4d45cbd630fb i do not understand what's changed in this file and if actually need to be changed, could you remove this change if is not needed?

Regatds

WeaxMe:
Hello,
https://github.com/orientechnologies/orientdb/pull/9122/files#diff-6638c025443f4d6b83bc4d45cbd630fb - this file was changed for use new query API. Previously it used old API `database.query(new OSQLSynchQuery<Object>())`. I used new API `database.query(String)`.

Actually this change doesn't change anything.

WeaxMe:
@tglman I removed changes for RestrictedTest

core/src/main/java/com/orientechnologies/orient/core/schedule/OScheduledEvent.java
```diff
@@ -156,20 +119,25 @@ public Date getStartTime() {
...
+      interrupt();
+    }
+    internalSchedule(new ScheduledTimerTask(this));
     return this;
   }
 
+  private void internalSchedule(ScheduledTimerTask timerTask) {
+    synchronized (this) {
+      nextExecutionId.incrementAndGet();
+      timer = Orient.instance().scheduleTask(timerTask, cron.getNextValidTimeAfter(new Date()), 0);
```
tglman: could you schedule the using the scheduler in the OrientDBInternal instance instead on the Orient singlethon?
as well be aware that the scheduleTask in the singlethon get a runnable and return a TimerTask, so no need to pass in a TimerTask itself, in the new Implementation you need to pass in the TimerTask, as well could you use the `OrientDBInternal.execute(db,user, function)`  for executing the logic (with "admin" user for now) ? this will remove the need of the ThreadLocalFactory.

WeaxMe: Sure. I will do this today or tommorrow

WeaxMe:
Hello.
I added changes. Please review

PR #9147, opened 2020-02-20
Wip begin transaction

tglman:
This add the the logging of an metadata from the transaction to the WAL, and recover on open/restore and allow to access the last logged metadata, the current implementation do not support concurrent operations for the transactions metadata, there is just a callback to let the high level know that the current metadata is logged so a newer transaction can be sent, this is more then enough for what is needed today.


core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OAbstractPaginatedStorage.java
```diff
@@ -5159,8 +5186,11 @@ private OLogSequenceNumber restoreFromWAL() throws IOException {
         if (!fuzzyCheckPointIsComplete) {
           OLogManager.instance().warn(this, "FUZZY checkpoint is not complete.");
 
-          final OLogSequenceNumber previousCheckpoint = ((OFuzzyCheckpointStartRecord) checkPointRecord.get(0))
-              .getPreviousCheckpoint();
+          OFuzzyCheckpointStartRecord checkpointStartRecord = (OFuzzyCheckpointStartRecord) checkPointRecord.get(0);
```
laa: Quite smart :-)

PR #9183, opened 2020-03-21
Changed comparison to be ge rather than gt

matanshukry:
Currently it will result in an exception and a warning in the database, so it's not that bad, but unless there's a special reason I'm missing it should be `>=`

Warning that occurs without this fix:
`java.lang.IndexOutOfBoundsException: Index: 0, Size: 0`


luigidellaquila:
Thanks!

Luigi

PR #9203, opened 2020-04-02
NPE in write buffer capacity limit during int type overflow

dritter-sap:
During construction of OCASDiskWriteAheadLog, method log(new OEmptyWALRecord()) was called before write buffers writeBufferOne and writeBufferTwo were initialized. In the case that maxCacheSize is smaller than qsize (e.g. after integer type overflow like maxPagesCacheSize * pageSize with maxPagesCacheSize = Integer.MAX). This led to a subsequent doFlush and caused a NPE. Added a test as well.


core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/paginated/wal/cas/OCASDiskWriteAheadLog.java
```diff
@@ -103,7 +103,7 @@
   private final AtomicLong logSize     = new AtomicLong();
   private final AtomicLong queueSize   = new AtomicLong();
 
-  private final int maxCacheSize;
+  protected final int maxCacheSize;
```
laa, line 106: Hi, why did you change visibility?

dritter-sap:
As the one in the develop branch was merged, this one is a downport of the develop one.

PR #9205, opened 2020-04-02
stored the last metadata in the storage configuration 

tglman:
add the last metadata in storage configuration

remove the read of last metadata from the WAL.


tglman:
if a restore procedure was ran the `lastMetadata` would have a value, from the WAL. so the check on null will be false

tglman:
already merged manually closing this

Issue #9200, opened 2020-04-02
Exception in commit when Live Query running


pkendall64:
## Expected behavior  
Adding edges should work without failure when a live query is running

## Actual behavior  
We have a live query in progress and read a vertex from the database that has some edges to other vertices. We then add another vertex and an edge to this already existing vertex that has other edges and we get the following exception.

```
	Suppressed: com.orientechnologies.orient.core.exception.OSerializationException: Cannot read transaction record from the network. Transaction aborted
	DB name="test"
		at com.orientechnologies.orient.server.tx.OTransactionOptimisticServer.begin(OTransactionOptimisticServer.java:208)
		at com.orientechnologies.orient.core.db.document.ODatabaseDocumentAbstract.rawBegin(ODatabaseDocumentAbstract.java:1243)
		at com.orientechnologies.orient.server.OConnectionBinaryExecutor.executeCommit37(OConnectionBinaryExecutor.java:1269)
		at com.orientechnologies.orient.client.remote.message.OCommit37Request.execute(OCommit37Request.java:124)
		at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.sessionRequest(ONetworkProtocolBinary.java:310)
		at com.orientechnologies.orient.server.network.protocol.binary.ONetworkProtocolBinary.execute(ONetworkProtocolBinary.java:212)
		at com.orientechnologies.common.thread.OSoftThread.run(OSoftThread.java:69)
	Caused by: com.orientechnologies.orient.core.exception.ODatabaseException: Error on saving record #57:0
	DB name="test"
		at com.orientechnologies.orient.server.tx.OTransactionOptimisticServer.addRecord(OTransactionOptimisticServer.java:442)
		at com.orientechnologies.orient.server.tx.OTransactionOptimisticServer.begin(OTransactionOptimisticServer.java:158)
		... 6 more
	Caused by: java.lang.IllegalArgumentException: Invalid property value for OResult: [#25:0] - com.orientechnologies.orient.core.db.record.ridbag.ORidBag
		at com.orientechnologies.orient.core.sql.executor.OResultInternal.checkType(OResultInternal.java:67)
		at com.orientechnologies.orient.core.sql.executor.OResultInternal.setProperty(OResultInternal.java:40)
		at com.orientechnologies.orient.core.query.live.OLiveQueryHookV2.calculateBefore(OLiveQueryHookV2.java:232)
[...]

PR #9208, opened 2020-04-03
unbox Ridbags in live query hook

pkendall64:
This fixes an exception in the query hook because an OResult property cannot be assigned an ORidBag
This fixes issue #9200 (Exception in commit when Live Query running) and provides a unit test


luigidellaquila:
Hi @pkendall64 

That's perfect, thank you very much!

Luigi

Issue #9229, opened 2020-04-26
Many just plain wrong assumptions in ONative#findMemoryGCRoot


joekhoobyar:
## Expected behavior  
Many things are not working as expected:

- The `DEFAULT_MEMORY_CGROUP_PATH` should be a sensible default for most modern linux systems: `/sys/fs/cgroup/memory`
- The `fstype` field should be looked for in the *3rd* field of `/proc/mounts`
- The logic in `findMemoryGCRoot` should help prevent OOM errors in dockerized Java installations, not be the root cause of them.

## Actual behavior  

- The `DEFAULT_MEMORY_CGROUP_PATH` is an ancient default: `/sys/fs/memory`
- The `fstype` field is *incorrectly* looked for in the *1st* field of `/proc/mounts`
- The logic in `findMemoryGCRoot` causes OOM errors in dockerized Java installations.

## Steps to reproduce  
```
- Path to 'memory' cgroup is '/kubepods/burstable/pode5c1e1d9-9a54-490b-bd41-4f6bc3c375c1/92732085684eb5b85725b9fab1d9417802e0e295f073e0efac7cb2f3178fd878' 
[nexus-688dc7bd98-mcl9w nexus] 2020-04-26 20:33:09,849+0000 INFO  [FelixStartLevel] *SYSTEM com.orientechnologies.common.jna.ONative - Mounting path for memory cgroup controller is '/sys/fs/memory' 
[nexus-688dc7bd98-mcl9w nexus] 2020-04-26 20:33:09,849+0000 INFO  [FelixStartLevel] *SYSTEM com.orientechnologies.common.jna.ONative - Can not find '/sys/fs/memory/kubepods/burstable/pode5c1e1d9-9a54-490b-bd41-4f6bc3c375c1/92732085684eb5b85725b9fab1d9417802e0e295f073e0efac7cb2f3178fd878' path for memory cgroup, it is supposed that process is running in container, will try to read root '/sys/fs/memory' memory cgroup data 
[nexus-688dc7bd98-mcl9w nexus] 2020-04-26 20:33:09,849+0000 INFO  [FelixStartLevel] *SYSTEM com.orientechnologies.common.jna.ONative - Can not read memory soft limit for cgroup '/sys/fs/memory' 
[...]

PR #9230, opened 2020-04-26
Fixes detection of memory cgroups

joekhoobyar:
Fixes incorrect detection of memory cgroups on modern Linux distributions.


joekhoobyar:
Fixes #9229

tglman:
Hi, 
This looks good for me, @laa would this change break some compatibility with the past ? shall we do some check on multiple paths ?

Bye

laa:
Sure looks very good. Thank you !

PR #9243, opened 2020-05-07
First version of key normalizers

dritter-sap:
First normalizer versions for all supported data types, with tests and benchmarks comparing to the previsously used serializers. The new normalizers use ICU lib for String normalization, which adds a performance penalty compared to the previous serializers.
The `BigDecimal` type still loses precision due to `double` conversion.
Next steps: NULLS handling, compaction of NULL byte to bit, and building a new index with normalized keys.

Fixed broken `OUnsafeByteArrayComparator` by adding a corrected version `OUnsafeByteArrayComparatorV2.java` to ensure compatibility of the existing persistence.


dritter-sap:
Thanks for the review and feedback. The renamings slipped into the commits. I actually wanted to remove the files completely, because they seem to be outdated and did not know, who is still using them. But I left them in for now.
Changed to Camel-case naming conventions in tests and benchmarks.

PR #9246, opened 2020-05-12
minor lucene analyzer, builder changes for readability

dritter-sap:
Mainly changes to analyzer, builder et al. for readability and some tests added.


wolf4ood:
@dritter-sap looks great! 
Thanks

PR #9253, opened 2020-05-16
Added interval function

matanshukry:
I needed an Interval function, much like the one exist in MySQL ( https://www.w3resource.com/mysql/comparision-functions-and-operators/interval-function.php ), so I've created one.

It includes a bit of explanation at the top of the function class and a test.


luigidellaquila:
Hi @matanshukry 

the PR seems good, the only little glitch is in the javadocs: I think the first example is wrong

```
interval(43, 35, 5, 15, 50) = 1
```

it should return 3, right?

Thanks

Luigi

matanshukry:
@luigidellaquila yeah my bad. There's even at set with the same arguments that is correct and using 3. Anyway fixed it

PR #9258, opened 2020-05-21
Fixed NPE when passing any "bad value"

matanshukry:


matanshukry:
Would love if someone can also backport it into 3.0.X releases ?

PR #9269, opened 2020-05-28
Fix eval with result and variable

### LINKED ISSUE #9240: Difference in LET variables initialization between 2.2.24 and 3.0.30
### OS: Windows 10  

We are migrating one of our applications from OrientDB 2.2.24 to 3.0.30.

The following query returns different results depending on the OrientDB version :

## Expected behavior  
OrientDB 2.2.24:
```
select $rank as rank id from Report let $rank = eval('$rank + 1')

+----+----+----+
|#   |rank|id  |
+----+----+----+
|0   |1   |rep1|
|1   |2   |rep2|
+----+----+----+

2 item(s) found. Query executed in 0.079 sec(s).
```
## Actual behavior  
OrientDB 3.0.30:
```
select $rank as rank, id from Report let $rank = $rank+1

+----+----+----+
|#   |rank|id  |
+----+----+----+
|0   |1   |rep1|
|1   |1   |rep2|
+----+----+----+

2 item(s) found. Query executed in 0.002 sec(s).
```

## Steps to reproduce  
See previous queries

matanshukry:
- OFunctionCall:isEarlyCalculate - if eval function, check if string
contains the varName.
- Let expression step: save the result in the context (where it will be
taken again in the next row)
- isEarlyCalculate: also get the variable name the result is put in.

resolves #9240 

Notes:
- I've looked and tested what I can, but I'm not familiar enough with the area of the code changes, so a thorough review should be done to verify this code is valid in all cases

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/executor/LetExpressionStep.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/executor/LetExpressionStep.java
@@ -40,6 +40,7 @@ public class LetExpressionStep extends AbstractExecutionStep {
       public OResult next() {
         OResultInternal result = (OResultInternal) source.next();
         Object value = expression.execute(result, ctx);
+        ctx.setVariable(varname.getStringValue(), value);
         result.setMetadata(varname.getStringValue(), value);
         return result;
       }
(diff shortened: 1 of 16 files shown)


tglman:
hi @matanshukry,

This look ok, anyway there is no test case that prove the case, could you add one?

Bye

matanshukry:
@tglman .. It's been 4 months.

I'm afraid I don't have time for this now.

lvca:
@luigidellaquila WDYT?

andrii0lomakin:
Closing this PR because it is out of date. If you are still interested to provide contributions please open a new one which is based on to date sources for develop the branch.

PR #9297, opened 2020-06-15
update hazelcast and plugin to enable hazelcast on k8s

pxsalehi:
This will make it possible to deploy OrientDB on K8s in distributed mode by defining the environment variable `ENABLE_KUBERNETES_DISCOVERY` for the pod. All pods with the same label name/value (specified in hazelcast.xml) will form one cluster.

### DIFF
--- a/distributed/pom.xml
+++ b/distributed/pom.xml
@@ -37,7 +37,7 @@
     <name>OrientDB Distributed Server</name>
 
     <properties>
-        <hz.version>3.10.6</hz.version>
+        <hz.version>3.11.7</hz.version>
         <javac.src.version>1.6</javac.src.version>
         <javac.target.version>1.6</javac.target.version>
         <jar.manifest.mainclass>com.orientechnologies.orient.server.OServerMain</jar.manifest.mainclass>
@@ -134,6 +134,11 @@
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast</artifactId>
             <version>${hz.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.hazelcast</groupId>
+            <artifactId>hazelcast-kubernetes</artifactId>
+            <version>1.5.3</version>
         </dependency>
     </dependencies>
 
--- a/distributed/src/main/java/com/orientechnologies/orient/server/hazelcast/OHazelcastPlugin.java
+++ b/distributed/src/main/java/com/orientechnologies/orient/server/hazelcast/OHazelcastPlugin.java
@@ -21,6 +21,7 @@ package com.orientechnologies.orient.server.hazelcast;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.config.FileSystemXmlConfig;
+import com.hazelcast.config.JoinConfig;
 import com.hazelcast.core.EntryEvent;
 import com.hazelcast.core.EntryListener;
 import com.hazelcast.core.Hazelcast;
@@ -36,8 +37,6 @@ import com.hazelcast.core.Member;
 import com.hazelcast.core.MemberAttributeEvent;
 import com.hazelcast.core.MembershipEvent;
 import com.hazelcast.core.MembershipListener;
-import com.hazelcast.config.JoinConfig;
-import com.hazelcast.config.NetworkConfig;
 import com.hazelcast.spi.exception.RetryableHazelcastException;
 import com.orientechnologies.common.concur.OOfflineNodeException;
 import com.orientechnologies.common.concur.lock.OInterruptedException;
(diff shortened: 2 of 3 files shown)


pxsalehi:
It can wait. No problem on my side. Probably better to keep it for 3.2.

dritter-sap:
Thanks, could you kindly provide the simplest possible automatic test (or at least a manual one).

It might be possible to make the configuration automatic by checking environment variables for `KUBERNETES_SERVICE_HOST` or similar, check k8s environment variables.

pxsalehi:
If you only want to test the code added by the PR, the simplest is this:
`ENABLE_KUBERNETES_DISCOVERY=true ./dserver.sh `

As the title of the PR suggests, these are only the minimal changes required to be able to use Hazelcast on Kubernetes. It doesn't mean OrientDB is 100% working on Kubernetes. For that we need to run all our tests on Kubernetes as well. That requires quite some modifications, especially to the distributed IT suite, which is something I can work on but will take a while.

I have manually tried out also creating an OrientDB cluster on Kubernetes. But at that stage, I'm mostly testing Hazelcast on Kubernetes not OrientDB! You'd need a cluster (minikube will work probably), and you should create two or more pods. You'd also need a clusterRoleBinding (it would probably be better to follow the Hazelcast Kubernetes page) but here are some manifests I used: https://github.com/pxsalehi/orientdb/tree/k8s/distributed/src/test/resources/kubernetes

As @luigidellaquila mentioned before, without the PR OrientDB 3.1. is guaranteed NOT to run on Kubernetes. But again, it is perfectly fine for me, we can wait for 3.2 to upgrade hazelcast.

I don't understand the value of hard-coding many different environment variables. All we need is a flag to tell the containerized OrientDB that it is running on Kubernetes. If at some point, we need to also pass the cloud provider (for whatever reason) we can add more.

dritter-sap:
I agree to what @luigidellaquila mentioned about the versions, which was out of the question for me. It is also clear that you did not fully enable k8s for ODB, as also discussed before.

A test (even just a new hazelcast configuration) usually does not only run the software, but asserts certain conditions. 

The idea with the environment was about a better usability. If we don't want or think that we don't need it now then let's keep it in mind for later.

distributed/src/main/java/com/orientechnologies/orient/server/hazelcast/OHazelcastPlugin.java
```diff
@@ -844,7 +845,28 @@ protected HazelcastInstance configureHazelcast() throws FileNotFoundException {
         .setMergePolicy(OHazelcastMergeStrategy.class.getName());
     // Disabled the shudown hook of hazelcast, shutdown is managed by orient hook
     hazelcastConfig.setProperty("hazelcast.shutdownhook.enabled", "false");
-
+    boolean enableKubernetesDiscoveryEnvVar = false;
+    String enableKubernetesDiscoveryEnvVarStr = System.getenv("ENABLE_KUBERNETES_DISCOVERY");
```
tglman, line 849: Hi,

Is this really  needed ? as far as I understood this is just overriding the settings in the hazelcast.xml, I see two possible alternative solution to this, one is this: https://hazelcast.com/blog/xml-variables/ (if this version of HZ support it), and another is use OGlobalCofiguration for the variable, that's where all the variable a configured in OrientDB

distributed/pom.xml
```diff
@@ -37,7 +37,7 @@
     <name>OrientDB Distributed Server</name>
 
     <properties>
-        <hz.version>3.10.6</hz.version>
+        <hz.version>3.11.7</hz.version>
```
tglman, line 40: I think we can update also up to 3.12.x, we may have some additional feature there tu use as well

distributed/pom.xml
```diff
@@ -135,6 +135,11 @@
             <artifactId>hazelcast</artifactId>
             <version>${hz.version}</version>
         </dependency>
+        <dependency>
```
tglman, line 138: Is this a compile time dependency or a runtime dependency? HZ hast a tons of different plugins, i would prefer to avoid to add all of them as dependency if is not needed, but yes not having this as the dependency may add additional steps to the deploy procedure on kubernates, not strongly against this, but if is not needed i would prefer to keep the dependency out.

distributed/pom.xml
pxsalehi: When I wanted to choose a version, I wanted to find a version that has Kubernetes support and doesn't require changes on our side. I tried out also 3.12.x, but that requires some changes in our code. I think not that much, but still we need to change something. I forgot exactly what.
So, if you think 3.12 has something we need, I can try it again, and also see what needs to change.

distributed/src/main/java/com/orientechnologies/orient/server/hazelcast/OHazelcastPlugin.java
pxsalehi: well, we need a way to "configure" the dockerized orientDB to use kubernetes discovery. This works. The other two options might be more consistent with the rest of our configs. I'll check.

PR #9302, opened 2020-06-16
Use Object for keys

pxsalehi:


pxsalehi:
@tglman Please have a look and let me know if there needs to be any changes.

tglman:
Hi @pxsalehi,

Looks nice, I think there is only a missing case, that are the composite key (and it seems that case is also missed in remote), you could try the case doing a IT test case very similar to `SimpleIndexDistributedIT` but with a unique composite index, this should give you an highlight on how composite keys are modelled in the transaction, on top of my memory in the key should be there an instance of `OCompositeKey` of witch inside has a list of keys that also could be composite keys, or just all the plain values that could be handled by `OType`, but as I said double check it with an IT test.

Done this i think is good do go, for the remote case we may handle that in future as well.
Regards

PR #9312, opened 2020-06-24
minor optimistic transaction test changes

dritter-sap:
Checking optimistic transaction tests with minor changes.
Added fix for global dump issue + test


dritter-sap:
CI passed, but does not update the PR’s status.

PR #9324, opened 2020-07-06
Update to Hazelcast 3.12.8

pxsalehi:
I think the maven dependency for hazelcast-kubernetes need to be there.

### DIFF
--- a/distributed/pom.xml
+++ b/distributed/pom.xml
@@ -139,6 +139,7 @@
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast-kubernetes</artifactId>
             <version>1.5.3</version>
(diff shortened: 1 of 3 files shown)


distributed/pom.xml
```diff
@@ -135,6 +135,11 @@
             <artifactId>hazelcast</artifactId>
             <version>${hz.version}</version>
         </dependency>
+        <dependency>
```
tglman, line 138: Could we have this as optional dependency ? 
Is not needed for everyone that do not run kubernates to have kubernates :)

pxsalehi: As I wrote in the PR text, I don't know a way. It needs to be in the image.

PR #9332, opened 2020-07-09
Run distributed integration tests on Kubernetes - Part1

pxsalehi:
This is a ~draft~ PR to discuss the approach to run distributed integration tests on Kubernetes. Details are open for discussion if there is a reason to change them. The main effort of this version is minimizing changes required to the tests and allowing the tests to be easily runnable on both a local setup (as it is now) or on a Kubernetes cluster. 

`BasicSyncIT` is the only test adapted for now, as an example of the setup changes required. I can gradually make changes to other tests as well. While running the test on a local setup can be done via:
`mvn test -Dtest=BasicSyncIT#sync`
Running the same test on an available Kubernetes cluster can be done with:
`mvn test -Dtest=BasicSyncIT#sync -P it-k8s -DkubeConfig=/path/to/kubeconfig -DorientdbDockerImage=orientdb:3.2.0`

Which will use the `default` namespace. The namespace can be changed via:
`-DkubernetesNamespace=namespace`
The namespace must already exist.

Essentially, all that is needed is providing a Kubeconfig. If the `kubeConfig` system property is not set, a local setup is used for the test. The docker image needs to be provided since our latest docker image doesn't support Hazelcast on Kubernetes. You'd need to build the image from `develop`. I've used the Dockerfile in the `distribution` module.

Unfortunately, the two common Kubernetes Java clients both seem to have problem using Java8. So you should run the test with a newer JDK version. I'm using 11.


pxsalehi:
@tglman: I have changed the PR so it includes no changes to any existing test. Only setup/config code.

PR #9349, opened 2020-07-27
Regression on SQL range queries with indexes

dritter-sap:
Fixes broken range query, e.g., `select * from class where something >= "another"` due to unsupported NULL values.


dritter-sap:
Code coverage +0.04%, but still below 5x%

PR #9430, opened 2020-11-02
Add command timeout and force kill

luigidellaquila:
Gentle for the storage (avoid to kill the storage in critical paths) but general enough to also cover script execution (eg. Gremlin)

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
@@ -126,13 +126,19 @@ public class ODatabaseDocumentEmbedded extends ODatabaseDocumentAbstract
   private OrientDBConfig config;
   private OStorage storage;
 
-  boolean interruptingCommand = false;
-  InterruptTimerTask commandInterruptTimer = new InterruptTimerTask();
+  InterruptTimerTask commandInterruptTimer;
 
   class InterruptTimerTask extends TimerTask {
+
+    Thread executionThread;
+
+    InterruptTimerTask(Thread executionThread) {
+      this.executionThread = executionThread;
+    }
+
     @Override
     public void run() {
-      ODatabaseDocumentEmbedded.this.interruptingCommand = true;
+      interruptExecution(executionThread);
     }
   }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
```diff
@@ -133,6 +126,22 @@
   private OrientDBConfig config;
   private OStorage storage;
 
+  InterruptTimerTask commandInterruptTimer;
+
+  class InterruptTimerTask extends TimerTask {
+
+    Thread executionThread;
```
tglman: Can we have the private/protected visibilities as needed

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
```diff
@@ -661,27 +683,40 @@ public OResultSet execute(String language, String script, Map<String, ?> args) {
...
-        sharedContext
-            .getOrientDB()
-            .getScriptManager()
-            .getCommandManager()
-            .getScriptExecutor(language);
-    OResultSet original;
-
-    ((OAbstractPaginatedStorage) this.storage.getUnderlying())
-        .pauseConfigurationUpdateNotifications();
+    if (OGlobalConfiguration.COMMAND_TIMEOUT.getValueAsInteger() > 0) {
```
tglman: instead of: `OGlobalConfiguration.COMMAND_TIMEOUT.getValueAsInteger() ` we should use ` getContextConfiguraion().getValueAsInteger(OGlobalConfiguration.COMMAND_TIMEOUT) `

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
```diff
@@ -661,27 +683,40 @@ public OResultSet execute(String language, String script, Map<String, ?> args) {
...
+      commandInterruptTimer = new InterruptTimerTask(Thread.currentThread());
+      getSharedContext()
+          .getOrientDB()
+          .scheduleOnce(
+              commandInterruptTimer, OGlobalConfiguration.COMMAND_TIMEOUT.getValueAsInteger());
```
tglman: Same here use getContextConfiguration

core/src/main/java/com/orientechnologies/orient/core/sql/executor/OSelectExecutionPlanner.java
```diff
@@ -113,6 +73,10 @@ private void init() {
     info.limit = this.statement.getLimit();
     info.lockRecord = this.statement.getLockRecord();
     info.timeout = this.statement.getTimeout() == null ? null : this.statement.getTimeout().copy();
+    if (info.timeout == null && OGlobalConfiguration.COMMAND_TIMEOUT.getValueAsLong() > 0) {
```
tglman: should use the context configuration here not the static value as well

core/src/main/java/com/orientechnologies/orient/core/sql/executor/OSelectExecutionPlanner.java
```diff
@@ -113,6 +73,10 @@ private void init() {
...
+      info.timeout = new OTimeout(-1);
+      info.timeout.setVal(OGlobalConfiguration.COMMAND_TIMEOUT.getValueAsLong());
```
tglman: should use the context configuration here not the static value as well

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OStorageInterruptionManager.java
```diff
@@ -0,0 +1,44 @@
+package com.orientechnologies.orient.core.storage.impl.local;
+
+import com.orientechnologies.common.concur.lock.OInterruptedException;
+import com.orientechnologies.common.log.OLogManager;
+
+public class OStorageInterruptionManager {
+
+  int depth = 0;
```
tglman: private/protected visibility here needed

tglman (changes requested):
I see this is integrated only in the execute methods i think we should integrate this also in query and execute, that actually can use functions that may have similar problems of long executions.

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
tglman: In the database actually the configuration is in the `getConfiguration` not in `getContextConfiguration` so it should be: `getConfiguration().getValueAsInteger(OGlobalConfiguration.COMMAND_TIMEOUT)`

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
luigidellaquila: Yep, definitely. The commands have their own management of timeout (see below, OSelectExecutionPlanner) but there can still be long running scripts. 
So yes, it makes sense

core/src/main/java/com/orientechnologies/orient/core/storage/impl/local/OStorageInterruptionManager.java
luigidellaquila: Done

core/src/main/java/com/orientechnologies/orient/core/sql/executor/OSelectExecutionPlanner.java
luigidellaquila: Done

core/src/main/java/com/orientechnologies/orient/core/sql/executor/OSelectExecutionPlanner.java
luigidellaquila: Done

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
luigidellaquila: Done

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
luigidellaquila: Done

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentEmbedded.java
luigidellaquila: Done

luigidellaquila:
OK @tglman, I did the "technical" fixes. 
What about the general approach? Is there anything better we can do?

tglman (approved):

andrii0lomakin:
Looks good.

andrii0lomakin (approved):

Issue #9421, opened 2020-10-26
ORecordLazyList.get(index) started to return ORecordId instead of ODocument


PhantomYdn:
### OS: Windows, Linux

Some of our tests started to fail after switching from 3.1.3 to 3.1.4 due to ClassCastException.
After review we have figured out that problem is in ORecordLazyList: first item is ODocument, but all subsequent are just ORecordId (but a corresponding  ODocument does exist)

## Expected behavior  
ORecordLazyList should return ODocument

## Actual behavior  
All items after first one are ORecordId

## Steps to reproduce  
Not fully sure how to reproduce from scratch. The following our testcase stopped working: https://github.com/OrienteerBAP/Orienteer/blob/future/orienteer-notification/src/test/java/org/orienteer/notifications/test/TestNotificationLifecycle.java

PhantomYdn:
@tglman `git bisect` between 3.1.3 and 3.1.4 show that the issue was introduced by the following commit: https://github.com/orientechnologies/orientdb/commit/6265642427def0a8c474c66e1619fe7b658c8815

PhantomYdn:
@tglman, @laa, @luigidellaquila, OrientDB team, do you have ideas how to fix that? 3.1.4 is unusable due to this issue...

PhantomYdn:
Apparently, the reason is in fact that mentioned above commit doesn't affect logic of `iterator()`. Defect is pretty visible in the following testcase:

```java
@Test
public void testLinkListAndODocumentConversion()
{
	ODatabaseSession db = wicket.getTester().getDatabaseSession();
	OSchemaHelper helper = OSchemaHelper.bind(db);
	helper.oClass("TestODocumentConverstion")
			.oProperty("name", OType.STRING)
			.oProperty("items", OType.LINKLIST);
	ODocument mainDoc = helper.oDocument("name", "document").saveDocument().getODocument();
	helper.oClass("TestODocumentConverstionItems")
		  	.oProperty("name", OType.STRING);
	ODocument doc1 = helper.oDocument("name", "item1").saveDocument().getODocument();
	ODocument doc2 = helper.oDocument("name", "item2").saveDocument().getODocument();
	ODocument doc3 = helper.oDocument("name", "item3").saveDocument().getODocument();
	List<ODocument> items = Arrays.asList(doc1, doc2, doc3);
	mainDoc.field("items", items);
	mainDoc.save();
	db.commit();
	db.begin();
[...]

PR #9437, opened 2020-11-12
Fix issue #9421: broken conversion within ORecordLazyList

PhantomYdn:
Issue #9421 is pretty annoying for our project. @tglman, @laa, @luigidellaquila, please, please, help with merging to `develop`.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/db/record/ORecordLazyList.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/db/record/ORecordLazyList.java
@@ -155,12 +155,14 @@ public class ORecordLazyList extends OTrackedList<OIdentifiable> implements ORec
   @Override
   public boolean add(OIdentifiable e) {
     preAdd(e);
+    lazyLoad(true);
     return super.add(e);
   }
 
   @Override
   public void add(int index, OIdentifiable e) {
     preAdd(e);
+    lazyLoad(true);
     super.add(index, e);
   }
   
@@ -180,7 +182,6 @@ public class ORecordLazyList extends OTrackedList<OIdentifiable> implements ORec
         e = e.getIdentity();
       else contentType = ORecordMultiValueHelper.updateContentType(contentType, e);
     }
-    lazyLoad(true);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


PhantomYdn:
Also let me mention that there was a difference between `public void add(int index, OIdentifiable e)` and `public boolean add(OIdentifiable e)`: last one did not invoke `ORecordInternal.track(sourceRecord, e);`. If I'm right - that was also a mistake. If I'm not right and in case `public boolean add(OIdentifiable e)` tracking should not be invoked (so it was OK) - please let me know.

PhantomYdn:
Travis CI failed due to some other floating problem: it was observed on previous builds as well. For example: https://travis-ci.org/github/orientechnologies/orientdb/builds/739208716

Locally all tests are good.

core/src/main/java/com/orientechnologies/orient/core/db/record/ORecordLazyList.java
```diff
@@ -154,20 +154,23 @@ public boolean contains(final Object o) {
...
 
   @Override
   public void add(int index, OIdentifiable e) {
+    preAdd(e);
+    super.add(index, e);
+  }
+  
+  @Override
+  public boolean addInternal(OIdentifiable e) {
+	preAdd(e);
```
tglman, line 171: in this case the lazy load should not run, so is not correct to use preAdd as it is here.

tglman (changes requested):
Looks correct, it just need to remove the lazy load logic in addInternal, and make sure the formatting is correct.

core/src/main/java/com/orientechnologies/orient/core/db/record/ORecordLazyList.java
PhantomYdn: @tglman , that's actually the actual fix. Without it - my testcase will fail.

PhantomYdn: Without calling `preAdd(a)` - list will not update `contentType` and first iterator will switch `contentType` from EMPTY to ALL_RECORDS which prevents the conversion of the rest of the items in the list.

PhantomYdn: Next commit is fixing this

PhantomYdn:
@tglman , We really needs this in 3.1.5. Please consider merging before releasing the next version.

tglman (approved):

PR #9439, opened 2020-11-14
Fix OETLProcessorConfigurator SPI by returning createDefaultContext()

PhantomYdn:
Small fix of SPI of OETLProcessorConfigurator to avoid breaking of backward compatibility


luigidellaquila:
Merged, thanks!

Luigi

Issue #9432, opened 2020-11-02
OFunction scripting stopped working


PhantomYdn:
### OS: Ubuntu  

Create OFunction with the following code:

```java
java.lang.System.out.println("Test")
```

## Expected behavior  
Execution of this function should be successful with corresponding message in logs or console.

## Actual behavior  
Exception happened:
```
java.lang.ClassNotFoundException: java.lang.System.out.println

Root cause:

java.lang.ClassNotFoundException: java.lang.System.out.println
     at jdk.nashorn.internal.runtime.NativeJavaPackage.classNotFound(NativeJavaPackage.java:162)
     at jdk.nashorn.internal.scripts.Script$Recompilation$75$25$\^eval\_.handleContent(<eval>:2)
     at jdk.nashorn.internal.runtime.ScriptFunctionData.invoke(ScriptFunctionData.java:637)
     at jdk.nashorn.internal.runtime.ScriptFunction.invoke(ScriptFunction.java:494)
     at jdk.nashorn.internal.runtime.ScriptRuntime.apply(ScriptRuntime.java:393)
     at jdk.nashorn.api.scripting.ScriptObjectMirror.callMember(ScriptObjectMirror.java:199)
     at jdk.nashorn.api.scripting.NashornScriptEngine.invokeImpl(NashornScriptEngine.java:386)
     at jdk.nashorn.api.scripting.NashornScriptEngine.invokeFunction(NashornScriptEngine.java:190)
     at com.orientechnologies.orient.core.command.script.OCommandExecutorFunction.executeInContext(OCommandExecutorFunction.java:99)
     at com.orientechnologies.orient.core.metadata.function.OFunction.executeInContext(OFunction.java:161)
     at com.orientechnologies.orient.core.metadata.function.OFunction.execute(OFunction.java:131)
```

## Notes
It worked perfectly in OrientDb 2.x

PhantomYdn:
"Issue was introduced' by the following commit: https://github.com/orientechnologies/orientdb/commit/40544dc82105884de0c78a893a84439871029e23

Is there any documentation on how to use these security scripts?

PhantomYdn:
Btw, there are few comments to this implementation:
1) Assumed that user can define list of allowed packages in OScriptManager, but according to code: it's not packages, but classNames - which makes configuration much more granular and inconvinient. Will be good to have either on package level or use wildcards.
2) There is no way to disable this behavior. It would be nice to allow users to decide: use or not.
[...]

PR #9440, opened 2020-11-16
Using GraalVM instead of Nashorn

lvca:
Usage of the new GraalVM instead of Nashorn for JS and Polyglot script execution.


lvca:
I've used the wrong formatter, updating this with the right formatter (the one from google java)

lvca:
Ok, the last test passes. I've replaced the GraalVM JSR223 compatible API with the native ones. They work better and support more complex return types.

PhantomYdn:
Please consider comments from #9432

core/src/main/java/com/orientechnologies/orient/core/command/script/OScriptManager.java
```diff
@@ -87,10 +88,20 @@
...
 
+    final boolean useGraal = OGlobalConfiguration.SCRIPT_POLYGLOT_USE_GRAAL.getValueAsBoolean();
+
     executorsFactories.put(
-        "javascript", (lang) -> new OJavascriptScriptExecutor(lang, new OScriptTransformerImpl()));
+        "javascript",
+        (lang) ->
+            useGraal
+                ? new OPolyglotScriptExecutor(lang, new OScriptTransformerImpl())
+                : new OJsr223ScriptExecutor(lang, new OScriptTransformerImpl()));
```
tglman, line 98: will the the Jsr223 be "loadable" in future java releases ? we may need to automatically bail out or even provide a pluggable module for the different jvms, so far this can be good as first step, but we may need to move the jsr223 in a external module if create problems with recent jvms.

core/src/test/java/com/orientechnologies/orient/core/command/script/JSScriptTest.java
```diff
@@ -245,7 +258,7 @@ public void jsSandboxWithOrient() {
           db.execute("javascript", "Orient.instance().getScriptManager().addAllowedPackages([])")) {
         Assert.assertEquals(1, resultSet.stream().count());
       } catch (Exception e) {
-        Assert.assertEquals(e.getCause().getClass(), ScriptException.class);
```
tglman, line 248: Here we are changing the ScriptException to PolyglotException, this change the error management logic, i think would be batter remap the PolyglotException to the ScriptException for guarantee backward compatible behaviour.

tglman (changes requested):
I would prefer, to have a separate pluggable module instead of the flag, apart of that looks good, minor changes in error management required.

core/src/test/java/com/orientechnologies/orient/core/command/script/JSScriptTest.java
lvca: Good idea, pushed.

core/src/main/java/com/orientechnologies/orient/core/command/script/OScriptManager.java
lvca: It's not really clear if and when jsr223 will be completely out from the jdk. I think it's ok to keep both with a switch and when a new JDK comes out without that package we could remove it and support only the new graalvm.

lvca: I think creating 2 external modules for this is overkill right now.

tglman (approved):

PR #9473, opened 2020-12-02
Refactor OHazelcast/ODistributedAbstract (part 3)

pxsalehi:


lvca:
@pxsalehi what's the reason for this refactor?

pxsalehi:
@lvca Hi Luca. 

The refactoring for OHazelCast/ODistributed plugins are minimal and are currently limited to only breaking down some methods if necessary and moving them between these two classes. The reason is to be able to clearly separate what the OHazelcastPlugin does and limit it to cluster management and distributed metadata/configuration management. Ultimately, it would be useful to have those functionalities behind an interface to both simplify more fine-grained testing (I think currently we can only do end-to-end integration tests since the code is heavily coupled) and allow experimenting with new implementations if necessary.  

While we have talked about this in detail with @tglman, it would be great if in case you have time, you could give us some hints/suggestions/feedback for this plan, since I think you're the one who knows this part better than us! :-)

tglman (approved):

pxsalehi:
@lvca All tests are green, and I would like to merge the PR so it doesn't get too big and continue with a new PR. Please let me know if you'd like to have any changes to the PR.

lvca (approved):
Sounds good to me ;-)

PR #9544, opened 2021-02-15
Reject opening distributed DB which is not online!

pxsalehi:
Couldn't really create a test case to explicitly reproduce the stalling open issue. However, I think the fact that the `ConnectionStrategiesIT` test had to be adapted to wait for online status to be able to hit all servers, shows that sync'ing DBs now handle openDB requests better.

### DIFF
--- a/distributed/src/main/java/com/orientechnologies/orient/core/db/OrientDBDistributed.java
+++ b/distributed/src/main/java/com/orientechnologies/orient/core/db/OrientDBDistributed.java
@@ -181,11 +181,13 @@ public class OrientDBDistributed extends OrientDBEmbedded implements OServerAwar
   }
 
   private void checkDbAvailable(String name) {
-    if (OSystemDatabase.SYSTEM_DB_NAME.equals(name))
-      return;
-    ODistributedServerManager.DB_STATUS dbStatus = plugin.getDatabaseStatus(plugin.getLocalNodeName(), name);
-    if (dbStatus != ODistributedServerManager.DB_STATUS.ONLINE && dbStatus != ODistributedServerManager.DB_STATUS.BACKUP) {
-      throw new OOfflineNodeException("database " + name + " not online on " + plugin.getLocalNodeName());
+    if (OSystemDatabase.SYSTEM_DB_NAME.equals(name)) return;
+    ODistributedServerManager.DB_STATUS dbStatus =
+        plugin.getDatabaseStatus(plugin.getLocalNodeName(), name);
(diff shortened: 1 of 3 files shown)


luigidellaquila:
It makes sense, checking the code now

distributed/src/main/java/com/orientechnologies/orient/core/db/OrientDBDistributed.java
```diff
@@ -178,6 +180,30 @@ public void drop(String name, String user, String password) {
     }
   }
 
+  private void checkDbAvailable(String name) {
+    if (OSystemDatabase.SYSTEM_DB_NAME.equals(name)) return;
+    ODistributedServerManager.DB_STATUS dbStatus =
+        plugin.getDatabaseStatus(plugin.getLocalNodeName(), name);
+    if (dbStatus != ODistributedServerManager.DB_STATUS.ONLINE
+        && dbStatus != ODistributedServerManager.DB_STATUS.BACKUP) {
+      throw new OOfflineNodeException(
```
luigidellaquila, line 189: Not sure whether we should use this exception or ODistributedRedirectException.
Looking at the client, I'd say both have problems:
- ODistributedRedirectException needs an explicit node to redirect to, that could be challenging
- OOfflineNodeException removes this node from the server list, not sure it's what we want, since it's a temporary situation that will recover by itself very soon

luigidellaquila (approved):
Apart from the exception type (that is arguable and definitely fine as it is IMHO), all the rest is fine.

distributed/src/main/java/com/orientechnologies/orient/core/db/OrientDBDistributed.java
pxsalehi: Yes, I think the problem with redirent is that you need to provide an alternative, which in this case doesn't apply. For the offline exception, if you check the ConnectionStrategies test, I think it is covered. Meaning, once the DB is online, a new config is published and the clients gen update.

luigidellaquila:
Yep, it makes sense.
IMHO we can merge, then we can check if there are other cases that follow a similar pattern and

distributed/src/main/java/com/orientechnologies/orient/core/db/OrientDBDistributed.java
tglman: Even if the node is removed from the list, should be re-added when a notification from the distributed make the node available, or when there are no-more node left to connect to. So should be fine to return the  OOfflineNodeException

tglman (approved):

PR #9548, opened 2021-02-18
create a separate role for auditing

pxsalehi:
A separate 'auditor' system role, which is not assigned to any system user by default.


tglman:
This looks good, only doubt is if a user with only a auditing role need some other permission to allow him to connect to some dashboard to do the administration, all the rest is ok

luigidellaquila (approved):

PR #9601, opened 2021-05-06
Index entries refactor and optimization

luigidellaquila:
Refactor of OTransactionIndexChangesPerKey: all the fields were public, accessed in read/write mode everywhere, impossible to optimise without breaking things. 
This refactor reduces visibility of all the fields and encapsulates the get/set behaviour to allow further optimizations.

Second step is the optimization of OTransactionIndexChangesPerKey.add(), to avoid iteration of all the entries at each execution (O(N^2) on the tx size)

### DIFF
--- a/client/src/main/java/com/orientechnologies/orient/client/remote/message/OMessageHelper.java
+++ b/client/src/main/java/com/orientechnologies/orient/client/remote/message/OMessageHelper.java
@@ -313,11 +313,11 @@ public class OMessageHelper {
       network.writeInt(size);
       if (indexChange.getKeyChanges().nullKeyChanges != null) {
         network.writeByte((byte) -1);
-        network.writeInt(indexChange.getKeyChanges().nullKeyChanges.entries.size());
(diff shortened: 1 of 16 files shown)


luigidellaquila:
Some microbenchmarking:

This fix adds a 10% overhead to small transactions and in cases where there are no collisions on the keys.

On a case with many collisions, this implementation is constant time (ie. same performance as no collisions), while the previous one tends to increase the execution time.

eg.

 100 tx, 20k records per tx, 100% key collisions (ie. all the records have the same index key)
- old implementation takes ~200 seconds
- new implementation takes ~50 seconds

 100 tx, 20k records per tx, 10 different keys equally distributed in the record set 
- old implementation takes ~75 seconds
- new implementation takes ~50 seconds

 100 tx, 20k records per tx, 100 different keys equally distributed in the record set 
- old implementation takes ~50 seconds
- new implementation takes ~50 seconds

tglman:
looks good, merging

PR #9604, opened 2021-05-11
Adapting ODB export format for streaming

dritter-sap:
Adapting ODB export format for streaming, i.e. moving `@FieldTypes` after the signature and in front of the data. A stream parser will then be able to read the type definition, before it gets to parse the data itself.

- [x] Re-using the ODB's `format` settings to introduce `earlyTypes`, which will place `@fieldTypes` in front of its scope, instead of putting it behind it. This is necessary for stream parsing, which requires to have the type information to apply it during the first pass through the scope.
- [x] Adding new format and fixing existing tests
- [x] Adapt `ODatabaseExport` by introducing a new schema version (`12` -> `13`)
- [x] Check impact of new export mechanism (expected that `new` slightly slower than `old`). Simple benchmark with different types (could be further extended):
```
Benchmark                                Mode  Cnt   Score    Error  Units
JSONExportBenchmark.testExportFormatNew  avgt   15  52,459 ±  9,281  us/op
JSONExportBenchmark.testExportFormatOld  avgt   15  50,185 ± 13,272  us/op
```
- [x] Checked tests regarding different `fetch` types

**Disclaimer**
Does not contain performance improvements or stream-based export.

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/fetch/OFetchContext.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/fetch/OFetchContext.java
@@ -24,59 +24,56 @@ import com.orientechnologies.orient.core.record.impl.ODocument;
 /** @author Luca Molino (molino.luca--at--gmail.com) */
 public interface OFetchContext {
 
-  public void onBeforeFetch(final ODocument iRootRecord) throws OFetchException;
+  void onBeforeFetch(final ODocument iRootRecord) throws OFetchException;
 
-  public void onAfterFetch(final ODocument iRootRecord) throws OFetchException;
+  void onAfterFetch(final ODocument iRootRecord) throws OFetchException;
 
-  public void onBeforeArray(
+  void onBeforeArray(
       final ODocument iRootRecord,
       final String iFieldName,
       final Object iUserObject,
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/orientechnologies/orient/core/fetch/OFetchContext.java
```diff
@@ -24,59 +24,56 @@
 /** @author Luca Molino (molino.luca--at--gmail.com) */
 public interface OFetchContext {
 
-  public void onBeforeFetch(final ODocument iRootRecord) throws OFetchException;
```
luigidellaquila, line 27: A small thing: I'm sure Emanuele will complain about this ;-)
I'd say let's reduce the scope to protected if we can, otherwise let's keep them public

luigidellaquila (commented):
Apart from the small comment about public methods, the rest is good for me

core/src/main/java/com/orientechnologies/orient/core/fetch/OFetchContext.java
luigidellaquila: OK, never mind, it's an interface ;-)

PR #9610, opened 2021-05-19
Enable stream import based on new export format

dritter-sap:
From exporter version `13`, the `fieldTypes` metadata is placed before the actual data for streaming.

- [x] Adapt `ODatabaseImport` to switch to exporter version `13` using new `fromStream` method in `ORecordSerializerJSON` and `fromJSON` in `ORecordAbstract`.
- [x] Added `RidBag` stream support to `ORecordSerializerJSON`
- [x] Fix tests due to stream parsing
   - some tests had non-standard JSON
   - to fix that, a fallback to the `fromString` parser was added on `JsonParseException` (requires transactions)
- [x] Adapt existing import API for E2E streaming based on exporter version >=13.
   - [x] info
   - [x] clusters
   - [x] schema
   - [x] records
   - [x] indexes
   - [x] brokenRIDs

(ideas discarded in discussion with @luigidellaquila : no need to add transaction support during import, when falling back to old import logic in case of non-landard json (for compatibility)).

Out of scope (move to other PR or else):
 - [ ] Adapt existing import API
    - [ ] manualIndexes (probably not necessary, since stream import only works from exporter version `13` onwards and thus there should be no manual indexes exported)
 - [ ] Switch to new implementation
 - [ ] Fix tests


dritter-sap:
Jenkins CI ran into connection resets or pipeline failures. @luigidellaquila  and @dritter-sap could validate that the build passes on other computing environments.

PR #9621, opened 2021-06-06
Use non-blocking SecureRandom for embedded NETWORK_TOKEN_SECRETKEY

### LINKED ISSUE #9603: Use of SecureRandom.getInstanceStrong can cause database creation to block
## Expected behavior  
Database creation does not block on startup.

## Actual behavior  
When instantiating a database (e.g. an in-memory DB) the database creation blocks for a significant period (over 1 minute in our case).

## Steps to reproduce  
Create an embedded, in-memory, database without specifying a `NETWORK_TOKEN_SECRETKEY` configuration parameter.

## Root Cause Analysis
In 3.1, `OTokenHandlerImpl` used a `new RandomI()` to generate key bytes in the event that `NETWORK_TOKEN_SECRETKEY` was not defined in the configuration.
This was non-blocking (good) but not a secure PRNG (bad).

In 3.2 this has been changed (in `OTokenSignImpl`) to use `SecureRandom.getInstanceStrong()`, which (on Linux and Mac at least) defaults to the `NativePRNGBlocking` implementation, which uses the (blocking) `/dev/random` device.

A static `new SecureRandom()`, which will choose a secure non-blocking PRNG, would be the right thing to use in this scenario.

This issue has come up in a bunch of projects:
- https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/ explains what is going on in the SecureRandom
- https://github.com/akka/akka/issues/20908 is an example of Akka making this change

timw:
Avoid embedded database creation blocking due to unnecessary use of strong SecureRandom.getInstanceStrong.

Fixes #9603

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/security/OTokenSignImpl.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/security/OTokenSignImpl.java
@@ -5,7 +5,6 @@ import com.orientechnologies.common.exception.OSystemException;
 import com.orientechnologies.common.log.OLogManager;
 import com.orientechnologies.orient.core.config.OContextConfiguration;
 import com.orientechnologies.orient.core.config.OGlobalConfiguration;
-import com.orientechnologies.orient.core.exception.ODatabaseException;
 import com.orientechnologies.orient.core.metadata.security.OToken;
 import com.orientechnologies.orient.core.metadata.security.jwt.OKeyProvider;
 import com.orientechnologies.orient.core.metadata.security.jwt.OTokenHeader;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/orientechnologies/orient/core/security/OTokenSignImpl.java
```diff
@@ -135,14 +134,7 @@ public String getDefaultKey() {
     if (configKey != null && configKey.length() > 0) key = Base64.getUrlDecoder().decode(configKey);
 
     if (key == null) {
-      try {
-        key =
-            OSecurityManager.digestSHA256(
-                String.valueOf(SecureRandom.getInstanceStrong().nextLong()));
-      } catch (NoSuchAlgorithmException e) {
```
mmacfadden, line 142: Is there a reason we have taken away the try catch?  It seems like NoSuchAlgorithmException won't actually be thrown here I guess.  However looking at the code for digestSHA256, it looks like a OConfigurationException could be thrown.  Should we catch that, still, and re-throw as before?

mmacfadden (commented):
I read the various article and references on this topic, and this looks like a reasonable and correct change.  I did notice that some exception handling code was change instead of just changing the random generation. While it doesn't look like that should cause a problem and was probably just clean up. Could it be safer to just leave that and just make the single line change?

core/src/main/java/com/orientechnologies/orient/core/security/OTokenSignImpl.java
timw: Only NoSuchAlgorithmException was caught here before, and is no longer possible, so removed handler for it.

luigidellaquila:
Hi @timw 

That seems good, thank you very much! Merging

Luigi

PR #9628, opened 2021-06-24
Fix StackOverflow error when adding the owner document to OTracked*

luigidellaquila:


tglman:
Hi, 
In terms of persistence, having the owner of a set being an element of the set, is not possible, So definitely OK to have a specific exception in the case, no need to fix the behaviour though, we need to crash in this cases.

Bye

luigidellaquila:
Uhm... from the domain point of view, having a document that points to itself is a legitimate use case.
Previous versions did not have this limitation, it started to fail in v 3.1
What is the problem at persistence level?

PR #9645, opened 2021-08-04
Update README.md

andrii0lomakin:


tglman (commented):
I would suggest to use the .org domain for the docs that are mainly related to the community edition.

andrii0lomakin:
Hi, @tglman agree. I have changed the domain. Thank you for the suggestion.

PR #9648, opened 2021-08-06
Use .gitattributes to enforce LF line endings

krichter722:
Use .gitattributes to enforce LF line endings which reduces merge and rebase effort


andrii0lomakin:
@luigidellaquila @tglman what do you think? It looks quite a big PR but in nutshell it facilities development not slowing down it.

andrii0lomakin:
And all changes are quite simple, only line endings are changed.

luigidellaquila:
It's not a bad idea, we just have to double-check that the only change in all these files is line ending.
If that's actually the case (as I think it is) for me it's good to merge

Thanks

Luigi

krichter722:
Hi guys, should have mentioned that in the description. Only line endings are changed. It's a pain to introduce .gitattributes in an existing project, I have no idea why it's not added by `git init`. You probably need to diff this locally in order to verify that there's no changes. In case you want to rebase this, it's probably easier to recreate on a new branch with .gitattritbutes and `git rm -r --cached . && git add .`.

luigidellaquila:
OK, it's easy to check also with GitHub: 

https://github.com/orientechnologies/orientdb/pull/9648/files?w=1

The only "strange" thing is the core/.DS_Store file (that should not be there btw). 
I'm merging the PR and then deleting the .DS_Store manually

Thanks!

Luigi

luigidellaquila:
Oh, ok, it was removed already, even better!

Thanks!

Luigi

PR #9671, opened 2021-09-06
Wip transactions cleanups

tglman:
This PR include two main cleanups
first remove the specific method for lookup of transactions with support of micro-transactions, because micro-transactions do not exists any more.

second merge basic transaction interface into the transaction interface because is not needed to have this separated now that there are no micro-transactions


tglman:
Manually merged, closing

PR #9682, opened 2021-09-19
Remove cache from .travis.yml in order to be able to spot missing dependencies

krichter722:
What does this PR do?
Building for the first fime (with `./mvnw install`) currently fails due to failing download of https://download.java.net/maven/2/com/fasterxml/jackson/core/jackson-annotations/2.12.1/jackson-annotations-2.12.1.jar. It's probably not noticed because all CI systems use caches.

Motivation
Being able to build orientdb.

Related issues
./.

Additional Notes
Don't know if you're actively using Travis CI. It seems like a good double check for this.

Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 sudo: false
 language: java
-cache:
-  directories:
-  - $HOME/.m2
 install: true
 script:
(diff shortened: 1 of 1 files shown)


andrii0lomakin:
Thank you for contribution :-)

PR #9683, opened 2021-09-19
Activate pmd

krichter722:
What does this PR do?
I stumbled of PMD in the pluginManagement section of the parent pom.xml. In case you wanted to use it and never found the time to set it up: There you go. It's also a proposal to enhance code quality by automating the enforcement of rules. For starters I put in a very simple rule which is only active for one module.

Motivation
Rebasing a few older branches has been tedious due to numerous changes to the code style. Enforcing them makes live easier and poses no overhead if everyone configures their formatters correctly.

Related issues
A list of issues either fixed, containing architectural discussions, otherwise relevant
for this Pull Request.

Additional Notes
This now fails the build if someone pushes something that violates PMD rules and didn't check it locally. That should be discussed in the dev team. If you find this useful, I can continue the setup for all modules with the Diamond and/or other rule(s).

Checklist

### DIFF
--- a/code-analysis/pom.xml
+++ b/code-analysis/pom.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+
+<!-- ~ Copyright 2010-2013 Orient Technologies LTD ~ ~ Licensed under the
+    Apache License, Version 2.0 (the "License"); ~ you may not use this file
+    except in compliance with the License. ~ You may obtain a copy of the License
+    at ~ ~ http://www.apache.org/licenses/LICENSE-2.0 ~ ~ Unless required by
+    applicable law or agreed to in writing, software ~ distributed under the
+    License is distributed on an "AS IS" BASIS, ~ WITHOUT WARRANTIES OR CONDITIONS
+    OF ANY KIND, either express or implied. ~ See the License for the specific
+    language governing permissions and ~ limitations under the License. -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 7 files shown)


code-analysis/pom.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
```
krichter722, line 1: This setup is recommended and necessary afaik for multimodule projets, see https://maven.apache.org/plugins/maven-pmd-plugin/examples/multi-module-config.html for details

code-analysis/pom.xml
```diff
@@ -0,0 +1,24 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.orientechnologies</groupId>
+        <artifactId>orientdb-parent</artifactId>
+        <version>3.2.2-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <artifactId>code-analysis</artifactId>
+    <version>0.1.0</version>
```
krichter722, line 22: You can version this if you want or not. Maybe it's necessary if deployment is causing trouble due to overwriting release artifacts. For my understanding it's a different track. The version can even remain constant.

andrii0lomakin:
Hi guys.
We have already tried to integrate PMD into our project. Unfortunately PMD generates a lot of false-negative warnings about our source code and I personally prefer do not include it. I appreciate all the work done for this PR and the fact that there are people which want to improve code quality of the project but it also mean a lot of overhead for us in future. If we talk about tools that help to automate code review. I prefer to use findbugs instead.

Issue #9695, opened 2021-10-08
LET expression evaluates $current differently from SELECT $current when subquery is fetching from the class target


ikysil:
## Expected behavior  
The projection of the `$b` in the outer `SELECT` should match the projection of the `$current`.

## Actual behavior  
Following response is received:
```
{
    "result": [
        {
            "$current.*": {
                "sqa": 1,
                "sqc": "ORole"
            },
            "$b.*": "#5:0",
            "$b.@class": "ORole"
        },
        {
            "$current.*": {
                "sqa": 1,
                "sqc": "ORole"
            },
            "$b.*": "#5:0",
            "$b.@class": "ORole"
        }
    ],
...
```

Here, `$b.*` is returning RID of `ORole`, thus stripping additional fields from the projection in the subquery.

## Steps to reproduce  
- create an empty DB
- execute following query:
```
select $current.*, $b.*, $b.@class from (select 1 as sqa, @class as sqc from OIdentity LIMIT 2)
let $b = $current
```

## Behavior of OrientDB 2.2.37
Following response is received for the same query:
```
{
    "result": [
        {
            "@type": "d",
            "@rid": "#-2:0",
            "@version": 0,
            "$currentsqa": 1,
            "$currentsqc": "ORole",
            "$bsqa": 1,
            "$bsqc": "ORole"
        },
        {
            "@type": "d",
            "@rid": "#-2:1",
            "@version": 0,
            "$currentsqa": 1,
            "$currentsqc": "ORole",
            "$bsqa": 1,
            "$bsqc": "ORole"
        }
    ],
```

PR #9696, opened 2021-10-08
fix(9695): Repairs for the LET expression evaluates $current differen…

ikysil:
…tly from SELECT $current when subquery is fetching from the class target #9695

What does this PR do?
Repairs the evaluation of the projections from the subqueries fetch from the class targets

Motivation
Regression

Related issues
#9695

Additional Notes
Anything else we should know when reviewing?

Checklist
- [X] I have run the build using `mvn clean package` command

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/executor/ProjectionCalculationStep.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/executor/ProjectionCalculationStep.java
@@ -34,15 +34,14 @@ public class ProjectionCalculationStep extends AbstractExecutionStep {
       @Override
       public OResult next() {
         OResult item = parentRs.next();
-        Object oldCurrent = ctx.getVariable("$current");
-        ctx.setVariable("$current", item);
         OResult result = calculateProjections(ctx, item);
-        ctx.setVariable("$current", oldCurrent);
+        ctx.setVariable("$current", result);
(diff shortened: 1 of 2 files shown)


andrii0lomakin:
@luigidellaquila what do you think?

luigidellaquila:
Hi @ikysil 

Thank you very much, I had a quick look at the PR but I'm not 100% convinced that it's correct in principle.
In particular, $current is supposed to keep a reference to current record, replacing it with calculated projections could lead to unexpected results when it's used in later execution steps.
Probably, the problem is not evident and the tests are passing because projections calculation is one of the last steps in the execution plan, so the impact of this change is minimal on the rest of the execution, but I'd like to double-check it more in depth just to be sure.

I'll keep you updated

Thanks

Luigi

luigidellaquila:
Hi @ikysil 

Quick update on this PR: I had a closer look at the fix and I concluded that the problem is real, but there is a better way to address it. The root problem is that the subquery execution does not add the `$current` value to the context; your fix soles the problem by letting the projections set it, but the projections are not the real owner of this operation (and in some cases there are no projections at all, so the fix is partial anyway).

I'm merging the PR (the test case is excellent, and believe me, having a PR with a failing test case alone is much more valuable than a fix without a test) and then applying a slightly different fix.

Thanks!

Luigi

Issue #9700, opened 2021-10-11
Infinite loop when trying to import a database containing a vertex having an attribut's class type equals to Opoint


aberdev:
## Expected behavior  
Import any exported database will succeed

## Actual behavior  
Database import never finish until the timeout is triggered.  

## Steps to reproduce  

- Export an existing database (v 3.0.27 ) with vertices that at least one have an attribut with class type equals to Opoint
- Try to import the exported database in 3.2.0 or 3.2.2 instance 
Debugging shows that an infinite loop is in progress as shown by the attached capture
![orientdb_debug](https://user-images.githubusercontent.com/72909882/136820198-b8110eac-85bf-4dfe-8510-f5ce24e7bddb.png)

There is any fix or idea to avoid that ?
Thanks

aberdev:
Hello ,
Finally , I fixed the issue like shown by the attached screenshot.
Is it Ok for you ?
![fix_orientdb_issue-9700](https://user-images.githubusercontent.com/72909882/136926517-c5e7560a-176f-422a-b798-0fa563c2edf6.png)
It goes in infinite loop because the value of firstRecord does never change inside the while loop.
Thanks

andrii0lomakin:
@tglman could you take a look?

andrii0lomakin:
@aberdev any chances that you will provide a pull request?

aberdev:
@laa Yes , it's done(#9702)

andrii0lomakin:
@aberdev thank you very much for the help we will release new versions soon.

PR #9702, opened 2021-10-12
fix(9700) : Fix inifite loop on tracking/untracking element

aberdev:
What does this PR do?
It fixes a blocking issue when importing a database exported from 3.2.07 to 3.2.0.
The import runs infinitely until timeout.
The cause is an infinite loop at the modified class: methods track and untrack

Motivation
Regression.

Related issues
https://github.com/orientechnologies/orientdb/issues/9700

Checklist


andrii0lomakin:
OK for me

PR #9719, opened 2021-11-02
minor fix to avoid a storage to go to read only if the backup fail

tglman:
During the backup should not be there any operation that do changes that make the storage inconsistent, the only case where there should be the possibility of something going wrong in the storage side, is during the freeze logic, that already make the storage readOnly itself so no need to have an additional catch on top.

Regards


andrii0lomakin:
Sorry, @tglman that is not the correct fix, if something becomes wrong with storage not related to backup, we still need to go in read-only mode. I prefer to throw an exception explicit to the backup process itself and mark it as a high-level exception.

andrii0lomakin:
@tglman I suppose you are right, it should be safe to keep the change as is.

PR #9729, opened 2021-12-05
Improve MAVEN build Performance

hongdongni:
Maven allows you to run the compiler as a separate process by setting `<fork>true</fork>`. This feature can lead to much less garbage collection and make Maven build faster. This project has more than 1000 source files. We can consider enabling this feature.

Maven can recompile only the classes that were affected by a change. This feature is the default. We can activate it by setting `<useIncrementalCompilation>true</useIncrementalCompilation>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


tglman:
Hi, 

It looks good except it update the PDF used for the blob tests, that is unrelated, would accept it without that file updated.

Regards

tglman:
Closing this for lack of feedback

PR #9788, opened 2022-04-25
3.1 fix using wrong variables

pkendall64:
Fixes bad math and synchronisation of the request/response index variables and using the wrong array length in one case.

Previously a volatile was used on the variables, but an increment is not an anomic operation! Also, When the request/response index was < 0 then the array index would be set to 0, so after 2.3billion messages only the first channel would be used for the next 2.38billion!

Checklist


tglman:
Hi,

This looks good, thank you very much for your work on this, merging it.

PR #9790, opened 2022-04-26
3.1 Set keep alive on server accepted sockets

pkendall64:
Client create sockets turn on keep-alive, so this PR also turns them on on socket connections accepted on the server side.

Because the distributed server sockets are not written to the server is not notified of a closed connection if a firewall in the middle just drops the connection after a connection lifetime period expires. By enabling keep-alive on the server side, the server is notified of a failed/dropped connection after the keep-alive timeout.

Checklist


tglman:
Hi,

Looks good, merging

PR #9791, opened 2022-04-26
Cap string length in binary channels

pkendall64:
Use same logic as readBytes to cap max chunk length at  at NETWORK_BINARY_MAX_CONTENT_LENGTH.
This prevents OOME when invalid protocol traffic is directed at binary port (e.g. by security scanners).

Motivation:
In our deployment our SRE's have a security scanner that probes all open ports and shoves random data as well as well-known attacks down them. We noticed that we were getting OOM exceptions and traced the root cause to this. We have subsequently disabled the security scanner from running against these ports, but thought it prudent to add a sanity check to orientdb.

Checklist

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/enterprise/channel/binary/OChannelBinary.java
+++ b/core/src/main/java/com/orientechnologies/orient/enterprise/channel/binary/OChannelBinary.java
@@ -143,6 +143,17 @@ public abstract class OChannelBinary extends OChannel
       OLogManager.instance()
           .info(this, "%s - Reading string (4+N bytes)...", socket.getRemoteSocketAddress());
       final int len = in.readInt();
+      if (len > maxChunkSize) {
+        throw new IOException(
+            "Impossible to read a string chunk of length:"
+                + len
+                + " max allowed chunk length:"
+                + maxChunkSize
+                + " see NETWORK_BINARY_MAX_CONTENT_LENGTH settings ");
+      }
+      if (debug)
+        OLogManager.instance()
+            .info(this, "%s - Read string chunk length: %d", socket.getRemoteSocketAddress(), len);
       if (len < 0) return null;
 
       // REUSE STATIC BUFFER?


tglman:
Hi,

Thank you, this is neat, it will take a bit more to verify this case, because this is introducing a new network failure, that is based on the content of the message, so every retry of a "failing message" will produce a failure, so additional check should be put also at sender side to try to catch and propagate the issue to the "invalid message producer", so it does produce a clear and deterministic failure in the false negative case.

Regards

timw:
Happy to do some more work on this if you can point to where these checks might be applied - the failure is just a plain IOException (as is the bytes failure now), so not sure how they'd differentiate from other channel issues on the sender side.

The default limit works out to 16MB, which I thought was more than adequate for the use of strings in the binary protocols - essentially it's just a DOS limit, not a protocol limit.

tglman:
Hi, 

Yes this is conceptually correct, the issue is that is legit to produce a document with a field with 16MB of string, so you can let's say make a javascript function that store a document with that content, and the node that is creating that document will successfully send it to all the other nodes, causing the other nodes to reject is requests, as an IOException for all the retry it does, making the node think is isolated, and disconnect from the network, probably is worth to add some check also on the sender side, to make sure that it understand the message is sending is not correct and raise an exception that interrupt the request, avoiding the node to self isolate.

One way to try the case would be as a said just make a function that insert a document with that content, the execution of that function should produce a failure and not isolate the node.

timw:
I've had a look at this, and the new check in `readBytes` was actually copied from the original check in `writeBytes` - i.e. the check on the sender side that prevents sending too large messages is already in place.
This commit simply closes the case where the packet isn't created by a binary client that makes that check (i.e. in the case of security scanners and other similar DOS tools).

tglman:
Hi,

well yes the case was for `writeString` and `readString` anyway merged and added the check also in `writeString` should be all sorted.

Regards

timw:
Yeah, my bad - too many branches and patches going on. 
Thanks for the extra fix and merge.

PR #9792, opened 2022-04-26
3.1 check distributed lock success

pkendall64:
When gaining a distributed lock if a timeout occurs the LockManager would return false, but nothing checks the return value and would bindly assume it has the lock.

This PR throws ODistributedLockException if the timeout occurs when trying to gain the lock.

Checklist


tglman:
Hi, 

Will run distributed tests on this and verify no regression happen and then merge it.

Regards

PR #9794, opened 2022-04-28
Support masking of Lucene query in log/exceptions

timw:
What does this PR do?

Provide a 'reportQueryAs' metadata parameter that will be reported as the text of the Lucene query on any parse failure, allowing a client to prevent sensitive data included in Lucene SEARCH_CLASS, SEARCH_INDEX and SEARCH_FIELDS functions being logged or included in exception messages.

Motivation

We have Lucene full text indexes on the names of people, deployed in a privacy sensitive domain where logs containing the name data would be considered sensitive and not able to be shared with operations staff outside of a particular jurisdiction.
Parse errors on Lucene search queries provided in `SEARCH_CLASS` et al would be logged and reproduced in exception messages.
We fixed the escaping issues that caused the parse errors, but it's preferable to be able to assert that the database logs won't contain sensitive data.

Related issues

Additional Notes

Checklist


timw:
I should have mentioned - will do doc PR if/when this gets accepted.

tglman:
Hi,

This looks good, merging

Issue #9721, opened 2021-11-22
Cannot rebuild a lucene index in 3.2.0


aberdev:
## Expected behavior  
Lucene Indexes can be rebuilt normally

## Actual behavior  
Error triggers when trying to redbuild an index 

## Steps to reproduce  
1-Create a new Class
2-Add some properties
3-Create a lucene index on one of the properties
4-Try de rebuild the created index
The attached capture shows the error's informations
![error building index lucene](https://user-images.githubusercontent.com/72909882/142863473-b62621cf-6393-4c07-b6a1-1ebe461ab944.png)

The above scenario runs without problem in 3.0.27
Thanks for any help or explanation

aberdev:
@laa can you confirm that is a regression comparing to 3.0.x version ?
Thanks

andrii0lomakin:
@aberdev yes , sure that is regression which we will fix.

johny65:
This is related to #9498 (see comment https://github.com/orientechnologies/orientdb/issues/9498#issuecomment-898586327). The problem is that the database gets unusable until the server is restarted, no new connections are possible. Every exception thrown by Lucene turns the database into that broken state.

aberdev:
Yes ,the issue is caused by the same behavior.
Is that mean we can't rebuilt lucene indexes  ? But, we can only drop&create them ?

timw:
It looks like the fixes applied in 72e7d639a5f96be237dc49c43c815ef7c9255400 to stop the DB going into an internal error status on read errors would fix this in 3.2.x/develop as well.

The root cause is actually that a rebuild index on a Lucene index has always (at least in 3.1/3.2) thrown an UnsupportedOperationException when it is rebuilt.
The delete/doDelete method in OIndexAbstract isn't designed for extension, so when an index is rebuilt it executes the private `OIndexAbstract.doDelete`, which attempts to iterate all index items and remove them before removing the index engine.
If `OLuceneIndexNotUnique` provided an override of `doDelete` that does what it's `delete` does now (which is simply to remove the engine from the storage), it would probably eliminate the actual error.

tglman:
Hi,

Rolling 3.2.7 with the relative fix, closing this.

PR #9802, opened 2022-05-09
Fix rebuild lucene index

pkendall64:
As commented in issue #9721 the fix for this is to make `OIndexAbstract.doDelete` protected and in `OLuceneIndexNonUnique` create a `doDelete` override that does the actual work of deleting the index from the existing `delete` override and remove that overridden `delete` function.

Checklist


tglman:
Hi,

Looks good, merging

PR #9814, opened 2022-05-24
Fix the locking in the Lucene indexer

pkendall64:
We have a production cluster with 4 nodes running in AWS.
Every now and then we get a node which fails where the node prints that the Lucene Index is closed and have track it to the locking in the `OLuceneIndexEngineAbstract` having a race condition.

1. The scheduled close task would pass the double check for `shouldClose` and about to close the index.
2. Another thread would update the `lastAccess` timestamp and pass through into `openIfClosed` and be ready to use the index.
3. The scheduled close task would then close the index.
4. The other thread will then use the closed index and produce the exceptions, effectively stopping the node.

We have removed  the ReentrantLock and changed to synchronised blocks as the Lock was unnecessary, and added the extra synchronisation around the `openIfClosed`, `close`, `flush` and `freeze` methods.

Checklist


timw:
This error occurs about once a week in production, and given the 3.1 ALL_ONLINE quorum escalation, it brings down the entire cluster for writes.
We've patched it in a local build for now, but would appreciate this being merged.

PR #9832, opened 2022-06-23
Fix thread pool executors to operate as intended

timw:
What does this PR do?

- Consolidates all construction of ThreadPoolExecutors into a single factory class that standardises the logging, thread naming and thread group management (which was inconsistent across many of the instances)
- Fixes many of the thread pool instances to scale as intended 

Motivation
Most of the thread pool executors in OrientDB look like they were designed to scale, but because of the peculiarities of the JDK ThreadPoolExecutor, do not.

Additional Notes

The JDK ThreadPoolExecutor has a very unintuitive API, where the behaviour of the BlockingQueue used by the executor determines whether the pool will scale beyond the core (up to the max threads). Only a queue that fails on `offer` will cause growth of the pool beyond the core, so any executor using an unbounded LinkedBlockingQueue (for example) would never grow beyond core.

I've factored all the instantiations into a single factory class, and created thread pool implementations that actually behave as the construction seems to have intended (except for specific cases as noted below). Someone more familiar with the parts of the OrientDB engine should review to ensure this was the intended behaviour, or whether a different behaviour is required.

The scaling implementation uses a custom blocking queue that tricks the ThreadPoolExecutor into scaling based on a target queue size. The target queue size can be as low as 0 (which will force the thread pool to grow in favour of queuing), and triggers the thread pool probabilistically as the queue grows closer to the target size.

I haven't tried to implement blocking - this is surprisingly hard to achieve safely given all the edge cases in the thread pool lifecycle.

Notes on specific thread pools:

- `OrientDBEmbedded` - intent seems to have been to scale from 1 to available processors, but would only ever allocate 1 thread. It now scales as intended, with a target queue size of 100. 
- `OWowCache` - this is simply a cached thread pool construct
- `Orient` - this made an attempt to implement a blocking construct, but because the override of `offer` never returned false, the pool size would never scale beyond the core size. It now scales as intended but will queue instead of blocking.
- `ODistributedDatabaseImpl` - this was intended to scale to `totalWorkers` but would never grow past 1 thread. It now scales between 1 and `totalWorkers`, with a target queue size of 0 forcing thread scaling in preference to queuing (since these are latency sensitive).
- `OPushManager` - this scaled from 0 to 5 threads, but would reject executions. New implementation behaves as previously.

Checklist

### DIFF
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -1,4 +1,0 @@
-# Enable auto-env through the sdkman_auto_env config
-# Add key=value pairs of SDKs to use below
-java=8.0.322-zulu
-
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -1,0 +1,4 @@
+# Enable auto-env through the sdkman_auto_env config
+# Add key=value pairs of SDKs to use below
+java=8.0.322-zulu
+
--- a/core/src/main/java/com/orientechnologies/common/thread/OThreadPoolExecutors.java
+++ b/core/src/main/java/com/orientechnologies/common/thread/OThreadPoolExecutors.java
@@ -1,92 +1,10 @@
 package com.orientechnologies.common.thread;
 
-import java.util.Random;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicInteger;
 
 public class OThreadPoolExecutors {
 
   private OThreadPoolExecutors() {}
-
-  /**
-   * A {@link ThreadPoolExecutor} with an unbounded work queue, that still scales the pool beyond
-   * the core pool size.
-   */
-  protected static class ScalingThreadPoolExecutor extends OThreadPoolExecutorWithLogging {
-
-    /**
-     * An unbounded work queue for a {@link ThreadPoolExecutor} that causes the pool to scale beyond
-     * the {@link ThreadPoolExecutor#getCorePoolSize()} core pool size}. <br>
-     * This is achieved by defining a target queue capacity, and wiring the {@link
-     * RejectedExecutionHandler} for the thread pool to re-queue using {@link #safeOffer(Runnable)}.
-     * <br>
-     * The work queue will probabilistically reject {@link #offer(Runnable) offers} to the work
-     * queue (with that probability growing to 100% as the indicated capacity is reached). The
-     * rejection of the offer will trigger pool growth in the thread pool, and once the max pool
-     * size has been reached will cause {@link RejectedExecutionHandler#rejectedExecution(Runnable,
-     * ThreadPoolExecutor) rejection}. <br>
-     * The implementation of {@link RejectedExecutionHandler} for this queue will then add it
-     * directly to the work queue, bypassing the rejection logic.
-     */
-    private static class ScalingQueue extends LinkedBlockingQueue<Runnable> {
-      private final int capacity;
-
-      private final Random rand = new Random();
-
-      private volatile boolean maxPoolReached = false;
-
-      public ScalingQueue(int capacity) {
-        super(); // Don't limit actual queue - capacity is used to signal ThreadPoolExecutor to grow
-        this.capacity = capacity;
-      }
-
-      @Override
-      public boolean offer(Runnable r) {
-        if (isEmpty()) {
-          maxPoolReached = false;
-        }
-        if (!maxPoolReached) {
-          final int size = size();
-          final int trigger = capacity <= 1 ? (capacity - 1) : rand.nextInt(capacity);
-          if (size > trigger) {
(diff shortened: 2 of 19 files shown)


.sdkmanrc
```diff
@@ -0,0 +1,4 @@
+# Enable auto-env through the sdkman_auto_env config
```
tglman: Is this file needed for some reasons ?

core/src/main/java/com/orientechnologies/common/thread/OThreadPoolExecutors.java
```diff
@@ -0,0 +1,247 @@
...
+    return newSingleThreadScheduledPool(threadName, Thread.currentThread().getThreadGroup());
+  }
+
+  public static ScheduledExecutorService newSingleThreadScheduledPool(
+      String threadName, ThreadGroup parentThreadGroup) {
+    return new OScheduledThreadPoolExecutorWithLogging(
+        1, new SingletonNamedThreadFactory(threadName, parentThreadGroup));
+  }
+
+  private abstract static class BaseThreadFactory implements ThreadFactory {
```
tglman: Same better in a specific file

core/src/main/java/com/orientechnologies/common/thread/OThreadPoolExecutors.java
```diff
@@ -0,0 +1,247 @@
...
+      final Thread thread = new Thread(parentThreadGroup, r);
+      thread.setDaemon(true);
+      thread.setName(nextThreadName());
+      return thread;
+    }
+
+    protected abstract String nextThreadName();
+  }
+
+  private static final class SingletonNamedThreadFactory extends BaseThreadFactory {
```
tglman: Same better in a specific file

core/src/main/java/com/orientechnologies/common/thread/OThreadPoolExecutors.java
```diff
@@ -0,0 +1,247 @@
...
+      this.name = name;
+    }
+
+    @Override
+    protected String nextThreadName() {
+      return name;
+    }
+  }
+
+  private static final class NamedThreadFactory extends BaseThreadFactory {
```
tglman: Same better in a specific file

core/src/main/java/com/orientechnologies/common/thread/OThreadPoolExecutors.java
```diff
@@ -0,0 +1,247 @@
...
+     * <br>
+     * The work queue will probabilistically reject {@link #offer(Runnable) offers} to the work
+     * queue (with that probability growing to 100% as the indicated capacity is reached). The
+     * rejection of the offer will trigger pool growth in the thread pool, and once the max pool
+     * size has been reached will cause {@link RejectedExecutionHandler#rejectedExecution(Runnable,
+     * ThreadPoolExecutor) rejection}. <br>
+     * The implementation of {@link RejectedExecutionHandler} for this queue will then add it
+     * directly to the work queue, bypassing the rejection logic.
+     */
+    private static class ScalingQueue extends LinkedBlockingQueue<Runnable> {
```
tglman: I would prefer this inner class would go to a specific file

core/src/main/java/com/orientechnologies/orient/core/Orient.java
```diff
@@ -271,24 +271,8 @@ public Orient startup() {
       final int cores = Runtime.getRuntime().availableProcessors();
 
       workers =
-          new OThreadPoolExecutorWithLogging(
```
tglman, line 274: Will the new implementation be blocking ? looking at I think is not, would be better that this executor is kept blocking, may create some behaviour problem changing it to not blocking.

tglman (commented):
Hi,

This looks good, just a small set of corrections, as well will run some additional tests on our side before merge this, anything I think this is solving some side problem in scheduling that I experienced in my tests, thank you for the work!!

.sdkmanrc
timw: No, that shouldn't have been committed.

core/src/main/java/com/orientechnologies/orient/core/Orient.java
timw: That was one of the questions I had - I've added blocking behaviour back in now.

timw:
Inner classes have been moved to top level, and blocking functionality reinstated.

The blocking is done as it was before - it's not a generally safe implementation (for that you need to handle a whole lot of task cancellation/shutdown/termination scenarios) but it's similar to what was there before, except now the thread pool will scale before growing the queue to the max queue size and then blocking.

PR #9844, opened 2022-08-03
Prevent Live Query thread hot looping on idle queue.

timw:
When the size of the ops queue goes to 0, the live query thread repeatedly loops trying to fill a batch size of 0 items (the min of queue.size() and the batch size), which results in a 100% CPU hot loop on the thread.

What does this PR do?
This re-implementation of the batching loop blocks on the first item, and polls on subsequent items, so the thread will block until an item arrives in an idle system, and never process an empty batch.

Motivation
Had multiple production nodes observed utilising 100% CPU consistently on a single core, indicating a hot loop on a single thread. Thread dump analysis revealed the live query thread was continuously looping on an empty batch. Code level testing revealed the thread hot loops on an empty queue.

Related issues
Also fixed counting/reporting of total events, and made the stop flag volatile for cross-thread visibility.

Additional Notes

Checklist


tglman:
Hi,
looks good, merging

PR #9855, opened 2022-08-15
Lucene cross class queries

timw:
What does this PR do?

Port the `SEARCH_CROSS` Lucene cross class index search function from the enterprise agent codebase.

Motivation
`SEARCH_CROSS` is described in the documentation for enterprise edition, but is not built/distributed with the open source based enterprise agent.

Additional Notes
There's a minor hack (along the lines of the existing minor hack) to allow the `LUCENE_CROSS_CLASS` index to be created without triggering the manual index deprecation. 

Checklist

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/index/OIndexManagerShared.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/index/OIndexManagerShared.java
@@ -129,7 +129,7 @@ public class OIndexManagerShared extends OIndexManagerAbstract {
             || indexDefinition.getClassName() == null
             || indexDefinition.getFields() == null
             || indexDefinition.getFields().isEmpty();
-    if (manualIndexesAreUsed) {
+    if (manualIndexesAreUsed && !"LUCENE_CROSS_CLASS".equals(algorithm)) {
       OIndexAbstract.manualIndexesWarning();
     } else {
       checkSecurityConstraintsForIndexCreate(database, indexDefinition);
--- a/lucene/src/main/java/com/orientechnologies/lucene/OLuceneCrossClassIndexFactory.java
+++ b/lucene/src/main/java/com/orientechnologies/lucene/OLuceneCrossClassIndexFactory.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2010-2016 OrientDB LTD (http://orientdb.com)
+ *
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/orientechnologies/orient/core/index/OIndexManagerShared.java
```diff
@@ -129,7 +129,7 @@ public OIndex createIndex(
             || indexDefinition.getClassName() == null
             || indexDefinition.getFields() == null
             || indexDefinition.getFields().isEmpty();
-    if (manualIndexesAreUsed) {
+    if (manualIndexesAreUsed && !"LUCENE_CROSS_CLASS".equals(algorithm)) {
```
tglman, line 132: I would not include the specific check for "LUCENE_CROSS_CLASS" and just leave the default check.

tglman (changes requested):
Him

This looks good, only one change I would prefer to remove the specific enable of LUCENE_CROSS_INDEX and still rely on the:
```
  INDEX_ALLOW_MANUAL_INDEXES(
      "index.allowManualIndexes",
      "Switch which allows usage of manual indexes inside OrientDB. "
          + "It is not recommended to switch it on, because manual indexes are deprecated, not supported and will be removed in next versions",
      Boolean.class,
      true),
      ``` 
      that  is the base behaviour as today, I will probably re-enable the manual index when there will be a better support in distributed, as the current develop there are some changes already that would allow the support of manual index to be re-enabled soon.

core/src/main/java/com/orientechnologies/orient/core/index/OIndexManagerShared.java
timw: It looks like the cross class index isn't really a manual index, it just collides with the manual index logic as it's created as a special case index. This was just replicating the logic in `OCreateIndexStatement.execute`, which special cases the index.
It should probably be modelled as a distinct type of index (i.e. maybe a virtual index) that has no key fields, but since the existing hack was in place I replicated it rather than extend the index SPI.

As you say, we can work around that by enabling manual indexes, but it's a bit messy either way.

timw:
We didn't end up using cross class indexes, so don't have any push to continue this work - happy for you to close the PR.

PR #9951, opened 2023-03-27
Guard database create to avoid partially initialised storage being created

timw:
### What does this PR do?
If the plugin isn't online, initialisation of newly created database will fail, resulting in a partially initialised database that will break when used (usually because the schema hasn't been loaded).

### Related issues
Separated out from #9854


timw:
This depends on #9956 - will rebase once that's merged - and part of 3b88eca (which factors out `isDistributedEnabled` and reimplements the check for whether the server is in distributed mode).
@tglman  - are you happy for me to drag in 3b88eca on this PR?

PR #9952, opened 2023-03-27
Log transaction id on re enqueue

timw:
### What does this PR do?
Adds transaction ID to logging when re-enqueued.

### Related issues
Separated out from #9854


tglman:
Hi,

Thanks, this is easy to merge straight, merging.

PR #9954, opened 2023-03-27
Fix race in registering installing database to guard against concurrent installs

timw:
### What does this PR do?

Prevents multiple concurrent database installs from executing.

### Related issues
Separated out from #9854


tglman:
Hi, 

Looks good,  merging

PR #9955, opened 2023-03-27
Make ViewManager updates resilient to offline DB status.

timw:
### What does this PR do?

Prevents an offline DB status from breaking the view manager update task.

### Related issues
Separated out from #9854


timw:
@tglman  This depends on the tracing executor variants introduced in db457608a7124637442a6acd19f1417e9479f09c (which in turn depend on the introduction of tracing executors).
Do you want me to pull in the tracing changes now, or switch this to the non-tracing variant for now?

tglman:
Hi,

Thanks for this, yeah will check if this is actually needed, in the 3.2.x branch all this logic as been already re-worked, so this is relevant only for 3.1.x

Issue #10068, opened 2024-01-28
Remove com.orientechnologies.orient.core.storage.cache.OCachePointer#getBufferDuplicate 


andrii0lomakin:
The new JDK 21 API allows it to work with all ByteBuffer operations using offsets. Using internal positions inside the byte buffer is unnecessary.  Remove the creation of new ByteBuffer instances and use ByteBuffer API that does not rely on internal position inside byte buffer.

tglman:
Hi, 

That's good, what is the minimum java version this API is supported ?

andrii0lomakin:
Java 17 should work

andrii0lomakin:
Fixed.

PR #10089, opened 2024-02-06
Issue 10068

andrii0lomakin:
Fix of issue #10068 .

1. Exception on the Linux platform was fixed during storage initialization.
2. The build number is stored correctly during the build of the project.


andrii0lomakin:
This branch was based on 4.0.x, so I had to rebase a bunch of commits.

core/src/main/java/com/orientechnologies/orient/core/db/document/ODatabaseDocumentTx.java
tglman, line 1: I'm planning to delete completely this class on 4.0.x so no need to do evolution on this

andrii0lomakin: I got it: -) I planned to do the same.

andrii0lomakin:
I have also added `maven.release` property to avoid usage of API introduced in later JDKs.

tglman:
Hi,

A quick review on this:

this commit https://github.com/orientechnologies/orientdb/pull/10089/commits/7ee1aa6bfb7f773d59fcd8e9e8772d15a87f3a07 should be already somehow on develop this: https://github.com/orientechnologies/orientdb/pull/10089/commits/f92ddb45414ee789634a9a1752952dfb22eef490 we do support only java17+ so we cannot use jdk18+ dependencies develop should be already in a good state for cert generation

this: https://github.com/orientechnologies/orientdb/pull/10089/commits/0ba2b1406229755f78a3ef6d536978960a50dbc1 is quite welcome, but it has a bit of formatting in it that maybe should not be in this commit

this: https://github.com/orientechnologies/orientdb/pull/10089/commits/573b160d386c3ba3c05f19ecd5fc223ac3c4c03f is good to merge
 this: https://github.com/orientechnologies/orientdb/pull/10089/commits/53a3936e1831c7562e0f5e7855dca1f22a2763d6 good to merge
this: https://github.com/orientechnologies/orientdb/pull/10089/commits/404b3fcf213b21daaa209264a87700e8e4f78498 more than happy to merge it
this: https://github.com/orientechnologies/orientdb/pull/10089/commits/a280f2b76b66edc0f8d3b3089abffdb26ed57046 good to merge as well
 
 these two:https://github.com/orientechnologies/orientdb/pull/10089/commits/d289fd37c7e4f9cf4d5cc65c112a5a850be365ad,  https://github.com/orientechnologies/orientdb/pull/10089/commits/d6d8909df14debe313c359c12d23ec04d4a7de42 should be redundant as well, it's ok if some of the good commit introduce some specific java 21 changes
 
Code formatting, so : https://github.com/orientechnologies/orientdb/pull/10089/commits/18fe41df45e8127d1f16289feb034333164b8605 and  https://github.com/orientechnologies/orientdb/pull/10089/commits/60bcf318e44ef23212ed4e911d83e6012ea8558a  I do not want yet to upgrade the development formatter for another 1/2 weeks, because of some old changes I've pending, so I would prefer keep these out.

andrii0lomakin:
@tglman jdk8on is about 1.8+ not about 18+

andrii0lomakin:
@tglman may I ask you to add comments on commit level, so we can discuss observation ?

andrii0lomakin:
It is  hard to discuss observations otherwise

andrii0lomakin:
@tglman I will roll back code formatting.

andrii0lomakin:
@tglman, if you have already added changes, can we go with the current code formatter? The old one does not support new language features. Also as I mentioned, jdk18on dependency is not about Java 18, it is about Java 1.8+

core/pom.xml
```diff
@@ -276,6 +277,17 @@
             <artifactId>jnr-posix</artifactId>
             <version>3.1.15</version>
         </dependency>
+        <dependency>
```
tglman: I saw this new dependency, I'm not against using slf4j instead the current formatter, but better add the dependency when the integration is done, otherwise it just add liability without benefits

andrii0lomakin: Sure will remove it, agree.

PR #10107, opened 2024-02-08
test: refactor http tests and migrated to a newer API

tglman:
Hi,

I think this is fine we can merge it, I just want to understand why this was needed, is just for the more recent version or are other reasons.


andrii0lomakin (approved):
Exactly my intent. Thank you :-)

PR #10108, opened 2024-02-08
refactor: use direct locking instead of syncrhonize for legacy database

tglman:
Hi,

So this swap the synchronize for the locking for some legacy component, I guess this is for keep separate the remote and embedded locking, is this the case, being a legacy component that is plan to be delete before release I do not see much the point, but maybe this may improve the execution of some test case meanwhile they are migrated to the new approach.

Regards


andrii0lomakin (approved):
Just in time. I hope to move to virtual threads in next releases. Thank you for your changes.

PR #10297, opened 2024-09-03
Actually run the distributed tx expiry task

timw:
What does this PR do?

Actually create a timer and schedule the distributed transaction timeout task.
The `com.orientechnologies.orient.server.distributed.impl.ODistributedDatabaseImpl#startTxTimeoutTimerTask` method creates the timeout task, but never schedules it on a Timer.

Motivation

During heap dump analysis we observed some transactions remaining in the tracking list forever.

Related issues

Additional Notes

Checklist

### DIFF
--- a/distributed/src/main/java/com/orientechnologies/orient/server/distributed/impl/ODistributedDatabaseImpl.java
+++ b/distributed/src/main/java/com/orientechnologies/orient/server/distributed/impl/ODistributedDatabaseImpl.java
@@ -648,9 +648,11 @@ public class ODistributedDatabaseImpl implements ODistributedDatabase {
         new TimerTask() {
           @Override
           public void run() {
-            checkTxTimeout();
+            context.execute(() -> checkTxTimeout());
           }
         };
+    final long timeout = OGlobalConfiguration.DISTRIBUTED_TX_EXPIRE_TIMEOUT.getValueAsLong();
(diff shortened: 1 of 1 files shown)


distributed/src/main/java/com/orientechnologies/orient/server/distributed/impl/ODistributedDatabaseImpl.java
```diff
@@ -651,6 +654,9 @@ public void run() {
             checkTxTimeout();
           }
         };
+    final long timeout = OGlobalConfiguration.DISTRIBUTED_TX_EXPIRE_TIMEOUT.getValueAsLong();
+    timer = new Timer("Distributed database tx expiry", true);
+    timer.scheduleAtFixedRate(txTimeoutTask, timeout / 3, timeout / 3);
```
tglman: We do have actually a global timer in the context so ` context.schedule` should do, just to keep on check the number of threads running.

Probably is also better to execute the operation in the context thread pool, so do ~ ` context.execute(() -> checkTxTimeout())` because the timer is shared is better that the timer do just timing, I can add this myself later

(and did we really missed that? :facepalm: )

tglman (commented):
This look good, just using the global timer will do!

tglman:
Perfect, Thanks!

PR #10299, opened 2024-09-04
Support debug mode on all console exceptions.

timw:
What does this PR do?

Handles all uncaught console exceptions with the `printError` handler, which supports printing of exception stack traces by configuring `set debug true`.

Motivation

Attempting to `TRUNCATE RECORD` with a broken record in the database that was causing an `IndexOutOfBoundsException`, but not being able to see the full stack trace of where the exception was occurring.

Related issues

Additional Notes

Technically the `try { ... } catch (Exception e) { printError(e) }` pattern throughout the console is redundant with this change.

Checklist


tglman:
Hi,

Looks good, merging

Issue #10391, opened 2025-01-22
Wrong value saved in EMBEDDED


adriancoca:
OrientDB Version: 3.2.22
Java Version: 8
OS: linux docker

Hello
When creating or updating a vertex the saved value of producedQuantity is 35.92767**3**

create vertex package set data={"@type":"d",producedQuantity: 35.927675, numberOfPackages: 1}

![Image](https://github.com/user-attachments/assets/37bb783c-9804-4eb1-88ab-571a34707168)

Update

![Image](https://github.com/user-attachments/assets/649dde73-f3d9-4fff-aa3f-04be4ffdc82b)

I am not sure if we are doing something wrong or is a conversion issue that can be solved.

tglman:
Hi, 

This looks like some float/double rounding, what is the type of `producedQuantity` ?

OrientDB support also decimal type that is based on java `BigDecimal` is a bit more heavy in the implementation but it keep the precision.

adriancoca:
Vertex package have a property defined as below (data ). I don't think I can set the type on key producedQuantity .

![Image](https://github.com/user-attachments/assets/572eb444-c88c-4b4a-a2d3-6c67c5b5bd03)

tglman:
Hi,

You can set the type of that creating another class (preferably abstract) and used it as a linked_class of the of the data property, than the property of the embedded class should match the type of the linked class

adriancoca:
Hello again,

I tried `create vertex packagetest set data2={"@type":"d",producedQuantity: 35.927675}` but I receive the message 

`com.orientechnologies.orient.core.exception.OValidationException: The field 'PackageTest.data2' has been declared as EMBEDDED with linked class 'PackageDetail2' but the record is of class 'PackageDetail2' that is vertex class DB name="testcarettavineri" Error Code="4"`
Also I tried with PackageDetail Abstract and the error is 
`com.orientechnologies.orient.core.exception.OSchemaException: Cannot create a document of the abstract class 'PackageDetail' DB name="testcarettavineri"`

![Image](https://github.com/user-attachments/assets/19050162-8387-4fbd-b4ab-594bce6d4da4)

![Image](https://github.com/user-attachments/assets/6595ddf2-c65c-49c9-bb28-596ffac0f976)

tglman:
Hi,

It sounds like that `PackageDetail2` extends `V` it shouln't

adriancoca:
[...]

PR #10423, opened 2025-03-18
fix: Casting to float lose precision in some cases

EnascutaRazvan:
What does this PR do?
I reproduced the case and it seems like when casting to float the precision is lost
Related to -> #10391

Motivation
Because i use this database everyday

Related issues
#10391

Additional Notes
I don't know exactly why that cast to float was used, but it seems generates some problems

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OFloatingPoint.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OFloatingPoint.java
@@ -40,7 +40,13 @@ public class OFloatingPoint extends ONumber {
     } else {
       try {
         double returnValue = Double.parseDouble(stringValue) * sign;
+        float floatValue = (float) returnValue;
+        double check = (double) floatValue;
+        if(check == returnValue) {
+         finalValue= floatValue;
+        } else {
         finalValue = returnValue;
+        }
       } catch (Exception ignore) {
         return null; // TODO NaN?
       }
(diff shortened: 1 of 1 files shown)


tglman:
Hi,

Thank you, just triggered the ci build on this, if all pass will merge it and port it to 3.2.x

Bye

tglman:
Hi,

Tests are failing, so we do have cases where we rely on this conversion, looking in the code is possible to avoid this conversion specifying the double type at the end of the number, so something like: `35.927675D` would this work for your use case ?

Regards

tglman:
Hi,

Maybe another possible approach is replace the check: `if (Math.abs(returnValue) < Float.MAX_VALUE) ` with a reconversion to the double value, to see if the value still matches, like:
```
float floatValue = (float) returnValue;
double check = (double) flatValue;
if(check == returnValue) {
 finalValue= flatValue;
} else {
finalValue = returnValue;
}
```

Can you verify if a check like this works in your case and update the PR?

tglman:
One last thing (hopefully ...)

in case you update this PR can you add a test case for the case conversion error in OInsertStatementExecutionTest or in OUpdateStatementExecutionTest so we avoid to get regression in future.

EnascutaRazvan:
I ll try it tomorrow, ty

tglman:
Hi, 

Tried to run the tests, they fail for formatting, you can run `mvn spottless::apply` to fix all the formatting issues.

(waiting for the test as well )

tglman:
Hi, I had to rework this because the fix was going to create other failures, I made a fix that cover better the cases and landed it in 3.2.x and develop branch, closing this.

PR #10455, opened 2025-05-20
feat(backup): Log cluster names in debug only

ikysil:
### What does this PR do?

The background backup executor logs full list of cluster names in the database being backed up on the INFO level. This information is useless for the usual (non-debug) situations but consumes a lot of space in the output for a relatively large database (~8K clusters in our case).

This PR introduces a change so that the full list of cluster names is written only when `DEBUG` logging is enabled.

### Motivation

Improve OPS side of DEV.

### Related issues
N/A

### Additional Notes
N/A

### Checklist
- [ ] N/A - My unit tests cover both failure and success scenarios


tglman:
Hi,

That looks good, merging it.

PR #10512, opened 2025-08-14
Correct disk cache size tracking.

timw:
Increment cache size when new page pointers added to cache.

What does this PR do?

Update the `AsyncReadCache.cacheSize` when new page pointers mutate the cache map, otherwise when those entries are freed the cache size diverges from the actual cache data map size and goes below zero over time.

Motivation

We observed the disk cache used memory going negative (have exposed it via JMX) while the underlying data map size remains positive. After some debugging it looked like `addNewPagePointerToTheCache` was leaking a page in the counter every time it was called.


tglman:
Hi @timw, 

Thank you for this, all check passed will merge straight away, after that will run the long running tests that check low level memory usage that should catch issues around this part.

PR #10548, opened 2025-09-10
Correct calculation of cluster records size on record update and delete

timw:
This PR corrects the adjustment of records size in clusters on update and delete operations.
The chunk overheads were not being applied when calculating the records size delta, and additionally not all chunks in the record were being considered (so multi chunk records had very wrong deltas).
This is observed as records size dropping by (typically) 9 bytes on each update that doesn't change the actual record content size, or when a record is deleted.

Motivation
Found this while working on a feature that exposes records size (and file sizes and record counts) for clusters in an SQL command, and observing that adding properties (which applies record updates during DB migration) and updating records via SQL causes cluster records size counts to go to 0 or even negative values.

Additional Notes
I've also implemented checking the records size in the cluster ITs for the V2 cluster format (which covers the multi-chunk records as well).
The V0 and V1 clusters use a different scheme to calculate records size, and have different numbers than V2, so I've just stubbed out their record size checks for now - it might be an issue worth fixing overall, but I haven't delved into why the V0/V1 clusters have different records size numbers and don't use them.

Checklist


tglman:
Hi,

Thanks for checking and fixing this, will run the long integration test on this, and if all goes fine I will merge it.

Issue #10427, opened 2025-03-25
【BUG】Three masters in cluster mode, any master node re-pulled up leads to cluster synchronization data jamming


Ak1yama-mio:
In a cluster consisting of three masters, I took one of the master nodes offline and pulled it up again while maintaining write pressure, at which point the logs showed that the master node was repeatedly synchronizing data to that node, but could never complete the synchronization, which kept the write requests being denied.
This process will continue for a long time, even if the write request is stopped at this time, the synchronization can not be completed.

orientdb version: 3.0.34

The contents of the error log are as follows:

```
[OLocalPaginatedStorage]Error installing database from network
com.orientechnologies.orient.core.exception.ODatabaseException: Cannot create database 'test'
Caused by:
com.orientechnologies.orient.core.db.OrientDBEmbedded.restore(0rientDBEmbedded.java:451)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin$6.call(ODistributedAbstractPlugin.java:2006)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin$6.call(ODistributedAbstractPlugin.java:1945)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin.executelnDistributedDatabaseLock(ODistributedAbstractPlugin.java:1785)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin.installDatabaseOnLocalNode(ODistributedAbstractPlugin.java:1945)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin.installDatabaseFromNetwork(ODistributedAbstractPlugin.java:1612)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin.requestDatabaseFullSync(ODistributedAbstractPlugin.java:1427)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin.requestFullDatabase(ODistributedAbstractPlugin.java:1109)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin$3.call(ODistributedAbstractPlugin.java:1027)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin$3.call(ODistributedAbstractPlugin.java:957)
com.orientechnologies.orient.server.distributed.impl.ODistributedAbstractPlugin.executelnDistributedDatabaseLock(ODistributedAbstractPlugin.java:1785)
[...]

PR #10549, opened 2025-09-11
feat: Force embedded ridbags when distributed

ikysil:
What does this PR do?
Force embedded RID bags to avoid replication errors in distributed mode.
See https://orientdb.dev/docs/3.2.x/general/Concurrency.html#concurrency-when-adding-edges

Motivation
Minimize surprises when updating large graphs in distributed mode.

Related issues
N/A

Additional Notes
N/A

Checklist

### DIFF
--- a/distributed/src/main/java/com/orientechnologies/orient/server/distributed/impl/ODistributedPlugin.java
+++ b/distributed/src/main/java/com/orientechnologies/orient/server/distributed/impl/ODistributedPlugin.java
@@ -43,6 +43,7 @@ import com.orientechnologies.orient.core.OSignalHandler;
 import com.orientechnologies.orient.core.Orient;
 import com.orientechnologies.orient.core.command.OCommandDistributedReplicateRequest;
 import com.orientechnologies.orient.core.command.OCommandOutputListener;
+import com.orientechnologies.orient.core.config.OContextConfiguration;
 import com.orientechnologies.orient.core.config.OGlobalConfiguration;
 import com.orientechnologies.orient.core.db.ODatabaseDocumentInternal;
 import com.orientechnologies.orient.core.db.ODatabaseInternal;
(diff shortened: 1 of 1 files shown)


tglman:
@ikysil 

What issue do you have with the tree ridbag in distributed ? do you use lightweight edges ?

ikysil:
Hi @tglman 

The number of full syncs dropped to almost zero in a distributed cluster of 3 MASTER and 2 REPLICA nodes after applying this configuration.

We don't use lightweight edges (AFAIR).

Scenario before applying these properties:
- We use round-robin connection strategy on a client, so that write load is distributed between masters.
- We observed that vertex version is not modified when only edges were added and/or removed.
- One master commits a change with edge(s) modifications only - vertex version is not modified.
- Another master commits a change with property modification, modifying vertex version.
- If edge modification is applied after property modification, the cluster can not agree during the second phase of the transaction, complaining about stale version.
- A node retries a couple of times, then asks for a delta sync, retries that a few times, fails, then asks for a full sync.
- Occasionally, we observe behavior described in #10427. We have to pull affected node from a cluster and add it back later - during quiet time.

The full sync is very expensive for us and takes more than 30 minutes.

tglman:
Hi,

This sounds like a bug with the integration of ridbag trees in the distributed flow, I will try to write a test case for this scenario, this may be due to the fact that changes in the ridbag tree do not change the version of the document, allowing concurrent write with other write operations, we do have some distributed scheduling to avoid concurrent writing in the same document, but maybe this check is skipped for changes in ridbag trees.

Any support on reproducing the case is welcome!

tglman:
Hi,

I could write a somewhat minimal test case that reproduced the case and fixed it, so from the next hotfix the issue with inverted apply of distributed transaction due to ridbags tree is solved.

the commit that fix it is referring to this PR.

Bye

tglman:
Hi,

the 3.2.48 is released that should fix this case.

ikysil:
Hi @tglman 

TY for the fix and release.
It takes some time to verify as we don't want to risk production data integrity.

Let's close this PR as it is not needed by itself.
I will add a comment after the verification.

Issue #10452, opened 2025-05-08
Is it possible to enable natural sort order for string-fields in ORDER BY?


StenGruener:
hi,

is there some built-in possibility to do a natural string sorting in orientDB?

We got elements sorted alphabetically as: M1, M10, M2, but require M1, M2, M10 order referred as "[Natural sort order](https://en.wikipedia.org/wiki/Natural_sort_order)".

Any included COLLATE statements to support those? If now how to workaround?

BR and thanks!

tglman:
Hi,

As today we do not have this kind of sorting supported in the engine, even though it should be possible to support this implementing a custom collate, trough define an implementation of this https://github.com/orientechnologies/orientdb/blob/develop/core/src/main/java/com/orientechnologies/orient/core/collate/OCollateFactory.java and plugging it in with a resource locator.

In this way you can add that to an existing distribution adding it with a jar.

We would accept a PR that add that in the main distribution, even though will be shipped by default only in the next major (4.0.x)

PR #10578, opened 2025-10-13
add feat "https://github.com/orientechnologies/orientdb/issues/10452"

klh666:
What does this PR do?
A brief description of the change being made with this pull request.
As today we do not have this kind of sorting supported in the engine, even though it should be possible to support this implementing a custom collate, trough define an implementation of this https://github.com/orientechnologies/orientdb/blob/develop/core/src/main/java/com/orientechnologies/orient/core/collate/OCollateFactory.java and plugging it in with a resource locator.

In this way you can add that to an existing distribution adding it with a jar.
Motivation
What inspired you to submit this pull request?

Related issues
A list of issues either fixed, containing architectural discussions, otherwise relevant
for this Pull Request.
#10452

Additional Notes
Anything else we should know when reviewing?

Checklist

### DIFF
--- a/core/src/main/java/com/orientechnologies/orient/core/collate/NaturalStringCollate.java
+++ b/core/src/main/java/com/orientechnologies/orient/core/collate/NaturalStringCollate.java
@@ -1,0 +1,84 @@
+package com.orientechnologies.orient.core.collate;
+
+import java.util.Comparator;
+
+public class NaturalStringCollate implements OCollate{
+
+
+    public static final String NAME = "NATURAL";
+
+    private final Comparator<Object> naturalComparator = new NaturalStringComparator();
+
+    @Override
+    public String getName() {
+        return NAME;
+    }
+
+    @Override
+    public Object transform(Object obj) {
+        return obj;
+    }
+
+    @Override
+    public int compareForOrderBy(Object o1, Object o2) {
+        return naturalComparator.compare(o1, o2);
+    }
+
+    // 自然字符串比较器实现
+    private static class NaturalStringComparator implements Comparator<Object> {
+        @Override
+        public int compare(Object o1, Object o2) {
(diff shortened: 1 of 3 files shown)


tglman:
Hi,

Thanks to submit this, will review it asap.

core/src/main/java/com/orientechnologies/orient/core/collate/NaturalStringCollate.java
```diff
@@ -0,0 +1,84 @@
...
+        return obj;
+    }
+
+    @Override
+    public int compareForOrderBy(Object o1, Object o2) {
+        return naturalComparator.compare(o1, o2);
+    }
+
+    // 自然字符串比较器实现
+    private static class NaturalStringComparator implements Comparator<Object> {
```
tglman, line 28: Hi, 

Is there any reason for not use directly https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html#naturalOrder-- 

I do prefer keep a generic implementation that is already battle tested

core/src/test/java/com/orientechnologies/orient/core/collate/NaturalStringCollateTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+import static org.junit.Assert.*;
+
+public class NaturalStringCollateTest {
+
+    @Test
+    public void testNaturalSortOrder() {
+        NaturalStringCollate collate = new NaturalStringCollate();
+
+        // 测试基本的自然排序
```
tglman, line 17: We do not have anybody in our team now that can read this language, the project official comment language is English, so the comments in other languages will be removed on merge

tglman (changes requested):
Can you double check this couple of things than is good to merge for me.

PR #10615, opened 2025-12-03
Fix flaky test in OLuceneSortTest by making assertion not relying on order

lbh930:
What does this PR do?
This PR fixes a flaky test `OLuceneSortTest.shouldSortByReverseDocScore` by relaxing the assertion to ignore the order of the results.

Motivation
The test `shouldSortByReverseDocScore` was detected to be flaky by NonDex. The test sorts results by Lucene's internal Doc ID (`type:'DOC'`). However the index is created on existing data and the order in which OrientDB iterates over the records to populate the index is not deterministic. Since Doc IDs are assigned based on indexing order the resulting sort order is also non-deterministic. Asserting a specific order is brittle for this test.

Related issues
None.

Additional Notes
None.

Checklist
[√] I have run the build using `mvn clean package` command
[N/A] My unit tests cover both failure and success scenarios


tglman:
Hi,

Thank you very much, will run the CI and merge it

PR #10626, opened 2025-12-07
Fix flaky test in HttpImportTest.testImport by fixing race condition

lbh930:
What does this PR do?
This PR fixes the flaky test `com.orientechnologies.orient.test.server.network.http.HttpImportTest.testImport`. The registration of the import and upload commands in ONetworkProtocolHttpDb is moved to happen earlier within registerStatelessCommands() instead of after super.config().

Motivation
The test was detected to be failing by NonDex. The issue is a race condition in ONetworkProtocolHttpDb. The config method calls super.config() that starts the network listener thread. However the import command was being registered after this call. If the test request hit the server before the main thread reached that registration line, the command wouldn't be found. Moving the registration to registerStatelessCommands makes sure the commands are available before-hand.

Related issues
None.

Additional Notes
None.

### Checklist
- [✓] I have run the build using `mvn clean package` command
- [N/A] My unit tests cover both failure and success scenarios


tglman:
Hi, 

This looks correct, merging it

PR #10812, opened 2026-06-30
test: cover omatchstatement copy behavior

amirdeljouyi:
What does this PR do?

Adds focused `OMatchStatementTest` coverage for copying a complex `MATCH` statement and handling an invalid copy state.

The tests verify that copied statements keep the expected clauses, duplicate mutable nested structures instead of sharing them, and fail when required match expressions are missing.

Covered lines include [`OMatchStatement.java:121-134`](https://github.com/orientechnologies/orientdb/blob/develop/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OMatchStatement.java#L121-L134), and the main copy logic in [`OMatchStatement.java:410-448`](https://github.com/orientechnologies/orientdb/blob/develop/core/src/main/java/com/orientechnologies/orient/core/sql/parser/OMatchStatement.java#L410-L448).
Line coverage increases by around 24%.

Motivation

The copy behavior was not directly verified for mutable nested structures or a missing required expression list, so those regressions could slip through.

Related issues

None.

Additional Notes

This is a test-only change.

Checklist

### DIFF
--- a/core/src/test/java/com/orientechnologies/orient/core/sql/parser/OMatchStatementTest.java
+++ b/core/src/test/java/com/orientechnologies/orient/core/sql/parser/OMatchStatementTest.java
@@ -1,5 +1,8 @@
 package com.orientechnologies.orient.core.sql.parser;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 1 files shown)


tglman:
Hi,

This looks good, it would be more interesting to know the context of this patch, but no reason to hold this, merging.

